PR #2, opened 2018-10-11
Deposit control

roeierez:
This PR aligns breez mobile with breez library account changes.
In addition it: 
1. Doesn't allow deposit in case the user is not allowed.
2. In "add funds" page show the user the right amount he can send on-chain.


roeierez:
@yaslama it is a product requirement to let the client know that only when the balance is bellow this threshold he will be able to deposit.
it doesn't contradict the maxAllowedDeposit.

PR #4, opened 2018-10-11
BREEZ-414: Increase the 'processing payment' snackbar duration to mat…

icota:
…ch the timeout in lightninglib

### DIFF
--- a/lib/routes/user/home/account_page.dart
+++ b/lib/routes/user/home/account_page.dart
@@ -57,7 +57,7 @@ class _AccountPageState extends State<_AccountPage> {
     _statusSubscription = widget._accountBloc.accountStream.listen((acc) {
       if (acc.paymentRequestInProgress != null && acc.paymentRequestInProgress.isNotEmpty && acc.paymentRequestInProgress != _paymentRequestInProgress) {        
         Scaffold.of(context).showSnackBar(new SnackBar(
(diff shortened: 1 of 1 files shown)


lib/routes/user/home/account_page.dart
```diff
@@ -57,7 +57,7 @@ class _AccountPageState extends State<_AccountPage> {
     _statusSubscription = widget._accountBloc.accountStream.listen((acc) {
       if (acc.paymentRequestInProgress != null && acc.paymentRequestInProgress.isNotEmpty && acc.paymentRequestInProgress != _paymentRequestInProgress) {        
         Scaffold.of(context).showSnackBar(new SnackBar(
-            duration: new Duration(seconds: 5), content: new Text("Processing Payment...")));
+            duration: new Duration(seconds: 60), content: new Text("Processing Payment...")));
```
roeierez: I guess we want to show the snackbar until and event of either success or failed payment arrived.
I think infinite duration will better fit here.

icota: I agree but I couldn't find a way specify an infinite Duration. If parameter left unset it defaults to 4 seconds. Any ideas?

roeierez: Maybe this can work?
new Duration(seconds: Int32.MAX_VALUE.toInt())

PR #6, opened 2018-10-12
Notify client when lightninglib not started

icota:
Update protobuf files


lib/routes/pos/home/pos_home_page.dart
```diff
@@ -27,6 +33,29 @@ class PosHome extends StatefulWidget {
 class PosHomeState extends State<PosHome> {
   final GlobalKey<ScaffoldState> _scaffoldKey = new GlobalKey<ScaffoldState>();
 
+  void _listenNoConnection(BreezBridge breezLib) {
```
roeierez: Two comments:
1. It is better to keep the UI without access to the services layer. I think the listening to breezLib should be from the bloc and expose something like ACCOUNT_SERVICE_DOWN event via stream.
2. It is better to extract the same code that exists in the client and in the pos to its own file and re-use it in both Home pages.

PR #10, opened 2018-10-22
Calendar filter

erdemyerebasmaz:


lib/widgets/calendar_dialog.dart
```diff
@@ -0,0 +1,188 @@
...
+  _clearFilter() {
+    setState(() {
+      _startDate = widget._firstDate;
+      _endDate = new DateTime.now();
+      _startDateController.text = "";
+      _endDateController.text = "";
+    });
+  }
+
+  String _formatTransactionDate(DateTime date) {
```
roeierez: _formatTransactionDate is not relevant here, there is no transactions...
I addition to that it is duplicated there are lots of uses of: 
* DateFormat.yMd 
* DateFormat.Md
This means we need a simple DateUtil that have two functions:
1. formatMonthDate
2. formatYearMonthDay
And use it whenever needs formatting.

lib/routes/user/home/payments_filter.dart
```diff
@@ -83,11 +88,28 @@ class PaymentsFilter extends StatelessWidget {
...
+  }
+
+  _buildDateFilterChip(DateTime startDate,DateTime endDate){
+    if(_startDate != null && _endDate != null){
+      return new Padding(padding: EdgeInsets.only(left: 16.0),child: Chip(label: Text(_formatFilterDateRange(_startDate,_endDate)),onDeleted: (){ _onFilterChanged(_filter,null,null);} ,));
+    }
+    return Container();
+  }
+
+  String _formatFilterDateRange(DateTime startDate,DateTime endDate) {
```
roeierez: This function is also duplicated in the pos and here. See my comment on CalendarFilter regarding DateUtils. it maybe belong there as well.

lib/routes/user/home/account_page.dart
```diff
@@ -19,7 +18,7 @@ import 'package:breez/logger.dart';
 
 const DASHBOARD_MAX_HEIGHT = 188.0;
 const DASHBOARD_MIN_HEIGHT = 70.0;
-const FILTER_MAX_SIZE = 70.0;
+var FILTER_MAX_SIZE = 70.0;
```
roeierez: Don't change this constant, if you need add another constant and use calculated getter to get the result you need based on the current context.

lib/routes/pos/transactions/pos_transactions_page.dart
```diff
@@ -53,68 +65,111 @@ class _PosTransactionsState extends State<_PosTransactionsPage> {
...
+          AccountModel account = snapshot.data;
+          return StreamBuilder<List<PaymentInfo>>(
+              stream: widget._accountBloc.paymentsStream,
+              builder: (context, snapshot) {
+                List<PaymentInfo> payments;
+                if (snapshot.hasData) {
+                  payments = snapshot.data;
+                  if(payments.isNotEmpty){
+                    DateTime _firstPaymentDate = DateTime.fromMillisecondsSinceEpoch(payments[payments.length-1].creationTimestamp.toInt() * 1000);
+                    _firstDate = new DateTime(_firstPaymentDate.year, _firstPaymentDate.month, _firstPaymentDate.day, 0, 0, 0, 0, 0);
```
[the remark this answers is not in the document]
erdemyerebasmaz: This will be addressed with https://breeztech.atlassian.net/browse/BU-120

PR #11, opened 2018-10-23
Created Calendar Filter dialog and DateUtils

erdemyerebasmaz:
I just see "3. Integration into pos and user UI views."
I've integrated some parts into POS that are almost identical.

### DIFF
--- a/lib/bloc/account/account_bloc.dart
+++ b/lib/bloc/account/account_bloc.dart
@@ -45,11 +45,10 @@ class AccountBloc {
   final _withdrawalResultController = new StreamController<String>.broadcast();
   Stream<String> get withdrawalResultStream => _withdrawalResultController.stream;
 
+  DateTime _firstDate;
   final _paymentsController = new BehaviorSubject<List<PaymentInfo>>();
   Stream<List<PaymentInfo>> get paymentsStream => _paymentsController.stream;
-  Stream<DateTime> get firstDate {
-    return paymentsStream.map( (payments) => DateTime.fromMillisecondsSinceEpoch(payments[0].creationTimestamp.toInt() * 1000));
-  }
+  DateTime get firstDate => this._firstDate == null ? DateTime(2018) : this._firstDate;
 
   Stream<List<PaymentInfo>> get receivedPayments {
     return paymentsStream.map( (payments) => payments.where( (p) => [PaymentType.DEPOSIT, PaymentType.RECEIVED].contains(p.type)));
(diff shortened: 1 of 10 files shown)


lib/bloc/account/account_bloc.dart
```diff
@@ -47,6 +47,10 @@ class AccountBloc {
 
   final _paymentsController = new BehaviorSubject<List<PaymentInfo>>();
   Stream<List<PaymentInfo>> get paymentsStream => _paymentsController.stream;
+  Stream<DateTime> get firstDate {
+    return paymentsStream.map( (payments) => DateTime.fromMillisecondsSinceEpoch(payments[0].creationTimestamp.toInt() * 1000));
+  }
+
```
erdemyerebasmaz: I'm fixing this to get first date from unfiltered payment list.

lib/widgets/calendar_dialog.dart
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  void _applyFilter() {
+    // Check if filter is unchanged
+    if (_startDate != widget.firstDate || _endDate.day != DateTime.now().day) {
+      DateTime startDate = DateTime(_startDate.year, _startDate.month, _startDate.day, 0, 0, 0);
+      DateTime endDate = DateTime(_endDate.year, _endDate.month, _endDate.day, 0, 0, 0);
+      // If selected dates are equal to each other, apply filter to the whole day
+      if (startDate.compareTo(endDate) == 0) {
+        endDate = DateTime(_endDate.year, _endDate.month, _endDate.day, 23, 59, 59, 999);
```
erdemyerebasmaz: We don't need this comparison, I'll just return startDate with beginning of the day(0, 0, 0, 000) and endDate with end of the day(23, 59, 59, 999);

lib/utils/date.dart
```diff
@@ -0,0 +1,20 @@
+import 'dart:io' show Platform;
+import "package:intl/intl.dart";
+import 'package:fixnum/fixnum.dart';
+
+class DateUtils {
+  static final DateFormat _monthDateFormat = new DateFormat.Md(Platform.localeName);
+  static final DateFormat _yearMonthDayFormat = new DateFormat.yMd(Platform.localeName);
+  static final DateFormat _transactionDateFormat = new DateFormat.yMd(Platform.localeName).add_jm();
```
roeierez: In DateUtils let's not use "transaction..." it should be general (like formatMonthDate, formatYearMonthDay, etc...)

lib/utils/date.dart
```diff
@@ -0,0 +1,20 @@
...
+
+  static String formatMonthDate(DateTime d) => _monthDateFormat.format(d);
+  static String formatYearMonthDay(DateTime d) => _yearMonthDayFormat.format(d);
+  static String formatPosTransactionDate(Int64 t) =>
```
roeierez: For consistency let's keep the parameter DateTime and not Int64.

PR #13, opened 2018-10-24
Payment filter model

erdemyerebasmaz:
Listening to paymentFilterStream
Applying payment filter
Change paymentsStream to return PaymentsModel


lib/bloc/account/account_bloc.dart
```diff
@@ -45,15 +45,12 @@ class AccountBloc {
...
-  Stream<List<PaymentInfo>> get paymentsStream => _paymentsController.stream;
+  final _paymentsController = new BehaviorSubject<PaymentsModel>();
+  Stream<PaymentsModel> get paymentsStream => _paymentsController.stream;
 
-  Stream<List<PaymentInfo>> get receivedPayments {
-    return paymentsStream.map( (payments) => payments.where( (p) => [PaymentType.DEPOSIT, PaymentType.RECEIVED].contains(p.type)));
-  }
-  Stream<List<PaymentInfo>> get sentPayments {
-    return paymentsStream.map( (payments) => payments.where( (p) => [PaymentType.WITHDRAWAL, PaymentType.SENT].contains(p.type)));
-  }
```
erdemyerebasmaz: Do we need or use these streams?

lib/bloc/account/account_bloc.dart
```diff
@@ -200,19 +199,67 @@ class AccountBloc {
...
+         }
+         _paymentsController.add(PaymentsModel(_filterPayments(payments), _paymentFilterController.value, _firstDate));
       })
-      .catchError(_paymentsController.addError); 
-    }  
-  
+      .catchError(_paymentsController.addError);
+    }
+
+  _filterMockDataPayments(List<PaymentInfo> paymentsList) {
```
roeierez: We don't need this function. use the same function _filterPayments for the real data and mock data.
You will need to change its prototype to get List<PaymentInfo> but it worth the value of removing this function and unifying the flows.

lib/routes/user/home/account_page.dart
```diff
@@ -49,6 +49,8 @@ class _AccountPageState extends State<_AccountPage> {
   StreamSubscription<AccountModel> _statusSubscription;
   String _filter;  
   String _paymentRequestInProgress;
+  PaymentFilterModel _paymentFilterModel;
```
roeierez: don't save this value in the class, use it locally in the stream in the build function.

lib/routes/user/home/account_page.dart
```diff
@@ -49,6 +49,8 @@ class _AccountPageState extends State<_AccountPage> {
...
+  DateTime _firstDate;
```
roeierez: Also here, don't save.

PR #14, opened 2018-10-25
Calendar filter ui changes

erdemyerebasmaz:
Implemented communication between UI elements for applying filter
Show date filter chips above transaction lists

### DIFF
--- a/lib/bloc/account/account_bloc.dart
+++ b/lib/bloc/account/account_bloc.dart
@@ -211,7 +211,7 @@ class AccountBloc {
       if (MockPaymentInfo.isMockData) {
         List<PaymentInfo> _paymentsList = _filterPayments(MockPaymentInfo.createMockData());
         if(_paymentsList.length > 0){
-          _firstDate = DateTime.fromMillisecondsSinceEpoch(_paymentsList.elementAt(_paymentsList.length-1).creationTimestamp.toInt() * 1000);
+          _firstDate = DateTime.fromMillisecondsSinceEpoch(_paymentsList.last.creationTimestamp.toInt() * 1000);
         }
         _paymentsController.add(PaymentsModel(_paymentsList, _paymentFilterController.value, _firstDate ?? DateTime(DateTime.now().year)));
         return;
@@ -220,7 +220,7 @@ class AccountBloc {
       breezLib.getPayments().then( (payments) {
         List<PaymentInfo> _paymentsList =  payments.paymentsList.map((payment) => new PaymentInfo(payment, _currentUser.currency)).toList();
         if(_paymentsList.length > 0){
-          _firstDate = DateTime.fromMillisecondsSinceEpoch(_paymentsList.elementAt(_paymentsList.length-1).creationTimestamp.toInt() * 1000);
+          _firstDate = DateTime.fromMillisecondsSinceEpoch(_paymentsList.last.creationTimestamp.toInt() * 1000);
         }
         _paymentsController.add(PaymentsModel(_filterPayments(_paymentsList), _paymentFilterController.value, _firstDate ?? DateTime(DateTime.now().year)));
       })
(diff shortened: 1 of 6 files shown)


lib/bloc/account/account_bloc.dart
```diff
@@ -220,7 +220,7 @@ class AccountBloc {
       breezLib.getPayments().then( (payments) {
         List<PaymentInfo> _paymentsList =  payments.paymentsList.map((payment) => new PaymentInfo(payment, _currentUser.currency)).toList();
         if(_paymentsList.length > 0){
-          _firstDate = DateTime.fromMillisecondsSinceEpoch(_paymentsList.elementAt(0).creationTimestamp.toInt() * 1000);
+          _firstDate = DateTime.fromMillisecondsSinceEpoch(_paymentsList.elementAt(_paymentsList.length-1).creationTimestamp.toInt() * 1000);
```
roeierez:
```suggestion
          _firstDate = DateTime.fromMillisecondsSinceEpoch(_paymentsList.last.creationTimestamp.toInt() * 1000);
```

lib/routes/pos/transactions/pos_transactions_page.dart
```diff
@@ -55,73 +55,103 @@ class _PosTransactionsState extends State<_PosTransactionsPage> {
...
-        body: StreamBuilder<AccountModel>(
+    return StreamBuilder<AccountModel>(
             stream: widget._accountBloc.accountStream,
             builder: (context, snapshot) {
               AccountModel account = snapshot.data;
               return StreamBuilder<PaymentsModel>(
                   stream: widget._accountBloc.paymentsStream,
                   builder: (context, snapshot) {
-                    List<PaymentInfo> payments = new List<PaymentInfo>();
+                    PaymentsModel paymentsModel = new PaymentsModel.initial();
```
roeierez: Don't create models in the UI, they are only created in the bloc and retrieved in the stream. If you get null from the stream it has meaning (the bloc hasn't been initialized yet)

lib/routes/pos/transactions/pos_transactions_page.dart
```diff
@@ -55,73 +55,103 @@ class _PosTransactionsState extends State<_PosTransactionsPage> {
...
                     if (snapshot.hasData) {
-                      payments = snapshot.data.paymentsList ?? new List<PaymentInfo>();
-                      _firstDate = snapshot.data.firstDate ?? DateTime(2018);
+                      paymentsModel = snapshot.data;
                     }
 
-                    if (account == null || payments == null) {
+                    if (account == null || paymentsModel == null) {
```
roeierez: Here it is also the reason why you shouldn't create a model in this case, here explicitly there is a check for null, which is good but will never pass because of the previous line that creates a model.

lib/routes/pos/transactions/pos_transactions_page.dart
```diff
@@ -55,73 +55,103 @@ class _PosTransactionsState extends State<_PosTransactionsPage> {
...
-  Widget _buildTransactions(List<PaymentInfo> payments) {
+  Widget _calendarButton(DateTime firstDate) {
+    return Padding(padding: EdgeInsets.only(right: 16.0), child:
+    IconButton(icon: ImageIcon(AssetImage("src/icon/calendar.png"), color: Colors.white, size: 24.0),
+        onPressed: () =>
+            showDialog(
+                context: context,
+                builder: (_) => CalendarDialog(context, firstDate)).then(((result) =>
+                widget._accountBloc.paymentFilterSink.add(PaymentFilterModel(
+                    [PaymentType.RECEIVED, PaymentType.DEPOSIT, PaymentType.SENT, PaymentType.WITHDRAWAL
```
roeierez: Here you reset the payment type filter, don't you want to take the current filter and just use copyWith function with the dates?

PR #15, opened 2018-10-26
Receive/send payment to another LN wallet

icota:


lib/routes/user/create_invoice/create_invoice_page.dart
```diff
@@ -0,0 +1,226 @@
...
+  @override
+  State<StatefulWidget> createState() {
+    return new _CreateInvoiceState();
+  }
+}
+
+class _CreateInvoiceState extends State<_CreateInvoicePage> {
+  final _formKey = GlobalKey<FormState>();
+  String _errorMessage = "";
+  String _bolt11 = "";
```
roeierez: I think this widget can be stateless as we don't need all the state variables:
currency can be taken from AccountStream that already listened to in the build function
_bolt11 can also be taken using StreamBuilder in for the readyInvoiceStream in the build function.
This way we can also save the subscriptions.
In general subscriptions in state are only needed when we need to do action that render elements like dialogs and popups that can't be expressed as part of the build function and stateless widget should only be preferred.

PR #27, opened 2018-11-07
Subswap ui integration

roeierez:
This PR integrate all the UI for refund submarine swaps addresses.

### DIFF
--- a/lib/routes/user/get_refund/get_refund_page.dart
+++ b/lib/routes/user/get_refund/get_refund_page.dart
@@ -80,9 +80,9 @@ class _GetRefundState extends State<GetRefund> {
                                 padding: const EdgeInsets.all(16.0),
                                 child: SizedBox(
                                     height: 36.0,
-                                    width: 140.0,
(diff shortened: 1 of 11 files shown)


lib/routes/user/get_refund/get_refund_page.dart
```diff
@@ -85,9 +80,9 @@ class _GetRefundState extends State<GetRefund> {
                                 padding: const EdgeInsets.all(16.0),
                                 child: SizedBox(
                                     height: 36.0,
-                                    width: 140.0,
+                                    width: 145.0,
                                     child: SubmitButton(
-                                        "REFUND",
+                                        item.refundBroadcasted ? "BROADCASTED" : "REFUND",
```
icota: Maybe SENT is a better choice of word, just a suggestion

roeierez: Technically it is for the state where the transaction is broadcasted and not yet confirmed, hence the name. I will ask Roy for his take here.

PR #29, opened 2018-11-15
communication based on encrypted state exhcnage

roeierez:
This PR integrates the doubleratchet APi in breez library into the CTP sessions.

### DIFF
--- a/lib/bloc/connect_pay/encryption.dart
+++ b/lib/bloc/connect_pay/encryption.dart
@@ -1,0 +1,24 @@
+import 'package:breez/services/breezlib/breez_bridge.dart';
+
+abstract class MessageInterceptor {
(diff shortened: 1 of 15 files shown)


lib/bloc/connect_pay/encryption.dart
```diff
@@ -0,0 +1,24 @@
+import 'package:breez/services/breezlib/breez_bridge.dart';
+
+abstract class MessageInterceptor {
+  Future<String> transformOutgoingMessage(String message);
+  Future<String> transformIncoingMessage(String message);
```
icota: Typo: incoing

roeierez: thanks, fixed.

PR #31, opened 2018-11-19
Add Note POS - Pay Someone Nearby

erdemyerebasmaz:
Added description to POS and Pay Someone Nearby Invoices
Merged addStandardInvoice-addInvoice into one.


lib/bloc/invoice/invoice_bloc.dart
```diff
@@ -36,6 +33,7 @@ class InvoiceBloc {
   Stream<bool> get paidInvoicesStream => _paidInvoicesController.stream;
 
   Int64 payBlankAmount = Int64(-1);
+  String description = "";
```
roeierez: What is the purpose of this "description" field? I also don't see it assigned anywhere.

lib/routes/user/pay_nearby/pay_nearby_page.dart
```diff
@@ -58,6 +59,7 @@ class PayNearbyState extends State<_PayNearbyPage> {
                     if (_formKey.currentState.validate()) {
                       _formKey.currentState.save();
                       widget._invoiceBloc.payBlankAmount = _amountToSendSatoshi;
+                      widget._invoiceBloc.description = _descriptionController.text;
```
erdemyerebasmaz: > What is the purpose of this "description" field? I also don't see it assigned anywhere.

It is assigned here.

roeierez: It is wrong to assign a bloc variable from outside. Everything in the bloc should be private except for the stream interfaces.
I see also previously we had payBlankAmount behaves like this. Let's fix it.

erdemyerebasmaz:
Use sinks to assign payBlankAmount and description.

PR #33, opened 2018-11-20
Pubspec and gradle improvements

icota:
- build.gradle: suppress warnings by using the 'implementation' directive
- pubspec: don't use git versions od firebase plugins
- Bump google-services to 4.2.0
- build.gradle: don't hardcode dependencies


icota:
Tested working with CTP

PR #42, opened 2018-11-27
Drive backup

icota:


android/app/src/main/java/com/breez/client/plugins/breez/BreezBackup.java
```diff
@@ -0,0 +1,362 @@
...
+        }
+        if (call.method.equals("listBackupFolders")) {
+            listBackupFolders();
+        }
+        if (call.method.equals("restore")) {
+            if (m_driveResourceClient == null) {
+                promptToAuthorize();
+            }
+            else {
+                m_result = result;
```
roeierez: The nature of the methods in this channel is asynchronous. Therefore saving the m_result as class variable and use it will mix different requests with different results. We should pass the result to the method that is executed to be sure the request/response matches.

android/app/src/main/java/com/breez/client/plugins/breez/BreezBackup.java
```diff
@@ -0,0 +1,362 @@
...
+    private void promptToAuthorize() {
+        Intent signInIntent = m_googleSignInClient.getSignInIntent();
+        m_activity.startActivityForResult(signInIntent, AUTHORIZE_ACTIVITY_REQUEST_CODE);
+    }
+
+    @Override
+    public void onMethodCall(MethodCall call, MethodChannel.Result result) {
+        if (call.method.equals("authorize")) {
+            if (m_driveResourceClient == null) {
+                promptToAuthorize();
```
roeierez: Here is a case where we can save the m_result because it invokes a separate activity and we can use the m_result to success/fail in the "onActivityResult" callback depends on the status so the flutter will get the right result.

android/app/src/main/java/com/breez/client/plugins/breez/BreezBackup.java
```diff
@@ -0,0 +1,362 @@
...
+                    listBackupFolders();
+                } else {
+                    getNodeIdFolder(nodeId);
+                }
+            }
+        }
+    }
+
+    private void notifyBackupDisabled() {
+        m_eventsListener.success(false);
```
roeierez: Observing the use of these methods (notifyBackupDisabled, notifyBackupEnabled) I saw they server as mechanism to notify backup success/failure. I think this information should be retrieved from the result of the backup/restore methods without the need for the event channel.
If notification stream is needed for the UI the flutter side can manage it.

android/app/src/main/java/com/breez/client/plugins/breez/BreezBackup.java
```diff
@@ -0,0 +1,362 @@
...
+                            if (backupsMap.values().size() == 1) {
+                                backupsMap.forEach((k, v) -> getNodeIdFolder(k));
+                            }
+                            else {
+                                m_result.success(backupsMap);
+                            }
+                        })
+                    .addOnFailureListener(m_activity, e -> {
+                        Log.e(TAG, "Error retrieving files", e);
+                        m_result.success(false);
```
roeierez: why not m_result.error ?

android/app/src/main/java/com/breez/client/plugins/breez/BreezBackup.java
```diff
@@ -0,0 +1,362 @@
...
+                            Log.w(TAG, metadataBuffer.toString());
+                            if (metadataBuffer.getCount() == 0) {
+                                m_result.success(false);
+                            }
+
+                            fetchBackupFiles(metadataBuffer.get(0).getDriveId().asDriveFolder());
+                        })
+                    .addOnFailureListener(m_activity, e -> {
+                        Log.e(TAG, "Error retrieving files", e);
+                        m_result.success(false);
```
roeierez: m_result.error ?

android/app/src/main/java/com/breez/client/plugins/breez/BreezBackup.java
```diff
@@ -0,0 +1,362 @@
...
+                metadataBuffer -> {
+                    for (Metadata m : metadataBuffer) {
+                        m_driveResourceClient.openFile(m.getDriveId().asDriveFile(), DriveFile.MODE_READ_ONLY)
+                                .addOnSuccessListener(contents -> {
+                            DriveContents fileContents = contents;
+                            InputStream remoteFileInputStream = fileContents.getInputStream();
+
+                            File file = new File(m_activity.getCacheDir().getPath() + "/" + m.getTitle());
+                            FileOutputStream outputStream = null;
+                            try {
```
roeierez: instead of wrapping try/catch block many times, better to have one try/catch block that wraps the whole block of code. Also if the outputstream is not initialized correctly because of exception, there is no use to move on

android/app/src/main/java/com/breez/client/plugins/breez/BreezBackup.java
```diff
@@ -0,0 +1,362 @@
...
+        for (String path: paths) {
+            m_driveResourceClient.createContents()
+                .continueWithTask(task -> {
+                    DriveContents contents = task.getResult();
+                    OutputStream outputStream = contents.getOutputStream();
+
+                    File file = new File(path);
+                    FileInputStream inputStream = new FileInputStream(file);
+
+                    byte[] buffer = new byte[inputStream.available()];
```
roeierez: available function doesn't return all the bytes in the stream (see here: https://docs.oracle.com/javase/7/docs/api/java/io/FileInputStream.html#available())
Reading from InputStream in java requires a loop until read method returns -1.
Something like this:
```
do {
    byte[] buf = new byte[1024];
     i = inputStream.read(buf);               
} while (i != -1);
```

lib/bloc/account/account_bloc.dart
```diff
@@ -74,34 +74,44 @@ class AccountBloc {
...
+      _listenWithdrawalRequests(_breezLib);
+      _listenSentPayments(_breezLib);
+      _listenFilterChanges(_breezLib);
+      _listenAccountChanges(_breezLib);
+      _listenPOSFundingRequests(server, _breezLib);
+      _listenMempoolTransactions(_device, notificationsService, _breezLib);
+      _listenRoutingNodeConnectionChanges(_breezLib);
+    }
+
+    void startLightning() {
```
roeierez: Suggest to make this method private so we will be sure that is called only from the account bloc

lib/bloc/app_blocs.dart
```diff
@@ -23,14 +25,16 @@ class AppBlocs {
     POSProfileBloc posProfileBloc = new POSProfileBloc();    
     InvoiceBloc invoicesBloc = new InvoiceBloc();
     ConnectPayBloc connectPayBloc = new ConnectPayBloc(userProfileBloc.userStream, accountBloc.accountStream);
-    
+    BackupBloc backupBloc = new BackupBloc(accountBloc);
```
roeierez: The BackupBloc only needs the account stream so it is better to pass it instead of the accountBloc

lib/bloc/backup/backup_bloc.dart
```diff
@@ -0,0 +1,144 @@
...
+            getApplicationDocumentsDirectory().then((appDir) {
+              breezLib.copyBreezConfig(appDir.path).then((done) {
+                breezLib.bootstrapFiles(appDir.path, new List<String>.from(restoreResult)).then((done) {
+                  _accountBloc.startLightning();
+                });
+              });
+            });
+          });
+        }
+        else if (restoreResult is bool) {
```
roeierez: Better to stay with on type for the method return type and raise error from the API instead.

lib/widgets/enable_backup_dialog.dart
```diff
@@ -0,0 +1,83 @@
...
+    return new EnableBackupDialogState();
+  }
+}
+
+class EnableBackupDialogState extends State<EnableBackupDialog> {
+  bool _isChecked = true;
+
+  @override
+  Widget build(BuildContext context) {
+    return showEnableBackupDialog();
```
roeierez: Dialog shouldn't be shown from the build function because every trigger to build again will prompt with another dialog... 
We can just leave a function here: 
"createEnableBackupDialog" which returns an AlertDialog and then when needed just call:
showDialog(createEnableBackupDialog())

lib/widgets/restore_dialog.dart
```diff
@@ -0,0 +1,82 @@
...
+  @override
+  RestoreDialogState createState() {
+    return new RestoreDialogState();
+  }
+}
+
+class RestoreDialogState extends State<RestoreDialog> {
+  @override
+  Widget build(BuildContext context) {
+    return showRestoreDialog();
```
roeierez: Same here for the dialog from build function

lib/widgets/enable_backup_dialog.dart
roeierez: Looking again, actually this is OK since it only returns a widget. Taking back the feedback. sorry.

android/app/src/main/java/com/breez/client/plugins/breez/BreezBackup.java
```diff
@@ -0,0 +1,362 @@
...
+        if (call.method.equals("authorize")) {
+            if (m_driveResourceClient == null) {
+                promptToAuthorize();
+            }
+            else {
+                m_result.success(true);
+            }
+        }
+        if (call.method.equals("backup")) {
+            List<String> paths = call.argument("paths");
```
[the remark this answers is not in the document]
roeierez: @icota you don't think it is a good idea to authorize here if needed as you do for restore?

lib/bloc/account/account_bloc.dart
```diff
@@ -72,36 +72,56 @@ class AccountBloc {
...
-       _listenRefundBroadcasts(breezLib);
+      _listenStartLightningRequests(breezLib, device);
+    }
+
+    void _startLightning() {
+
+    }
+
+    void _listenStartLightningRequests(BreezBridge breezLib, Device device) {
+      _startLightningController.stream.listen((start) {
```
roeierez: There are four places that pushing to this sink in attempt to start the lightning lib. I think it makes it complicated to track. Perhaps a better way is to just to start the lightningLIb in the account bloc when the user is registered. This I think will cover all the cases (startup with/without restore, startup with already registered user ..)
And it will be a single point of decision without need for the UI to do it from different widgets.
Also it is simple to implement because the account bloc already tracks the user stream.

android/app/src/main/java/com/breez/client/plugins/breez/BreezBackup.java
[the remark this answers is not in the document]
icota: @roeierez not sure what you mean. I now authorize on every call if needed:
`    @Override
    public void onMethodCall(MethodCall call, MethodChannel.Result result) {
        if (m_driveResourceClient == null) {
            m_result = result;
            m_call = call;
            promptToAuthorize();
        }
        else {
            handleMethodCall(call, result);
        }
    }`

lib/bloc/backup/backup_bloc.dart
```diff
@@ -0,0 +1,127 @@
...
+      backup();
+    });
+  }
+
+  _listenBackupPaths(BreezBridge breezLib, Future<SharedPreferences> sharedPrefrences) {
+    Observable(breezLib.notificationStream).where((event) {
+      return event.type ==
+          NotificationEvent_NotificationType.BACKUP_FILES_AVAILABLE;
+    }).listen((event) {
+      _currentBackupPaths = event.data;
```
roeierez: Better to pas the backup paths directly to the backup function instead of keeping them in a class variable. It eliminates all sorts of races.

lib/bloc/backup/backup_bloc.dart
```diff
@@ -0,0 +1,127 @@
...
+              });
+            });
+          });
+        }
+        else {
+          // We need to pick from different node IDs
+          _multipleRestoreController.add(new Map<String, String>.from(restoreResult));
+        }
+      }).catchError((error) {
+        // Notify user somehow?
```
roeierez: perhaps we should add error to one of the streams and show error in the restore dialog UI?

android/app/src/main/java/com/breez/client/plugins/breez/BreezBackup.java
```diff
@@ -0,0 +1,333 @@
...
+
+                            try {
+                                outputStream = new FileOutputStream(file);
+                            } catch (FileNotFoundException e) {
+                                result.error("CACHE_FILE_FAILURE", "Couldn't find cache file to save to", null);
+                            }
+
+                            try {
+                                int i;
+                                do {
```
roeierez: When the result of "remoteFileInputStream.read(buf)" returns negative we can't write to the outputStream. Here is a suggestion:
```suggestion
                               byte[] buf = new byte[1024];
                               while (i = remoteFileInputStream.read(buf) > 0) {                                  
                                    outputStream.write(buf, 0, i);
                                }
```

lib/bloc/account/account_bloc.dart
```diff
@@ -177,9 +171,22 @@ class AccountBloc {
...
 
-    _listenUserChanges(Stream<BreezUserModel> userProfileStream, BreezBridge breezLib){
-      userProfileStream.listen((user) { 
+    _listenUserChanges(Stream<BreezUserModel> userProfileStream, BreezBridge breezLib, Device device){
+      userProfileStream.listen((user) {
         _currentUser = user;
+
+        if (user.registered && !_startedLightning) {
+          breezLib.bootstrap().then((done) {
+            _startedLightning = true;
```
roeierez: I am afraid of race conditions here when two subsequent user events are emitted fast. In that case we might end up trying to call the start lightning flow twice. Maybe just move the line: "_startedLightning = true;" right after the "if" ?

icota: Good catch, fixed.

PR #55, opened 2019-01-27
Bitrefill widget integration

erdemyerebasmaz:


lib/routes/shared/dev/dev.dart
```diff
@@ -282,4 +287,74 @@ class DevViewState extends State<DevView> {
...
+                        .pushReplacementNamed("/marketplace");
+                  }
+                },
+              ),
+            ],
+          );
+        });
+  }
+
+  bool _validatePassword(String value) {
```
erdemyerebasmaz: (!)

erdemyerebasmaz: !

lib/routes/shared/dev/dev.dart
```diff
@@ -282,4 +287,74 @@ class DevViewState extends State<DevView> {
...
+                        .of(_scaffoldKey.currentState.context)
+                        .pushReplacementNamed("/marketplace");
+                  }
+                },
+              ),
+            ],
+          );
+        });
+  }
+
```
erdemyerebasmaz: !

lib/routes/shared/dev/dev.dart
```diff
@@ -282,4 +287,74 @@ class DevViewState extends State<DevView> {
...
+  bool _validatePassword(String value) {
+    if (value == "135642") {
```
roeierez: Please put this password (different one now) in breez.conf file.

lib/routes/user/embedded_widgets/bitrefill_widget.dart
```diff
@@ -0,0 +1,83 @@
...
+    // TODO: implement dispose
+    super.dispose();
+    _widgetWebview.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    InvoiceBloc invoiceBloc = AppBlocsProvider.of<InvoiceBloc>(context);
+    if (_lightningLink != null) {
+      invoiceBloc.newLightningLinkSink.add(_lightningLink);
```
roeierez: triggering events (including adding to sink) from the build method is not the way to do it. build method is only for rendering. Instead it is better to push the link to the sink at the same moment we get it from the webview  lightningLinkStream.

lib/routes/user/embedded_widgets/bitrefill_widget.dart
```diff
@@ -0,0 +1,83 @@
...
+class BitrefillPageState extends State<BitrefillPage> {
+  final _widgetWebview = new FlutterWebviewPlugin();
+  String _lightningLink;
+
+  @override
+  void initState() {
+    super.initState();
+    _lightningLink = null;
+
+    _widgetWebview.lightningLinkStream.listen((value) {
```
roeierez: lightningLinkStream is not a good name, it is lightning specific inside a generic webview. Maybe a better name that reflects the purpose of this stream may be "postMessageStream" ?

lib/routes/user/embedded_widgets/bitrefill_widget.dart
```diff
@@ -0,0 +1,83 @@
...
+        iconTheme: theme.appBarIconTheme,
+        textTheme: theme.appBarTextTheme,
+        backgroundColor: Color.fromRGBO(5, 93, 235, 1.0),
+        title: new Text(
+          widget._title,
+          style: theme.appBarTextStyle,
+        ),
+        elevation: 0.0,
+      ),
+      url: 'https://www.bitrefill.com/embed/lightning/?apiKey=GAj4sWRVqK3Uau1L&hideQr',
```
roeierez: the apiKey should go into breez.conf

lib/routes/user/marketplace/marketplace.dart
```diff
@@ -0,0 +1,72 @@
...
+          textTheme: theme.appBarTextTheme,
+          backgroundColor: Color.fromRGBO(5, 93, 235, 1.0),
+          title: new Text(
+            widget._title,
+            style: theme.appBarTextStyle,
+          ),
+          elevation: 0.0,
+        ),
+        body: ListView(
+          children: <Widget>[
```
roeierez: It will be nice to have the ListItem (Bitrefill in this case) rendered in its own widget, we can call it "Seller".
Later we will probably have a "MarketPlaceBloc" where all the sellers will be fetched from there. Right not we don't have to implement the bloc but as soon as we have a new seller we should do it.

lib/routes/user/embedded_widgets/bitrefill_widget.dart
```diff
@@ -0,0 +1,92 @@
...
+        ),
+        elevation: 0.0,
+      ),
+      url: 'https://www.bitrefill.com/embed/lightning/?apiKey=$_apiKey&hideQr',
+      withJavascript: true,
+      withZoom: false,
+    );
+  }
+
+  Future _getApiKey() async {
```
roeierez: This still bothers me.. the fact we parse the configuration to get the API key and concat it to the URL here in the UI. This is really business logic that indicates on a missing bloc.
The only thing that we have at this page that is bitrefill specific is the url and the name.
I suggest to do the following:

- Create a model called: SellerModel with the following props: (url, name)
- Create a bloc (very small for now) called: MarketplaceBloc with the a stream called sellersStream. All the config parsing and generating the right url for the UI will be done inside the bloc
- When you show the list of sellers in the Marketplace route use a StreamBuilder to read from the new bloc stream
- The BitRefill route can be changed to a generic "WebSellerRoute" that will get the SellerModel as context and render a webview, listen to post messages and redirect the invoice to the home page as it does now.

I think these changes puts the building blocs in the right place for the near future.

erdemyerebasmaz: I've completed these steps but there are a few static things left.

- How can we read vendors(sellers) api keys in a generic manner?
- How will we initialize vendor list of MarketplaceBloc, where will the vendor data come from?

lib/bloc/marketplace/marketplace_bloc.dart
```diff
@@ -0,0 +1,37 @@
...
+          "Bitrefill", bgColor: Color(0xFF3e99fa))
+    ];
+    _vendorController.add(vendorList);
+  }
+
+  Future<Config> _readConfig() async {
+    String lines = await rootBundle.loadString('conf/breez.conf');
+    return Config.fromString(lines);
+  }
+
```
erdemyerebasmaz: Todo:
Implement _generateUrl(String apiKey),
Move setting and reading marketplacePassword to another bloc. e.g. accountBloc

lib/bloc/marketplace/vendor_row.dart
```diff
@@ -0,0 +1,51 @@
...
+              border: Border.all(
+                  color: Colors.white, style: BorderStyle.solid, width: 1.0),
+              borderRadius: BorderRadius.circular(14.0)),
+          child: Row(
+            mainAxisAlignment: MainAxisAlignment.center,
+            crossAxisAlignment: CrossAxisAlignment.center,
+            children: <Widget>[
+              _vendorLogo,
+              Text(_vendor.name, style: theme.vendorTitleStyle),
+            ],
```
erdemyerebasmaz: Instead of using Image and Text field separately, use and image that has logo and the vendor name.
This way, we won't need to deal with text styles for each vendor.

lib/routes/shared/dev/dev.dart
```diff
@@ -282,4 +291,83 @@ class DevViewState extends State<DevView> {
...
+  }
+
+  Future _getMarketplacePassword() async {
+    String configString = await rootBundle.loadString('conf/breez.conf');
+    Config config = Config.fromString(configString);
+    setState(() {
+      _marketplacePassword =
+          config.get("Application Options", "marketplacepassword");
+    });
+  }
```
erdemyerebasmaz: @roeierez Where should we move this? AccountBloc?

lib/routes/user/embedded_widgets/bitrefill_widget.dart
roeierez: > * How should we read vendors(sellers) api keys in a generic manner?

every vendor will have his config properties prefixed by its name. e.g:
bitrefill.apikey=XXX
We can also use a different config file for the marketplace and show the marketplace UI only if this config file exists.

lib/routes/shared/dev/dev.dart
roeierez: > Where should we move this? AccountBloc?

No, to me it feels the right place.

lib/bloc/marketplace/marketplace_bloc.dart
erdemyerebasmaz: :+1:

lib/bloc/invoice/invoice_bloc.dart
```diff
@@ -111,6 +115,7 @@ class InvoiceBloc {
     })
     .asyncMap((paymentRequest) async { 
       //filter out our own payment requests
+      _readyInvoicesController.add(paymentRequest);
```
roeierez: Why it is needed to add the payment request to the _readyInvoicesController?

lib/routes/user/marketplace/marketplace.dart
```diff
@@ -0,0 +1,80 @@
...
+}
+
+class MarketplacePageState extends State<MarketplacePage> {
+  final String _title = "Marketplace";
+  final GlobalKey<ScaffoldState> _scaffoldKey = new GlobalKey<ScaffoldState>();
+  MarketplaceBloc _marketplaceBloc;
+  bool _isInit = false;
+
+  @override
+  void initState() {
```
roeierez: This method override can be removed since it only call super.initState()

lib/routes/user/marketplace/marketplace.dart
```diff
@@ -0,0 +1,80 @@
...
+    super.didChangeDependencies();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return StreamBuilder(
+        stream: _marketplaceBloc.vendorsStream,
+        builder: (context, snapshot) {
+          List<VendorModel> vendorsModel;
+          if (snapshot.hasData) {
```
roeierez: no need for the "if" statement. Just an assignment like this should work:
`List<VendorModel> vendorsModel = snapshot.data`

lib/routes/user/marketplace/marketplace.dart
```diff
@@ -0,0 +1,80 @@
...
+        });
+  }
+
+  Widget _buildScaffold(Widget body) {
+    return new Scaffold(
+      key: _scaffoldKey,
+      appBar: new AppBar(
+        iconTheme: theme.appBarIconTheme,
+        textTheme: theme.appBarTextTheme,
+        backgroundColor: Color.fromRGBO(5, 93, 235, 1.0),
```
roeierez: Can we add this to the theme?

lib/routes/user/marketplace/vendor_webview.dart
```diff
@@ -0,0 +1,89 @@
...
+        Navigator.of(context).pop();
+      }
+    });
+  }
+
+  @override
+  void didChangeDependencies() {
+    InvoiceBloc invoiceBloc = AppBlocsProvider.of<InvoiceBloc>(context);
+
+    _postMessageListener = _widgetWebview.onPostMessage.listen((postMessage) {
```
roeierez: didChangeDependencies may be called multiple times during the widget lifecycle. You should make sure the _postMessageListener is only initialized once and can do that by using an _init flag.

lib/routes/user/marketplace/vendor_webview.dart
```diff
@@ -0,0 +1,89 @@
...
+      if (postMessage != null) {
+        final order = JSON.jsonDecode(postMessage);
+        invoiceBloc.newLightningLinkSink.add(order['uri']);
+        _widgetWebview.hide();
+        Navigator.of(context).push(createLoaderRoute(context));
```
roeierez: Why do we need the loader here? when is it popped?

lib/routes/user/marketplace/vendor_webview.dart
```diff
@@ -0,0 +1,89 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    return new WebviewScaffold(
+      appBar: new AppBar(
+        leading: backBtn.BackButton(),
+        automaticallyImplyLeading: false,
+        iconTheme: theme.appBarIconTheme,
+        textTheme: theme.appBarTextTheme,
+        backgroundColor: Color.fromRGBO(5, 93, 235, 1.0),
```
roeierez: Can you add this to the theme please?

PR #59, opened 2019-02-03
Show Bitrefill logo on payments list

erdemyerebasmaz:
Show Bitrefill logo on payments list
Split description 'Bitrefill + order id' into title and description

### DIFF
--- a/android/key.properties
+++ b/android/key.properties
@@ -1,4 +1,0 @@
-keyAlias=foo
-keyPassword=foo
-storeFile=foo
-storePassword=foo
--- a/android/key.properties
+++ b/android/key.properties
@@ -1,0 +1,4 @@
+keyAlias=foo
+keyPassword=foo
+storeFile=foo
+storePassword=foo
--- a/lib/bloc/account/account_model.dart
+++ b/lib/bloc/account/account_model.dart
@@ -206,14 +206,25 @@ class PaymentInfo {
 
   bool get isTransferRequest => _paymentResponse?.invoiceMemo?.transferRequest == true;
 
-  String get description => type == PaymentType.DEPOSIT || type == PaymentType.WITHDRAWAL ? "Bitcoin Transfer" : _paymentResponse.invoiceMemo?.description;
-  
-  String get imageURL {    
-    String url = (type == PaymentType.SENT ? _paymentResponse.invoiceMemo?.payeeImageURL : _paymentResponse.invoiceMemo?.payerImageURL);
-    return (url == null || url.isEmpty) ? null : url;      
+  String get description =>
+      _paymentResponse.invoiceMemo.description.startsWith("Bitrefill")
+          ? _paymentResponse.invoiceMemo.description
+          .substring(10, _paymentResponse.invoiceMemo.description.length)
+          : type == PaymentType.DEPOSIT || type == PaymentType.WITHDRAWAL
+          ? "Bitcoin Transfer"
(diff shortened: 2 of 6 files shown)


android/key.properties
```diff
@@ -0,0 +1,4 @@
+keyAlias=foo
```
roeierez: Is this relevant? Or commited by mistake?

erdemyerebasmaz: Committed by mistake. I had it on my local .gitignore before it got squashed by pubspec update :)

lib/routes/user/home/payment_item.dart
```diff
@@ -15,6 +15,9 @@ class PaymentItem extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     String title = _paymentInfo.title;
+    if (title.contains("Bitrefill")) {
```
roeierez: Better to use `startsWith` then `contains`

kingonly: Are you sure? I'd keep contains

roeierez: It is a temporary solution and my thinking is to provide mistakes here (if someone enters Bitrefill in the text...).
Anyway If we want to be more "loose" here then we need to change the next line as well to:
`title = "Bitrefill"` instead of `title=title.substring(0,10)`

kingonly: I'm good either way

erdemyerebasmaz: > It is a temporary solution and my thinking is to provide mistakes here (if someone enters Bitrefill in the text...).
> Anyway If we want to be more "loose" here then we need to change the next line as well to:
> `title = "Bitrefill"` instead of `title=title.substring(0,10)`
I've replaced contains with startsWith and set title to "Bitrefill".

roeierez: Did a full pass over the code and since all the changes are regarding the description (title) and image I believe a better strategy is to do these two small changes in the PaymentInfo model itself. We can add these manipulations inside both `imageURL` and `description` getters and I think all these changes in the UI won't be needed.
Maybe a small change will need to be done in BreezAvatar.
After all when we will do the generic solution the usage will be like this (coming from the model) and we won't need to change the UI code.
What do you think @erdemyerebasmaz ?

erdemyerebasmaz: Yes, it makes sense to move data manipulations out of UI level.
I've moved them to PaymentInfo model and added _VendorAvatar to BreezAvatar for vendors.

lib/bloc/account/account_model.dart
```diff
@@ -206,14 +206,25 @@ class PaymentInfo {
 
   bool get isTransferRequest => _paymentResponse?.invoiceMemo?.transferRequest == true;
 
-  String get description => type == PaymentType.DEPOSIT || type == PaymentType.WITHDRAWAL ? "Bitcoin Transfer" : _paymentResponse.invoiceMemo?.description;
-  
-  String get imageURL {    
-    String url = (type == PaymentType.SENT ? _paymentResponse.invoiceMemo?.payeeImageURL : _paymentResponse.invoiceMemo?.payerImageURL);
-    return (url == null || url.isEmpty) ? null : url;      
+  String get description =>
```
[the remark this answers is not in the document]
erdemyerebasmaz: It's not a must but the payment detail dialog looks better this way.

PR #67, opened 2019-02-15
Freebitcoins UI Integration

erdemyerebasmaz:
Implemented UI for redeeming fastbitcoins.com vouchers

- Redeem button is put to the bottom of Add Funds page,
- which redirects to a leaner redeem voucher form similar to [the one on their website](https://fastbitcoins.com/redeem),
- once the transaction is complete it shows up on payment list with it's unique identifier.


lib/bloc/account/account_model.dart
```diff
@@ -230,6 +230,9 @@ class PaymentInfo {
     if (_paymentResponse.invoiceMemo.description.startsWith("Bitrefill")) {
       return "Bitrefill";
     }
+    if (_paymentResponse.invoiceMemo.description.startsWith("Fastbitcoins")) {
```
roeierez: No need for this, we are using our invoice here and we put the description as we want it on the creation fo the invoice.

lib/widgets/breez_avatar.dart
```diff
@@ -126,12 +126,20 @@ class _VendorAvatar extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return Container(
+    return avatarURL.contains("fastbitcoins") ? CircleAvatar(
```
roeierez: I think it is better to extract out to two components one for each vendor, more explicit and clear this way.

lib/bloc/account/account_model.dart
erdemyerebasmaz: This is for title of payment list.

roeierez: But we don't need it because the description is ok. "Fastbitcoins.com Voucher"

lib/theme_data.dart
```diff
@@ -109,4 +112,17 @@ class BreezColors {
     2: const Color.fromRGBO(0, 137, 252, 1.0),
     3: const Color.fromRGBO(0, 120, 253, 1.0),
   };
+}
+
+class VendorColors {
```
roeierez: I like the idea of expressing the vendor style in the theme but it is better no to use static variables.
One option is to define "VendorTheme" class with its properties (iconBgColor, iconFgColor ...) and instantiate one for each vendor. We can also add a map of vendor name to VendorTheme on the main theme object.

lib/bloc/account/account_model.dart
erdemyerebasmaz: I see... Well, then I'm changing maxLines of PaymentItem widget's title to 1. It was splitting into multiple lines on my device.

PR #78, opened 2019-03-11
Do not show beta warning dialog if restore option is used

erdemyerebasmaz:


lib/routes/shared/initial_walkthrough.dart
```diff
@@ -164,7 +159,13 @@ class InitialWalkthroughPageState extends State<InitialWalkthroughPage>
                       elevation: 0.0,
                       shape: const StadiumBorder(),
                       onPressed: () {
-                        _proceedToRegister();
+                        widget._registrationBloc.registerSink.add(null);
```
roeierez: I am not sure how popping this dialog works as it hards to follow. It seems that when the user accepts the warning there is a "pushReplacement" for Home route. Doesn't it create two Home routes on on top of another? Since I think we already have one on the stack...

erdemyerebasmaz: `initialRoute: user.registered ? null : '/splash',`
The name of the first route to show, if a Navigator is built.
`home: new Home(accountBloc, invoiceBloc,`
This is the route that is displayed first when the application is started normally, unless initialRoute is specified. It's also the route that's displayed if the initialRoute can't be displayed

We only show splash screen on initialRoute, if the user is not registered the Home route won't be initialized.

roeierez: @erdemyerebasmaz I think it is not accurate. I am pretty sure the Home screen is under the intro page. Otherwise how the old logic worked when after restore we only popped to reach the home page?

PR #85, opened 2019-04-09
support bip21 with bolt11 invoice

roeierez:
This PR adds support for bip21 with bolt11 invoice: https://github.com/lncm/ideas/wiki/BIP21-w-BOLT11-wallet-support.
The format contains an embedded invoice so we only need to recognize it is bip21 format and extract the lightning invoice.

### DIFF
--- a/lib/bloc/invoice/invoice_bloc.dart
+++ b/lib/bloc/invoice/invoice_bloc.dart
@@ -110,6 +110,13 @@ class InvoiceBloc {
       if (lower.startsWith("lightning:")) {
         return s.substring(10);
       }
+
+      // check bip21 with bolt11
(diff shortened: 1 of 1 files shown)


meeDamian:
Hey @roeierez a couple of nits 😁

* `bitcoin:` prefix is case insensitive, can occur in both variants, and making it uppercase is beneficial to QR code resolution (alphanumeric mode).
* `amount` and `lightning` can occur in any order, and `amount` is optional (AFAIR they both should always be lowercase).

roeierez:
@meeDamian thanks a lot for your feedback!


> * `bitcoin:` prefix is case insensitive

I converted all the string to lower case before matching the regex so it should cover all cases.



> * `amount` and `lightning` can occur in any order

good catch, wasn't aware of that.

PR #86, opened 2019-04-10
Better approach for parsing bip21 URIs.

roeierez:
This PR takes a better (more robust) approach for parsing bip21 URI's and extract the optional bolt11 invoice embedded there.
The bip (https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki) states that these URIs follow the general format for URIs as set forth in RFC 3986 so based on that the code was changes to use Uri.parse builtin in dart to get the query params and specifically the bolt11 we are interested in.
In addition, some unit tests were added based on feedback for the previous implementation.
Thanks @meeDamian.

### DIFF
--- a/lib/utils/bip21.dart
+++ b/lib/utils/bip21.dart
@@ -1,0 +1,13 @@
+String extractBolt11FromBip21(String bip21) {
+  String lowerBit21 = bip21.toLowerCase();
+  if (lowerBit21.startsWith("bitcoin:")) {
+    try {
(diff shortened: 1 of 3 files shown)


lib/utils/bip21.dart
```diff
@@ -0,0 +1,13 @@
+String extractBolt11FromBip21(String bip21) {
+  String lowerBit21 = bip21.toLowerCase();
```
meeDamian: `lowerBip21` :)

PR #94, opened 2019-05-04
Processing payment animation

erdemyerebasmaz:
Animations are added to payment flow. Issue: [BU-151](https://breeztech.atlassian.net/browse/BU-151)
 - Added states to payment request dialog.
 - Implemented collapse animation.
 - Implemented flip animation.
 - Payment request dialog will collapse into payment item on successful payments.
 - If payment item is created within last 10 seconds, it will animate flip animation.
 - LoadingAnimatedText now uses textStyle parameters style if it's available.

## TODO:
 - ~Scroll to top after payment is complete.~
 - ~Collapse the dialog to the exact position of first payment item.~
 - ~Prevent scroll till collapse animation is complete.~ (is handled by Flutter)
 - ~Collapse drawer if it was open before payment process~

### DIFF
--- a/lib/routes/user/home/flip_transition.dart
+++ b/lib/routes/user/home/flip_transition.dart
@@ -1,0 +1,73 @@
+import 'package:flutter/material.dart';
+import 'dart:math';
+import 'dart:async';
+
+class FlipTransition extends StatefulWidget {
+  Widget firstChild;
+  Widget secondChild;
+  Widget thirdChild;
+
+  FlipTransition(this.firstChild, this.secondChild, this.thirdChild);
+
+  @override
+  State<StatefulWidget> createState() {
+    return FlipTransitionState();
+  }
+}
+
+class FlipTransitionState extends State<FlipTransition>
+    with TickerProviderStateMixin {
+  AnimationController _flipAnimationController;
+  Animation _flipAnimation;
+  static const flipDuration = Duration(seconds: 2);
+
+  @override
+  void initState() {
+    super.initState();
+    _flipAnimationController =
+        new AnimationController(vsync: this, duration: flipDuration);
+    _flipAnimation = Tween(begin: 0.0, end: 1.0).animate(CurvedAnimation(
+        parent: _flipAnimationController,
+        curve: Interval(0.0, 1.0, curve: Curves.fastOutSlowIn)));
+
+    _flipAnimationController.addStatusListener((status) {
+      if (status == AnimationStatus.completed) {
+        _flipAnimationController.reverse();
+      }
+    });
+    new Timer(new Duration(milliseconds: 900), () {
+      setState(() {
+        widget.firstChild = widget.secondChild;
+      });
+    });
+    new Timer(new Duration(milliseconds: 3500), () {
+      setState(() {
+        widget.firstChild = widget.thirdChild;
+      });
+    });
+    _flipAnimationController.forward();
+  }
+
+  @override
+  void dispose() {
+    _flipAnimationController.dispose();
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return AnimatedBuilder(
+        animation: _flipAnimationController,
+        builder: (BuildContext context, Widget child) {
+          return Container(
+              width: 40.0,
+              height: 40.0,
+              child: Transform(
+                transform: Matrix4.identity()
+                  ..rotateY(pi * _flipAnimation.value),
+                alignment: Alignment.center,
+                child: widget.firstChild,
+              ));
+        });
+  }
+}
--- a/lib/routes/user/home/flip_transition.dart
+++ b/lib/routes/user/home/flip_transition.dart
@@ -1,13 +1,11 @@
 import 'package:flutter/material.dart';
 import 'dart:math';
-import 'dart:async';
 
 class FlipTransition extends StatefulWidget {
-  Widget firstChild;
-  Widget secondChild;
-  Widget thirdChild;
+  final Widget firstChild;
+  final Widget secondChild;
 
-  FlipTransition(this.firstChild, this.secondChild, this.thirdChild);
+  FlipTransition(this.firstChild, this.secondChild);
 
   @override
   State<StatefulWidget> createState() {
@@ -35,16 +33,6 @@ class FlipTransitionState extends State<FlipTransition>
         _flipAnimationController.reverse();
       }
     });
-    new Timer(new Duration(milliseconds: 900), () {
-      setState(() {
-        widget.firstChild = widget.secondChild;
-      });
-    });
-    new Timer(new Duration(milliseconds: 3500), () {
-      setState(() {
-        widget.firstChild = widget.thirdChild;
-      });
-    });
     _flipAnimationController.forward();
   }
 
@@ -66,7 +54,7 @@ class FlipTransitionState extends State<FlipTransition>
                 transform: Matrix4.identity()
                   ..rotateY(pi * _flipAnimation.value),
                 alignment: Alignment.center,
-                child: widget.firstChild,
+                child:_flipAnimationController.value >= 0.4 ? widget.secondChild : widget.firstChild,
               ));
         });
   }
--- a/lib/routes/user/home/payment_item.dart
+++ b/lib/routes/user/home/payment_item.dart
@@ -88,8 +88,7 @@ class PaymentItem extends StatelessWidget {
     if (_createdWithin(Duration(seconds: 10))) {
       return PaymentItemAvatar(_paymentInfo);
     } else {
-      return FlipTransition(PaymentItemAvatar(_paymentInfo), SuccessAvatar(),
-          PaymentItemAvatar(_paymentInfo));
+      return FlipTransition(PaymentItemAvatar(_paymentInfo), SuccessAvatar());
     }
   }
 
--- a/lib/routes/user/home/payment_item.dart
+++ b/lib/routes/user/home/payment_item.dart
@@ -10,14 +10,17 @@ import 'package:breez/theme_data.dart' as theme;
 class PaymentItem extends StatelessWidget {
   final PaymentInfo _paymentInfo;
   final bool _lastItem;
+  final bool _firstItem;
+  final GlobalKey tableKey;
 
-  PaymentItem(this._paymentInfo, this._lastItem);
+  PaymentItem(this._paymentInfo, this._lastItem, this._firstItem, this.tableKey);
 
   @override
   Widget build(BuildContext context) {
     return new Stack(alignment: Alignment.bottomCenter, children: <Widget>[
       ListTile(
         leading: _buildPaymentItemAvatar(),
+        key: _firstItem ? tableKey : null,
         title: Text(
           _paymentInfo.title,
           style: theme.transactionTitleStyle,
--- a/lib/routes/user/home/payment_item.dart
+++ b/lib/routes/user/home/payment_item.dart
@@ -1,11 +1,11 @@
 import 'package:breez/bloc/account/account_model.dart';
 import 'package:breez/routes/user/home/payment_item_avatar.dart';
+import 'package:breez/routes/user/home/success_avatar.dart';
 import 'package:breez/utils/date.dart';
 import 'package:flutter/material.dart';
 import 'package:breez/routes/user/home/flip_transition.dart';
 import 'package:breez/widgets/payment_details_dialog.dart';
 import 'package:breez/theme_data.dart' as theme;
-import 'dart:math';
 
 class PaymentItem extends StatelessWidget {
   final PaymentInfo _paymentInfo;
@@ -17,30 +17,7 @@ class PaymentItem extends StatelessWidget {
   Widget build(BuildContext context) {
     return new Stack(alignment: Alignment.bottomCenter, children: <Widget>[
       ListTile(
-        leading: DateTime.fromMillisecondsSinceEpoch(
-            _paymentInfo.creationTimestamp.toInt() * 1000).difference(
-            DateTime.fromMillisecondsSinceEpoch(DateTime
-                .now()
-                .millisecondsSinceEpoch)) < -Duration(seconds: 10)
-            ? PaymentItemAvatar(
-            _paymentInfo)
-            : FlipTransition(
-            PaymentItemAvatar(_paymentInfo),
-            Container(
-                decoration: BoxDecoration(
-                    color: Colors.white,
-                    borderRadius: new BorderRadius.all(
-                        new Radius.circular(20.0))),
-                width: 20.0 * 2,
-                height: 20.0 * 2,
-                alignment: FractionalOffset.center,
-                child: Transform(
-                    transform: Matrix4.identity()
-                      ..rotateY(pi),
-                    alignment: Alignment.center,
-                    child: Icon(Icons.check,
-                        color: theme.BreezColors.blue[500]))),
-            PaymentItemAvatar(_paymentInfo)),
+        leading: _buildPaymentItemAvatar(),
         title: Text(
           _paymentInfo.title,
           style: theme.transactionTitleStyle,
@@ -104,5 +81,23 @@ class PaymentItem extends StatelessWidget {
         indent: 72.0,
       ),
     ]);
+  }
+
+  Widget _buildPaymentItemAvatar() {
+    // Show Flip Transition if the payment item is created within last 10 seconds
+    if (_createdWithin(Duration(seconds: 10))) {
+      return PaymentItemAvatar(_paymentInfo);
+    } else {
+      return FlipTransition(PaymentItemAvatar(_paymentInfo), SuccessAvatar(),
+          PaymentItemAvatar(_paymentInfo));
+    }
+  }
+
+  bool _createdWithin(Duration duration) {
+    return DateTime.fromMillisecondsSinceEpoch(
+        _paymentInfo.creationTimestamp.toInt() * 1000)
+        .difference(DateTime.fromMillisecondsSinceEpoch(
+        DateTime.now().millisecondsSinceEpoch)) <
+        - duration;
   }
 }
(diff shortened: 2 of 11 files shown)


lib/routes/user/home/flip_transition.dart
```diff
@@ -0,0 +1,73 @@
...
+    _flipAnimation = Tween(begin: 0.0, end: 1.0).animate(CurvedAnimation(
+        parent: _flipAnimationController,
+        curve: Interval(0.0, 1.0, curve: Curves.fastOutSlowIn)));
+
+    _flipAnimationController.addStatusListener((status) {
+      if (status == AnimationStatus.completed) {
+        _flipAnimationController.reverse();
+      }
+    });
+    new Timer(new Duration(milliseconds: 900), () {
```
roeierez: I think the use of timers here is not a good idea. Since there is already a running animation, we can use its value in the build method to render the right child.

lib/routes/user/home/payment_item.dart
```diff
@@ -15,7 +17,30 @@ class PaymentItem extends StatelessWidget {
   Widget build(BuildContext context) {
     return new Stack(alignment: Alignment.bottomCenter, children: <Widget>[
       ListTile(
-        leading: PaymentItemAvatar(_paymentInfo),
+        leading: DateTime.fromMillisecondsSinceEpoch(
```
roeierez: Better extract out these 5 lines condition into its own function.

lib/routes/user/home/payment_item.dart
```diff
@@ -15,7 +17,30 @@ class PaymentItem extends StatelessWidget {
...
+            _paymentInfo.creationTimestamp.toInt() * 1000).difference(
+            DateTime.fromMillisecondsSinceEpoch(DateTime
+                .now()
+                .millisecondsSinceEpoch)) < -Duration(seconds: 10)
+            ? PaymentItemAvatar(
+            _paymentInfo)
+            : FlipTransition(
+            PaymentItemAvatar(_paymentInfo),
+            Container(
```
roeierez: Extract out this Container widget to its own widget class.

lib/routes/user/received_invoice_notification.dart
```diff
@@ -38,15 +48,25 @@ class InvoiceNotificationsHandler {
         _setLoading(false);
         _handlingRequest = true;
 
+        // Do not pop dialog if there's a payment being processed
+        Future<bool> _onWillPop() async {
```
roeierez: I think it is cleaner to encapsulate the _inPogress in the PaymentRequestDialog class. It will move all the details level code including "onWillPop", "_sentPaymentResultSubscription" into the dialog itself and leave the logic here cleaner and more maintainable.
The dialog can also return status and we can use the result of the showDialog to show the appropriate flushbar.

lib/routes/user/received_invoice_notification.dart
```diff
@@ -72,7 +92,9 @@ class InvoiceNotificationsHandler {
 
     _sentPaymentResultSubscription =
         _accountBloc.fulfilledPayments.listen((fulfilledPayment) {
-      showFlushbar(_context, message: "Payment was successfuly sent!");
+          // Trigger the collapse animation and show flushbar 200ms after the animation is completed
+          paymentRequest.key.currentState.controller.reverse();
```
roeierez: See above comment, if we will move the logic into the dialog itself, this static global key access won't be needed.

lib/widgets/loading_animated_text.dart
```diff
@@ -39,7 +39,7 @@ class LoadingAnimatedTextState extends State<LoadingAnimatedText> {
   Widget build(BuildContext context) {   
     return RichText(
       text: TextSpan(
-        style: DefaultTextStyle.of(context).style,
+        style: widget.textStyle == TextStyle() ? DefaultTextStyle.of(context).style : widget.textStyle,
```
roeierez: What is the purpose of this? Can you explain?

lib/widgets/payment_request_dialog.dart
```diff
@@ -10,58 +9,183 @@ import 'package:flutter_advanced_networkimage/flutter_advanced_networkimage.dart
...
+      ..addListener(() {
+        setState(() {});
+      });
+    _invoiceAmountController.addListener(() {
       setState(() {});
     });
   }
 
+  @override
+  void didChangeDependencies() {
```
roeierez: All this method content can be re-located into initState method.
The only reason to use didChangDependencies is when we need to fetch from the inheritedWidget which is not available in initState.

lib/widgets/loading_animated_text.dart
erdemyerebasmaz: Previously, LoadingAnimatedText never used the value of textStyle parameter. We're using DefaultTextStyle.of(context).style if the parameters value is 'TextStyle()' which is default value of the parameter. We could remove the initial value and use 'widget.textStyle ?? DefaultTextStyle.of(context).style'

lib/widgets/payment_request_dialog.dart
erdemyerebasmaz: We are using didChangeDependencies to fetch values of inheritedWidget to initialize transitionAnimation, but we can remove the status listener of animation controller to initState.

lib/routes/user/home/flip_transition.dart
erdemyerebasmaz: I also removed the thirdChild parameter.

lib/widgets/payment_request_dialog.dart
```diff
@@ -20,48 +27,237 @@ class PaymentRequestDialog extends StatefulWidget {
...
+          var lineCount = (widget.invoice.description.length / 30).ceil();
+          _yMargin = _yMargin - (lineCount*24);
+        }
+      }
+      var _paymentItemStartPosition = ((MediaQuery.of(context).size.height) * 37)/100;
+      var _paymentItemEndPosition = ((MediaQuery.of(context).size.height) * 50)/100;
+      transitionAnimation = new RelativeRectTween(
+          begin: new RelativeRect.fromLTRB(0.0, _paymentItemStartPosition, 0.0, _paymentItemEndPosition),
+          end: new RelativeRect.fromLTRB(32.0, _yMargin, 32.0, _yMargin))
+          .animate(controller);
```
erdemyerebasmaz: I also think we need a clever solution here. I'm not really comfortable with this.
We are controlling the dialog size with transitionAnimation here. 300.0 = base dialog height and we add 10.0 if there's payee image and 24.0 per line if that request has note.
- We should learn the size of the dialog if it was rendered with IntrinsicHeight and use that instead of 300.0 without the payeeImage/description calculations.

 Start and end position of payment item is an estimate. 
- We should scroll to top of payment list after successful payment and use the exact spot of first payment items start and end position.

lib/widgets/loading_animated_text.dart
roeierez: @erdemyerebasmaz in that case I think your suggestion is better and clearer.
Let's remove the initial value.
One caveat though is that I have noticed some usage of LoadingAnimatedText with explicit textStyle passed from outside and as we see now in the code this style was never applied... Now with the change we suggesting the style will be used (as it should) but we need to make sure it looks good and creates any regression.

lib/widgets/payment_request_dialog.dart
```diff
@@ -20,48 +27,237 @@ class PaymentRequestDialog extends StatefulWidget {
...
+  bool _inProgress = false;
+  bool _isInit = false;
+
   @override
-  void initState() {    
+  void initState() {
     super.initState();
-    _invoiceAmountController.addListener((){
+    widget.accountBloc.accountStream.listen((acc) {
+      if (acc.paymentRequestInProgress != null && acc.paymentRequestInProgress.isNotEmpty) {
```
roeierez: Instead of the 4 lines we can use this short expression:

```suggestion
      _inProgress = acc.paymentRequestInProgress != null && acc.paymentRequestInProgress.isNotEmpty;
```

lib/widgets/payment_request_dialog.dart
```diff
@@ -20,48 +27,237 @@ class PaymentRequestDialog extends StatefulWidget {
...
+        _inProgress = true;
+      } else if (acc.paymentRequestInProgress == null || acc.paymentRequestInProgress.isEmpty){
+        _inProgress = false;
+      }
+    });
+    _listenPaymentsResults();
+    setState(() {
```
roeierez: We shouldn't use `setState` within `initState` method. You can do the assignment without `setState`.

lib/widgets/payment_request_dialog.dart
roeierez: @erdemyerebasmaz  I share the same concerns with you...
we can't use relative height such as: `((MediaQuery.of(context).size.height) * 37)/100;` as it will land on different points in different resolutions.
The position in  Y axis of where the first payment item starts is the same on every device and we can pass it to the widget if needed.
I am suggesting the following simple approach to start with:

1.  Let's move all the animation initialization to the fulfilledPayments listener. This way we can safely query the height of the widget and know for sure it was fully rendered. 
2. for the animation we can find the dialog position and sizes using `context.findRenderObject`. For the end state of the animation I suggest to leave the Y Axis position as in the start state and only shrink the height. Let's see how this looks before we struggle to pass the exact position of the first item...
I think it will a lot simplify the code and take out all the calculations.

lib/widgets/payment_request_dialog.dart
```diff
@@ -20,48 +27,237 @@ class PaymentRequestDialog extends StatefulWidget {
...
+  @override
+  dispose() {
+    controller.dispose();
+    super.dispose();
+  }
+
+  _listenPaymentsResults() {
+    AccountSettings accountSettings;
+
+    widget.accountBloc.accountSettingsStream
```
roeierez: Missing cleanup for these subscriptions on dispose.

lib/widgets/payment_request_dialog.dart
```diff
@@ -36,6 +37,10 @@ class PaymentRequestDialogState extends State<PaymentRequestDialog>
     with SingleTickerProviderStateMixin {
 
   final _formKey = GlobalKey<FormState>();
+  GlobalKey _titleKey = GlobalKey();
```
roeierez: Since you are using fixed heights for title and actions, why do we need the `_titleKey` and `_actionsKey` ?

lib/routes/user/home/payment_item.dart
```diff
@@ -10,14 +10,17 @@ import 'package:breez/theme_data.dart' as theme;
 class PaymentItem extends StatelessWidget {
   final PaymentInfo _paymentInfo;
   final bool _lastItem;
+  final bool _firstItem;
+  final GlobalKey tableKey;
```
roeierez: According to the usage of this key shouldn't it be "firstPaymentItemKey" ?

lib/widgets/payment_request_dialog.dart
```diff
@@ -127,7 +96,45 @@ class PaymentRequestDialogState extends State<PaymentRequestDialog>
...
+          CurvedAnimation(parent: controller, curve: Curves.slowMiddle));
+      colorAnimation = new ColorTween(
+        begin: theme.BreezColors.blue[500],
+        end: theme.BreezColors.white[500],
+      ).animate(controller)
+        ..addListener(() {
+          setState(() {});
+        });
+
+      RenderBox _paymentTableBox = widget.tableKey.currentContext
```
roeierez: How the animation will work on the first item when the key is not used and not related to any widget?

lib/widgets/payment_request_dialog.dart
```diff
@@ -127,7 +96,45 @@ class PaymentRequestDialogState extends State<PaymentRequestDialog>
...
+          .findRenderObject();
+      var _paymentItemStartPosition = _paymentTableBox
+          .localToGlobal(Offset.zero)
+          .dy - 32.0;
```
roeierez: What does the 32.0 stand for?

lib/widgets/payment_request_dialog.dart
```diff
@@ -127,7 +96,45 @@ class PaymentRequestDialogState extends State<PaymentRequestDialog>
...
+      var _paymentItemEndPosition = (MediaQuery
+          .of(context)
+          .size
+          .height - _paymentItemStartPosition) - 72.0 - 32.0;
```
roeierez: I am guessing the 7.2.0 is the item height, there is already a constant defined for this number, we should use it instead.

lib/widgets/payment_request_dialog.dart
```diff
@@ -55,7 +60,13 @@ class PaymentRequestDialogState extends State<PaymentRequestDialog>
   String _amountToPayStr;
 
   bool _inProgress = false;
-  bool _isInit = false;
+
+  static const double _titleHeight = 48.0;
+  static const double _actionsHeight = 64.0;
+  double _dialogContentHeight;
```
roeierez: why do we need to define `_dialogContentHeight` as class variable? we only need to calculate it when needed on paymentfulfilled.

lib/widgets/payment_request_dialog.dart
```diff
@@ -466,6 +542,7 @@ class PaymentRequestDialogState extends State<PaymentRequestDialog>
     if (toPay > 0 && account.maxAllowedToPay >= toPay) {
       actions.add(SimpleDialogOption(
         onPressed: (() async {
+          _afterLayout(null);
```
roeierez: I think it is better to calculate the dialog height on payment fulfilled rather than here.

erdemyerebasmaz: We are using it to set the height of the consequent dialogs

lib/widgets/payment_request_dialog.dart
erdemyerebasmaz: https://github.com/breez/breezmobile/pull/94#discussion_r283110876

lib/widgets/payment_request_dialog.dart
erdemyerebasmaz: I will test this with another widget to be sure but the start position was at the center of the widget.
One of the methods may calculate distance from safearea's 0.0 while the other uses top of screen. The difference may be the system status bar. If so, I'll use the constant for it.

lib/widgets/payment_request_dialog.dart
erdemyerebasmaz: It won't work at the moment. As I mentioned earlier, findRenderObject() does not return a RenderBox when used on table so we can calculate it's size/position.

lib/widgets/payment_request_dialog.dart
erdemyerebasmaz: There are cases where there are no title or action widgets rendered.

I will calculate the height if they exist and they will take precedence over the constant values.

lib/widgets/payment_request_dialog.dart
roeierez: We need a solution for this case as well, it is the first payment the user does...
One solution is to add a "dummy" sizebox with size 0 only to get its location instead of the table.

lib/widgets/payment_request_dialog.dart
roeierez: > We are using it to set the height of the consequent dialogs

I understand that. I was trying to say that since the heights are only used after payment fulfilled, we can apply all these calculation at that point and make all the variables local to that context instead of calling here _afterLayout and maintain a class variable.

erdemyerebasmaz: I'm a bit lost. Heights are not only used after payment fulfilled.
First dialog takes as much space as it needs and it's always bigger than the following dialogs(confirmation, processing payment). When user presses approve, we call _afterLayout() to calculate the size of the initial dialog to set the height of following dialogs. So they are same in size.

lib/widgets/payment_request_dialog.dart
```diff
@@ -107,26 +107,31 @@ class PaymentRequestDialogState extends State<PaymentRequestDialog>
...
       });
-      // Scroll the payment list to top
-      widget._scrollController.hasClients ?
-      widget._scrollController.animateTo(widget._scrollController.position.minScrollExtent, duration: Duration(milliseconds: 300), curve: Curves.ease).whenComplete( () {
-        new Timer(new Duration(milliseconds: 100), () {
-          _initializeTransitionAnimation();
-          // Trigger the collapse animation and show flushbar after the animation is completed
-          controller.reverse().whenComplete(() =>
-              showFlushbar(context, message: "Payment was successfuly sent!"));
+      if (widget._scrollController.hasClients) {
```
roeierez: This block, which has a repetitive code inside:
```
     if (widget._scrollController.hasClients) {
        widget._scrollController.animateTo(
            widget._scrollController.position.minScrollExtent,
            duration: Duration(milliseconds: 200), curve: Curves.ease)
            .whenComplete(() {
          new Timer(new Duration(milliseconds: 50), () {
            _initializeTransitionAnimation();
            controller.value = 1.0;
            // Trigger the collapse animation and show flushbar after the animation is completed
            controller.reverse().whenComplete(() =>
                showFlushbar(context, message: "Payment was successfuly sent!"));
          });
        });
      } else {
        _initializeTransitionAnimation();
        controller.value = 1.0;
        // Trigger the collapse animation and show flushbar after the animation is completed
        controller.reverse().whenComplete(() =>
            showFlushbar(context, message: "Payment was successfuly sent!"));
      }
```

I suggest to change it to something like this (didn't compile it):

```
     Future scrollAnimationFuture = Future.value(null);
     if (widget._scrollController.hasClients) {
        scrollAnimationFuture = widget._scrollController.animateTo(
            widget._scrollController.position.minScrollExtent,
            duration: Duration(milliseconds: 200), curve: Curves.ease)           
        }).whenComplete(() => Future.delayed(Duration(milliseconds: 50));
      }
 
      scrollAnimationFuture .whenComplete(() {
          _initializeTransitionAnimation();
            controller.value = 1.0;
            // Trigger the collapse animation and show flushbar after the animation is completed
            controller.reverse().whenComplete(() =>
                showFlushbar(context, message: "Payment was successfuly sent!"));
```

lib/widgets/payment_request_dialog.dart
erdemyerebasmaz: It was the height of status bar. There was a 4.0 dp difference for my screen.

PR #97, opened 2019-05-17
Set minimum height to payment request dialogs initial state

erdemyerebasmaz:


roeierez:
I think it is better to keep it simple and put a constant height that covers the largest dialog case.
and omit the calculations.
Right now all the different calculations making things hard to follow.

roeierez:
@erdemyerebasmaz overall I suggest slightly small changes mainly for readability and a bit simplification, LMK what you think of these changes:
1. consolidates all dialog part heights into one (`_dialogContentHeigh`,  `_dialogTitleHeight`, `_dialogActionsHeight`), Let's call it `_initialDialogHeight` and calculate this value on the current _getDialogSize method.
2. use this value as minimum value for the two dialogs to follow, something like (height: max(_initialDialogSize, 200)

erdemyerebasmaz:
> @erdemyerebasmaz overall I suggest slightly small changes mainly for readability and a bit simplification, LMK what you think of these changes:
> 
> 1. consolidates all dialog part heights into one (`_dialogContentHeigh`,  `_dialogTitleHeight`, `_dialogActionsHeight`), Let's call it `_initialDialogHeight` and calculate this value on the current _getDialogSize method.
> 2. use this value as minimum value for the two dialogs to follow, something like (height: max(_initialDialogSize, 200)

Waiting for confirmation state has action row, processing payment doesn't. That's why the calculations are different for both states. Using single variable won't work for states with unequal fields.

Although, this has it's problems of it's own, to use a single variable, I will refactor AlertDialog and use Dialog widget that has a Container with minHeight wrapped to it's content. This minHeight will be max(_initialDialogSize, 200).

roeierez:
> Waiting for confirmation state has action row, processing payment doesn't. That's why the calculations are different for both states. Using single variable won't work for states with unequal fields.

Understood but at the same time that's what makes the code very hard to reason about.
So this complexity comes from the fact that we are trying to render three different dialogs at the same height combined with the way AlertDialog is implemented (three different parts: title, body, actions).
I agree that we can't simplify the code without major changes to this page. I think the requirement to have  three steps that are related to each other indicates a need for a something else like a  "wizard" maybe...

Let's take this bullet right now but if we have any more bugs here regarding height calculations we will refactor.

roeierez:
The last three commits are in the right direction!

lib/widgets/payment_request_dialog.dart
```diff
@@ -195,231 +77,96 @@ class PaymentRequestDialogState extends State<PaymentRequestDialog>
...
-          children: children,
-        ),
-      );
+    if (_state == PaymentRequestState.PROCESSING_PAYMENT) {
+      return ProcessingPaymentDialog(widget.context, widget.accountBloc, widget.firstPaymentItemKey, widget.scrollController, _initialDialogSize);
+    } else if (_state == PaymentRequestState.WAITING_FOR_CONFIRMATION) {
+      return PaymentConfirmationDialog(widget.accountBloc, widget.invoice, _initialDialogSize, _amountToPay, _amountToPayStr, (state) => _onStateChange(state));
     } else {
-      return null;
+      return _buildPaymentRequestDialog();
```
roeierez: The two other dialogs were well extracted out to their own widgets which contributes a lot to the code quality of this PR.
Any reason why the third dialog is not in its own widget?

roeierez:
@erdemyerebasmaz the changes looks good! Just one comment regarding the third dialog which by being addressed should finalize this PR.

lib/widgets/payment_request_dialog.dart
erdemyerebasmaz: I believe I was confused because of the class naming, PaymentRequestDialog. I extracted this dialog into a widget called PaymentRequestInfoDialog.

PR #99, opened 2019-05-17
BU-158 | Change all dialogs to round corners

erdemyerebasmaz:
This PR contains changes to all dialog borders and layout of refund form dialog.

### DIFF
--- a/lib/routes/pos/home/pos_invoice.dart
+++ b/lib/routes/pos/home/pos_invoice.dart
@@ -323,6 +323,8 @@ class POSInvoiceState extends State<POSInvoice> {
                   },
                 ),
               ],
+              shape: RoundedRectangleBorder(
+                  borderRadius: BorderRadius.all(Radius.circular(12.0))),
             );
           });
     } else {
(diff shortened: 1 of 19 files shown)


roeierez:
@erdemyerebasmaz it looks like we need one widget for dialog with all our custom style embedded instead of creating the same configuration for each dialog in a repetitive way.
From the code it looks like we can create our own `AlertDialog` widget that takes: title, content, actions as arguments and implements once the needed style.
What do you think?

erdemyerebasmaz:
> @erdemyerebasmaz it looks like we need one widget for dialog with all our custom style embedded instead of creating the same configuration for each dialog in a repetitive way.
> From the code it looks like we can create our own `AlertDialog` widget that takes: title, content, actions as arguments and implements once the needed style.
> What do you think?

I don't see a reason why we should create a new widget, to make subsequent changes to style easier? Will there be further customization to this 'BreezAlertDialog' such as transition animation.
Also, it won't be applicable to all dialogs. (e.g. BreezAvatarDialog, BreezDatePickerDialog, PaymentDetailsDialog)

roeierez:
Hi @erdemyerebasmaz appreciate your input.
To me duplicating the same behavior, including style changes that are not supported at the theme level (like round corners), are good motivation to consolidate. 
I also thing that the fact we are using AlertDialog is only an implementation detail that is convenient for us (and I see no reason not to keep using it).
You had to change 19 files only to get round corners across the board and on the next style customization, animation, etc... we will have to do it again instead of change in one central point.
To be clear you didn't introduce this problem, it already existed, I just think it is a good opportunity to take it one small step further. 
A good way to proceed here is push this PR and try my approach in a different PR. I am OK with that.

PR #116, opened 2019-06-28
Implemented fiat currency

erdemyerebasmaz:
#### [BU-163](https://breeztech.atlassian.net/browse/BU-163)
This PR implements fiat currency and currency converter dialog.
Long pressing balance field shows the balance equivalent in users fiat currency preference.
 - This preference can only be changed from currency converter dialog. 
 - The dialog is accessible by an icon button attached to every AmountFormField.

The currency converter dialog converts the selected fiat currency($,€,£,¥) to users currency(BTC/Bit/Sat) equivalent.
 - The exchange rate is updated periodically.
 - The update in exhange rate is informed to the user by a blink in the exhange rate label.
 - The converted amount is pasted to the amount form field that opened this dialog.


lib/routes/user/connect_to_pay/payment_details_form.dart
```diff
@@ -74,8 +75,13 @@ class _PaymentDetailsFormState extends State<PaymentDetailsForm> {
                         currency: widget._account.currency,
                         controller: _amountController,
                         decoration: new InputDecoration(
-                            labelText: widget._account.currency.displayName +
-                                " Amount"),
+                          labelText: widget._account.currency.displayName + " Amount",
```
roeierez: Instead of adding this decoration for every instance of AmountField, it is better to implement it once inside the AmountFormField.

lib/routes/user/home/wallet_dashboard.dart
```diff
@@ -18,58 +19,116 @@ class WalletDashboard extends StatelessWidget {
...
+  double _convertedBalance = 0.0;
+
+  @override
+  void initState() {
+    super.initState();
+    _breezLib = new ServiceInjector().breezBridge;
+  }
+
+  void _getExchangeRates() async {
+    Rates _rate = await _breezLib.rate();
```
roeierez: It is better that the UI won't access directly the services layer (e.g breezLib), the exchange rates should come from the bloc.

roeierez: I also think we should add fiatBalance on the AccountModel as a calculated getter based on the btc balance and the conversion rate.

lib/routes/user/home/wallet_dashboard.dart
```diff
@@ -18,58 +19,116 @@ class WalletDashboard extends StatelessWidget {
...
+                                style: theme.headline
+                                    .copyWith(fontSize: startHeaderSize - (startHeaderSize - endHeaderFontSize) * widget._offsetFactor)))
+                    : SizedBox(),
               ),
             )
-      ],
-    );
+          ],
+        ),
+        behavior: HitTestBehavior.translucent,
```
roeierez: Can't we use a gesture here instead of the pointerup/pointerdown logic? flutter support "long press" and "tap" gestures.

lib/widgets/currency_converter_dialog.dart
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  _updateExchangeRates() {
+    _exchangeRateTimer = Timer.periodic(Duration(seconds: 3), (_) async {
+      _getExchangeRate();
+    });
+  }
+
+  _getExchangeRate() async {
+    Rates _rate = await _breezLib.rate();
```
roeierez: let's use a bloc for data fetching.

roeierez:
As a general recommendation I think we should add the following fields on the AccountModel: 
`fiatCurrency` - which will include the symbols and exchange rate.
`fiatBalance` - a calculated getter that calculates the fiat balance based on the btc balance and the exchange rate.
This will allow to remove the direct calls to breezLib and use the AccountBloc instead.

lib/bloc/user_profile/fiat_currency.dart
```diff
@@ -0,0 +1,46 @@
...
+      case "¥":
+        return "JPY";
+      case "\$":
+        return "USD";
+      default:
+        return "USD";
+    }
+  }
+
+  String get logoPath {
```
jemilsson: Is this really a sustainable solution if we wish to support many fiat currencies? Do we really need a separate icon for each currency? Most fiat currencies have a unicode symbol, perhaps we could rely on those and avoid creating new icons for each currency? I think it would be great if the fiat conversion could be general enough so that more currencies could be added without much work.

kingonly:
We need to rely on the server response regarding the supported currencies. Although initially we want to support only a few, it makes send to be more generic in the client. Use something like: https://gist.github.com/Fluidbyte/2973986 if Dart doesn't have this information and create a generic icon for additional currencies (keep the specific icons, but if no specific icon, use a generic one).

yaslama:
https://github.com/wiredmax/world-currencies/blob/master/dist/json/currencies.json or https://github.com/xsolla/currency-format/blob/master/currency-format.json can also be used.

lib/bloc/user_profile/fiat_currency.dart
roeierez: @jemilsson good point!
@erdemyerebasmaz my suggestion is to use a data file (see this for example: https://developers.google.com/public-data/docs/canonical/currencies_csv) to represents all currencies.. My first approach would be to look for this data in dart/flutter but I couldn't find a built-in feature that supports both currency short name and symbol. Anyway we can change the data representation later if built in support is added. The important thing is that these currencies are dynamically loaded and future support for additional currencies won't require code changes.
As for the icon let's talk to Roy about his view before any changes.

lib/bloc/account/account_bloc.dart
```diff
@@ -105,6 +112,9 @@ class AccountBloc {
   Sink get optimizationWhitelistRequestSink =>
       _permissionsHandler.optimizationWhitelistRequestSink;
 
+  final _exchangeRateController = new BehaviorSubject<Rates>();
```
roeierez: We don't need this controller as we have List<FiatConversion> on the account model. Any consumer of exchange rate should be able to get it by listening to the account stream.

lib/bloc/user_profile/currency_data.dart
```diff
@@ -0,0 +1,25 @@
+import 'dart:convert';
+
+Map<String, CurrencyData> currencyDataFromJson(String str) =>
+    new Map.from(json.decode(str)).map((k, v) => new MapEntry<String, CurrencyData>(k, CurrencyData.fromJson(k, v)));
+
+class CurrencyData {
```
roeierez: CurrencyData class should be part of the Currency service and not under the user_profile directory.

lib/bloc/user_profile/fiat_conversion.dart
```diff
@@ -0,0 +1,36 @@
+import 'package:breez/bloc/user_profile/currency_data.dart';
+
+class FiatConversion {
```
roeierez: FiatConversion class should be under account bloc directory.

lib/bloc/user_profile/user_profile_bloc.dart
```diff
@@ -32,6 +36,9 @@ class UserProfileBloc {
   final _currencyController = new BehaviorSubject<Currency>();
```
roeierez: We have here two controllers:
1. _currencyController - not used should be deleted.
2. _fiatConversionController - which is used to detect currency changes, I think better to call it _currencyController as only accountBloc is dealing with conversion.

lib/widgets/amount_form_field.dart
```diff
@@ -24,33 +30,48 @@ class AmountFormField extends TextFormField {
...
+      Int64 intAmount = accountModel.currency.parse(value);
       if (intAmount <= 0) {
         return "Invalid amount";
       }
       String msg;
-      if (validatorFn != null) {
-        msg = validatorFn(intAmount);
-      }   
-      return msg;   
+      if (accountModel.validateOutgoingOnChainPayment != null) {
```
roeierez: As part of simplifying the use of AmountFormField the validator now always use `accountModel.validateOutgoingOnChainPayment` where on some cases it should use `accountModel. validateOutgoingPayment` or `accountModel.validateIncomingPayment`. I think for now it is better to pass this function as parameter (as was before the change).

lib/widgets/currency_converter_dialog.dart
```diff
@@ -0,0 +1,230 @@
...
+    if (!_isInit) {
+      _accountBloc = AppBlocsProvider.of<AccountBloc>(context);
+      _userProfileBloc = AppBlocsProvider.of<UserProfileBloc>(context);
+      registerListeners();
+      _isInit = true;
+    }
+    super.didChangeDependencies();
+  }
+
+  void registerListeners() {
```
roeierez: I think instead of all these listeners it will be more simplified to use one StreamBuilder for AccountModel inside the build function.

lib/bloc/account/account_bloc.dart
```diff
@@ -601,6 +615,42 @@ class AccountBloc {
...
+  _updateExchangeRates() {
+    _getExchangeRate();
+    SystemChannels.lifecycle.setMessageHandler((msg) {
+      switch (msg) {
+        case "AppLifecycleState.resumed":
+          _getExchangeRate();
+          _startExchangeRateTimer();
+          break;
+        default:
+        // cancel timer when AppLifecycleState is paused, inactive or suspending
```
roeierez: cancel timer?

lib/widgets/currency_converter_dialog.dart
```diff
@@ -0,0 +1,255 @@
...
+    return StreamBuilder<AccountModel>(
+        stream: _accountBloc.accountStream,
+        builder: (context, snapshot) {
+          AccountModel account = snapshot.data;
+          if (!snapshot.hasData) {
+            return Container();
+          }
+
+          _currency = account.currency;
+          _fiatCurrency = account.fiatCurrency;
```
roeierez: Don't need this variable _fiatCurrency, everything can be taken from accountModel.
Even if it is needed on state, it is not a good practice to initialize it within the build method.

lib/widgets/currency_converter_dialog.dart
```diff
@@ -0,0 +1,255 @@
...
+    return _exchangeRate == null
+        ? Text("", style: theme.smallTextStyle)
+        : Text("1 BTC = $_exchangeRate ${_fiatCurrency.currencyData.shortName}",
+            style: theme.smallTextStyle.copyWith(color: _colorAnimation.value));
+  }
+
+  _selectFiatCurrency(shortName) {
+    setState(() {
+      int _oldFractionSize = _fiatCurrency.currencyData.fractionSize;
+      _fiatCurrency = _fiatConversionList.firstWhere((f) => f.currencyData.shortName == shortName);
```
roeierez: Also here no need to count on this state variable as it is not needed. Instead I suggest to pass the `AccountModel` to this method that is called from the `build` method.

PR #124, opened 2019-07-18
Scroll to bottom of visible content after opening an expandable item

erdemyerebasmaz:
This PR solves [BU-168](https://breeztech.atlassian.net/browse/BU-168)
If opening an expandable menu item stretches beyond visible content, the navigation drawer will scroll to bottom.

### DIFF
--- a/lib/widgets/navigation_drawer.dart
+++ b/lib/widgets/navigation_drawer.dart
@@ -1,3 +1,4 @@
+import 'dart:async';
 import 'package:breez/bloc/blocs_provider.dart';
 import 'package:flutter/material.dart';
 import 'package:breez/theme_data.dart' as theme;
@@ -32,6 +33,8 @@ class NavigationDrawer extends StatelessWidget {
   NavigationDrawer(this._avatar, this._drawerGroupedItems,
       this._onItemSelected);
 
+  final ScrollController _scrollController = ScrollController();
+
   @override
   Widget build(BuildContext context) {
     UserProfileBloc userProfileBloc = AppBlocsProvider.of<UserProfileBloc>(context);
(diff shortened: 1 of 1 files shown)


lib/widgets/navigation_drawer.dart
```diff
@@ -193,6 +199,12 @@ class _ExpansionTile extends StatelessWidget {
             ),
           ),
           children: items.map((item) => Padding(padding: EdgeInsets.only(left: 28.0), child: item)).toList(),
+          onExpansionChanged: (isExpanded) {
+            if (isExpanded) Timer(Duration(milliseconds: 200), () =>
```
roeierez: 1. Missing parenthesis for the if block.
2. Why do we need the timer here? The scroll is configured to animate in  400 milliseconds already.
3. Nice to have - Currently this logic won't work if we will put an expanded item in the top of the list (because there is a scroll to the end of the list). Better try to scroll for the current item to be visible.

erdemyerebasmaz: 1. It is not required in this case, but I'll add it for readability.
2. The elements are not expanded yet when calling animateTo, that's why we wait for them to be visible.
3. True, I agree this is a specific and a quick solution to this story. We should implement a generic solution.

roeierez: Regarding #2 does it create a problem if we start the scroll animation before waiting for the expansion to complete?

erdemyerebasmaz: Yes, maxScrollExtent hasn't changed according to initial state, so it scrolls to the same position(does not move).

roeierez: > Yes, maxScrollExtent hasn't changed according to initial state, so it scrolls to the same position(does not move).

Got you.

PR #125, opened 2019-07-18
Invoice actions

erdemyerebasmaz:
This PR implements invoice bottom sheet menu as explained on [BU-167](https://breeztech.atlassian.net/browse/BU-167?utm_medium=referral-external&utm_source=jira-slack&utm_term=)


lib/routes/user/home/invoice_bottom_sheet.dart
```diff
@@ -0,0 +1,88 @@
...
+              _buildInvoiceMenuItem(widget.isSmallView, "PAY", "src/icon/qr_scan.png", () async {
+                String decodedQr = await BarcodeScanner.scan();
+                widget.invoiceBloc.decodeInvoiceSink.add(decodedQr);
+              }),
+              _buildInvoiceMenuItem(
+                  widget.isSmallView, "CREATE", "src/icon/paste.png", () => Navigator.of(context).pushNamed('/create_invoice')),
+            ]));
+  }
+
+  Widget _buildInvoiceMenuItem(bool isSmallView, String title, String iconPath, Function function, {bool isFirst = false}) {
```
roeierez: `isSmallView` is a class variable which this function has access to. No need to pass it as argument.

lib/routes/user/home/account_page.dart
```diff
@@ -164,6 +165,13 @@ class AccountPageState extends State<AccountPage> with SingleTickerProviderState
               return account != null && !account.initial ? FloatingActionsBar(account, height, heightFactor) : Positioned(top: 0.0, child: SizedBox());
             },
           ),
+          ScrollWatcher(
+            controller: widget.scrollController,
+            builder: (context, offset) {
+              double height = (DASHBOARD_MAX_HEIGHT - offset).clamp(DASHBOARD_MIN_HEIGHT, DASHBOARD_MAX_HEIGHT);
+              return account != null && !account.initial ? InvoiceBottomSheet(_invoiceBloc, height < 160.0) : Positioned(top: 0.0, child: SizedBox());
```
roeierez: We need to show the invoice sheet only if there is an opened channel, the same as we do with the corresponding menu items. (` if account.active` )

lib/routes/user/home/invoice_bottom_sheet.dart
```diff
@@ -0,0 +1,88 @@
...
+    return AnimatedContainer(
+      width: isSmallView ? 56.0 : 126.0,
```
roeierez: How can we be sure that this width wraps the text well for all resolutions?

erdemyerebasmaz: I followed the specifications on Zeplin. The values are in dp(density independent pixels) so there's no reason to believe they'd look different on other resolutions.

kingonly: I've tested Zoomed display and it looks good. We have an issue with large text size in iOS and Android, but it exists also in Connect to Pay button.
I suggest to commit this and I'll open a bug on the problematic buttons with large text size.

PR #128, opened 2019-07-30
Add optional PIN code

erdemyerebasmaz:
Story: [BU-169](https://breeztech.atlassian.net/browse/BU-169)

A new security layer is introduced with this PR. Users will now be able to set a 6-digit PIN code which will be prompted whenever they enter the app.

Security PIN menu item is added under Advanced in Drawer. On Security PIN page users can set PIN related settings such as create/disable, change PIN and enable secure backup with PIN.


lib/bloc/account/account_actions.dart
```diff
@@ -46,4 +46,11 @@ class ChangeSyncUIState extends AsyncAction {
   ChangeSyncUIState(this.nextState);
 }
 
-class FetchRates extends AsyncAction {}
\ No newline at end of file
+class FetchRates extends AsyncAction {}
+
+class UpdateSecurityModel extends AsyncAction {
```
roeierez: This action is related to UserProfileBloc, therefore should be moved under the user_profile folder.

lib/bloc/user_profile/user_profile_bloc.dart
```diff
@@ -39,6 +46,10 @@ class UserProfileBloc {
   final _fiatConversionController = new BehaviorSubject<String>();
   Sink<String> get fiatConversionSink => _fiatConversionController.sink;
 
+  final _securityModelController = new BehaviorSubject<SecurityModel>();
+  Sink<SecurityModel> get securityModelSink => _securityModelController.sink;
```
roeierez: We don't need this controller and sink. We should use the `_userController` and push the changes user there.

lib/bloc/user_profile/breez_user_model.dart
```diff
@@ -31,7 +34,9 @@ class BreezUserModel {
         name = json['name'],
         color = json['color'],
         animal = json['animal'],
-        _image = json['image'];
+        _image = json['image'],
+        securityModel = SecurityModel.fromJson(json['securityModel']),
+        waitingForPin = SecurityModel.fromJson(json['securityModel']).pinCode != null;
```
roeierez: why persisting the `watingForPin` ? isn't it used only on start of the app when reading the pin from the keychain?

lib/bloc/user_profile/breez_user_model.dart
```diff
@@ -9,12 +10,14 @@ class BreezUserModel {
   String color = '';
   String animal = '';
   String _image;
+  SecurityModel securityModel;
```
roeierez: securityModel and waitingForPin should be final. Also all other model variables should be final as models are immutable.

lib/bloc/user_profile/user_profile_bloc.dart
```diff
@@ -123,6 +149,25 @@ class UserProfileBloc {
...
+      var handler = _actionHandlers[action.runtimeType];
+      if (handler != null) {
+        handler(action).catchError((e) => action.resolveError(e));
+      }
+    });
+  }
+
+  Future _updateSecurityModel(UpdateSecurityModel updateSecurityModelAction) async {
+    if (updateSecurityModelAction.pinCode != null) {
+      await _secureStorage.write(key: 'pinCode', value: updateSecurityModelAction.pinCode);
```
roeierez: ideally we should only write to the secure storage if current pintCode is different than the new pinCode.

lib/routes/shared/security_pin/prompt_pin_code.dart
```diff
@@ -0,0 +1,298 @@
...
+
+  _setPinCodeForValidation() {
+    setState(() {
+      _tmpPassword = _enteredPinCode;
+    });
+    _setPinCodeInput("");
+    _setLabel("Re-enter your new PIN");
+  }
+
+  void _validatePinCode(bool isValid) {
```
roeierez: Nit: rename to `_onPinCodeValide(bool isValid)` since it is not really validating, just responding to the validation result that was done earlier.

lib/bloc/account/account_actions.dart
erdemyerebasmaz: 👍 
On top of this I suggest we separate AsyncAction from account_actions and import it in both account & user actions.

lib/bloc/user_profile/user_profile_bloc.dart
erdemyerebasmaz: 👍

lib/bloc/user_profile/breez_user_model.dart
erdemyerebasmaz: It's used to indicate that the pin was validated. We change it to false after validating pin to start registering notification handlers.

lib/routes/shared/security_pin/prompt_pin_code.dart
```diff
@@ -0,0 +1,298 @@
...
+  Container _numberButton(String number) {
+    return Container(
+      child: new FlatButton(
+        onPressed: () => _onNumButtonPressed(number),
+        child: new Text(number, textAlign: TextAlign.center, style: theme.numPadNumberStyle),
+      ),
+    );
+  }
+
+  _onNumButtonPressed(String numberText) {
```
roeierez: This method code is too complicated, it took me some time to understand that it actually handles 3 use cases:
1. Enter PIN to unlock.
2. Change existing PIN.
3. Set new PIN.

And that the code differentiate the cases by these variables: `setPassword`, `changePassword`
I think we can simplify:
1. use cases 2,3 are the same. The difference is that the old PIN is empty in case of new pin so no need to validate against it before changing.
2. We can use one flag `changePIN` which when true it refers to "changing existing pin" case and when false it refers to "Enter to unlock" case.

We will need to iterate again after this simplification. It needs to be very clear.

roeierez: One more thing that will simplify a lot is to separate the UI of the pinCode from the use cases that are handled in the stateful widget. 
I suggest to create a StatelessWidget and move all the stateless UI there (numpad, circles, errorMessage, etc...).
We can call it `PinCodeWidget` and can put it in the general widgets directory.
Then we can leave the validation and controller code here in the stateful component and see if it looks simple enough or we need to separate further.

lib/routes/shared/security_pin/security_pin_page.dart
```diff
@@ -0,0 +1,172 @@
...
+  @override
+  void didChangeDependencies() {
+    super.didChangeDependencies();
+    if (!_isInit) {
+      _userProfileBloc = AppBlocsProvider.of<UserProfileBloc>(context);
+      _isInit = true;
+    }
+  }
+
+  void _updateSecurityModel(SecurityModel securityModel, {String pinCode, bool secureBackupWithPin, bool delete = false}) {
```
roeierez: Why also delete as parameter? we can just pass the right values instead.

lib/routes/shared/security_pin/security_pin_page.dart
```diff
@@ -0,0 +1,172 @@
...
+                children: _buildSecurityPINTiles(snapshot.data.securityModel),
+              ),
+            );
+          }
+        });
+  }
+
+  List<Widget> _buildSecurityPINTiles(SecurityModel securityModel) {
+    List<Widget> _tiles = List();
+    final _disablePINTile = _buildDisablePINTile(securityModel);
```
roeierez: We don't need the variables here, we can simplify by creating an initial list with the disabledPINTile and then add the rest of the tiles in case `securityModel.pinCode != null`
If variables are needed then without underscore.

lib/routes/shared/security_pin/security_pin_page.dart
erdemyerebasmaz: I agree this is not a good solution, I'm open to suggestions but this was the reasoning:

We decided to use single user action that updates security model (UpdateSecurityModel).
If a parameter is not provided to this function, we are setting it to it's value in the current security model.
To delete pinCode we pass it as null, which would we passed as it's current value to the UpdateSecurityModel user action if we don't set delete param to true.

roeierez: I am suggesting this: 
```suggestion
  void _updateSecurityModel(SecurityModel securityModel) {
     _userProfileBloc.userActionsSink.add(updateSecurityModelAction);
  }
```
Or even better call directly
`_userProfileBloc.userActionsSink.add(updateSecurityModelAction);` from the context it is needed

lib/bloc/user_profile/user_profile_bloc.dart
```diff
@@ -123,6 +143,27 @@ class UserProfileBloc {
...
+    _userActionsController.stream.listen((action) {
+      var handler = _actionHandlers[action.runtimeType];
+      if (handler != null) {
+        handler(action).catchError((e) => action.resolveError(e));
+      }
+    });
+  }
+
+  Future _updateSecurityModel(UpdateSecurityModel updateSecurityModelAction) async {
+    if (updateSecurityModelAction.pinCode != null && _currentUser.securityModel.pinCode != updateSecurityModelAction.pinCode) {
```
roeierez: This is buggy since if the new pinCode equals the current pinCode we will delete the pinCode.

lib/bloc/user_profile/breez_user_model.dart
roeierez: I understand. But we don't need to persist it in JSON nor read it from there.

lib/routes/shared/security_pin/prompt_pin_code.dart
```diff
@@ -0,0 +1,174 @@
...
+    _setErrorMessage("");
+    if (_enteredPinCode.length < PIN_CODE_LENGTH) {
+      _setPinCodeInput(_enteredPinCode + numberText);
+    }
+    if (_enteredPinCode.length == PIN_CODE_LENGTH) {
+      if ((widget.changePin) && _tmpPinCode.isEmpty) {
```
roeierez: I can see now the next simplification:
I think it is more readable to handle each use case in its own logical block:
```suggestion
      if ((widget.changePin) {
         _handleChangePin()
      } else {
         _handleUnlockPin()
      }
      
      void _handleChangePin(){
         if(_tmpPinCode.isEmpty) {
            // set tmpPinCode
         } else {
            // validate change pin code
         }
      }
```

lib/routes/shared/security_pin/prompt_pin_code.dart
```diff
@@ -0,0 +1,178 @@
...
+    _setLabel("Re-enter your new PIN");
+    _setPinCodeInput("");
+  }
+
+  void _matchPinCodes(bool isMatched) {
+    if (isMatched) {
+      _setPinCode(_enteredPinCode);
+    } else {
+      _setLabel("Enter your new PIN");
+      _setPinCodeInput("");
```
roeierez: This method always called with empty string, better to change it to: "clearPinCode"

lib/routes/shared/security_pin/prompt_pin_code.dart
```diff
@@ -0,0 +1,178 @@
...
+    });
+    _setLabel("Re-enter your new PIN");
+    _setPinCodeInput("");
+  }
+
+  void _matchPinCodes(bool isMatched) {
+    if (isMatched) {
+      _setPinCode(_enteredPinCode);
+    } else {
+      _setLabel("Enter your new PIN");
```
roeierez: In case the pin code doesn't match shouldn't we revert to the first state and clear the `_tmpPinCode`?
Currently the `_tmpPinCode ` contains the old value

lib/routes/user/home/home_page.dart
```diff
@@ -97,12 +101,18 @@ class HomeState extends State<Home> {
...
   Set _hiddenRountes = Set<String>();
   StreamSubscription<String> _accountNotificationsSubscription;
+  StreamSubscription _accountLockSubscription;
 
   @override
   void initState() {
     super.initState();
 
-    _registerNotificationHandlers();
+    _accountLockSubscription = widget.userProfileBloc.userStream.listen((user) {
```
roeierez:
```suggestion
    _accountLockSubscription = widget.userProfileBloc.userStream.where(a => !a.waitigForPin).first().then(
        _registerNotificationHandlers();
    )
```

lib/bloc/user_profile/user_profile_bloc.dart
```diff
@@ -123,6 +142,28 @@ class UserProfileBloc {
...
+    _userActionsController.stream.listen((action) {
+      var handler = _actionHandlers[action.runtimeType];
+      if (handler != null) {
+        handler(action).catchError((e) => action.resolveError(e));
+      }
+    });
+  }
+
+  Future _updateSecurityModel(UpdateSecurityModel updateSecurityModelAction) async {
+    if (updateSecurityModelAction.pinCode != null) {
```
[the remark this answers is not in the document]
erdemyerebasmaz: Fixed

lib/bloc/user_profile/user_profile_bloc.dart
```diff
@@ -123,6 +142,28 @@ class UserProfileBloc {
...
+      // Write to storage if the new pin code is different from current pin code
+      if (updateSecurityModelAction.pinCode != _currentUser.securityModel.pinCode) await _secureStorage.write(key: 'pinCode', value: updateSecurityModelAction.pinCode);
+    } else {
+      await _secureStorage.delete(key: 'pinCode');
+    }
+    SecurityModel _securityModel = _currentUser.securityModel.copyWith(
```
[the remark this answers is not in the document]
erdemyerebasmaz: Fixed

lib/bloc/user_profile/user_actions.dart
```diff
@@ -0,0 +1,8 @@
+import 'package:breez/bloc/async_action.dart';
+
+class UpdateSecurityModel extends AsyncAction {
+  final String pinCode;
```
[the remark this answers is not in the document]
erdemyerebasmaz: [Fixed](https://github.com/breez/breezmobile/pull/128/commits/6ac1c1b634f9d41637f7c31e88e8456d167c93a9)

lib/routes/shared/security_pin/prompt_pin_code.dart
```diff
@@ -0,0 +1,178 @@
...
+import 'package:breez/theme_data.dart' as theme;
+import 'package:breez/widgets/back_button.dart' as backBtn;
+import 'package:breez/widgets/pin_code_widget.dart';
+import 'package:breez/widgets/route.dart';
+import 'package:flutter/material.dart';
+
+const PIN_CODE_LENGTH = 6;
+
+class LockScreen extends StatefulWidget {
+  final String label;
```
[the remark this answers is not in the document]
erdemyerebasmaz: [Fixed](https://github.com/breez/breezmobile/pull/128/commits/ecbbac808361de0002e70984bd6e68b45e16479d)

lib/routes/user/home/home_page.dart
erdemyerebasmaz: [Fixed](https://github.com/breez/breezmobile/pull/128/commits/23db3903b80c7d72706ad0291163e44e414305c6)

lib/routes/shared/security_pin/prompt_pin_code.dart
erdemyerebasmaz: [Fixed](https://github.com/breez/breezmobile/pull/128/commits/1471db5e2a95d320271a0deb1e2fb8c734541e84)

lib/bloc/user_profile/user_profile_bloc.dart
```diff
@@ -123,6 +142,28 @@ class UserProfileBloc {
...
+  Future _updateSecurityModel(UpdateSecurityModel updateSecurityModelAction) async {
+    if (updateSecurityModelAction.newModel.pinCode == null) {
+      await _secureStorage.delete(key: 'pinCode');
+    } else if (updateSecurityModelAction.newModel.pinCode != _currentUser.securityModel.pinCode) {
+      // Write to storage if the new pin code is different from current pin code
+      await _secureStorage.write(key: 'pinCode', value: updateSecurityModelAction.newModel.pinCode);
+    }
+    SecurityModel securityModel = _currentUser.securityModel.copyWith(
+        pinCode: updateSecurityModelAction.newModel.pinCode, secureBackupWithPin: updateSecurityModelAction.newModel.secureBackupWithPin);
+    _userController.add(_currentUser.copyWith(securityModel: securityModel));
```
roeierez: Why not just push the new model?
```suggestion
    _userController.add(_currentUser.copyWith(securityModel: updateSecurityModelAction.newModel));
```

PR #130, opened 2019-07-31
POS: refactor payment dialog

icota:
- Extract to a separate file and rename from NfcDialog to PosPaymentDialog
- Remove NFC specific code
- Change behaviour so it's not triggered when receiving null on `_readyInvoicesController` stream

### DIFF
--- a/lib/routes/pos/home/pos_invoice.dart
+++ b/lib/routes/pos/home/pos_invoice.dart
@@ -76,9 +76,6 @@ class POSInvoiceState extends State<POSInvoice> {
   void registerListeners() {    
     _focusNode = new FocusNode();
     _focusNode.addListener(_onOnFocusNodeEvent);
-
-    PosPaymentDialog _posPaymentDialog =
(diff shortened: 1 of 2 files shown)


lib/routes/pos/home/pos_invoice.dart
```diff
@@ -103,13 +102,14 @@ class POSInvoiceState extends State<POSInvoice> {
...
-        builder: (BuildContext context) {
-          return _nfcDialog;
-        }).then((result) {
-        setState(() {
+      if (invoiceReady != null) {
+        showDialog<bool>(
+            context: context,
+            barrierDismissible: false,
+            builder: (BuildContext context) {
+              return _posPaymentDialog;
```
roeierez: I prefer to write here: `return PosPaymentDialog(_invoiceBloc, _posProfileBloc, _scaffoldKey); `
and get rid of the `_posPaymentDialog ` variable.

icota: I think that's better too. I just didn't bother to change the old approach but will do now.

PR #134, opened 2019-08-05
inform user of on chain confirmation

gotcha:
Let user know that opening a channel depends on on chain confirmation and might need quite some time

### DIFF
--- a/lib/bloc/account/account_model.dart
+++ b/lib/bloc/account/account_model.dart
@@ -234,7 +234,7 @@ class AccountModel {
     }
 
     if (this.processingBreezConnection) {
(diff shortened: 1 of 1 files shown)


kingonly:
I would rather not provide time estimation in the message. I think "might take a while" suffices...
Not sure adding this additional sentence, which also doesn't look nice on low res devices, helps much...

gotcha:
@kingonly Take a while is pretty vague. 
I mainly opened this PR to trigger a discussion about the right wording. 
What I'd like to see is some info about the on chain transaction as it explains that "take a while" can be actually quite long.

kingonly:
Or short... Most of the channels are opened in the next block. The right fix here is to link to the on-chain tx information as we did in Add Funds.
You are welcome to try and this information. It's on our todos.

kingonly:
Added funding tx info a while ago.

PR #140, opened 2019-08-06
Wait for last digit color fill animation on ChangePinCode & RestorePin screens before validation

erdemyerebasmaz:
Apply [BU-171](https://breeztech.atlassian.net/browse/BU-171) changes to ChangePinCode & RestorePin screens


lib/routes/shared/security_pin/change_pin_code.dart
```diff
@@ -49,24 +49,26 @@ class _ChangePinCodeState extends State<ChangePinCode> {
...
-        setState(() {
-          _tmpPinCode = _enteredPinCode;
-          _label = "Re-enter your new PIN";
-          _enteredPinCode = "";
-        });
-      } else {
-        if (_enteredPinCode == _tmpPinCode) {
-          Navigator.pop(context, _enteredPinCode);
-        } else {
+      Future.delayed(Duration(milliseconds: 200), () {
```
roeierez: @erdemyerebasmaz instead of duplicating this code over all pin code variations what do you think about changing the `onNumButtonPressed` to be called after the animation delay?
This way we also don't need to guess the duration here.

erdemyerebasmaz: If we add it to ```onNumButtonPressed```, there's nothing to be animated because _enteredPinCode has not changed yet.
We only need to delay when the _enteredPinCode.length == PIN_CODE_LENGTH condition is met because that's where we reset the pin input/navigate elsewhere.

roeierez: I see...
In that case I think the PinCodeWidget should handle all the "single letter" events as it already knows the required length and render all the circles.
How about removing the `onNumButtonPressed` event and only expose `onPinEntered` event after the animation?
This will reduce some duplicate code from all the three pin code widgets and will require to convert the PinCodeWidget to be stateful.

erdemyerebasmaz: Fixed 👍

lib/routes/shared/security_pin/lock_screen.dart
```diff
@@ -43,50 +40,27 @@ class _AppLockScreenState extends State<AppLockScreen> {
...
-      });
+  _onPinEntered(String enteredPinCode) {
+    if (enteredPinCode == widget.securityModel.pinCode) {
+      if (widget.onUnlock != null) {
+        widget.onUnlock();
+        return;
+      }
+      Navigator.pop(context, true);
+    } else {
+      _key.currentState.setPinCodeInput("");
```
roeierez: Can we implement the clear input inside the `PinCodeWidget` ? In a way that we clear the value every time before calling `onPinEntered`. It will save us maintaining this in all widgets.

lib/routes/shared/security_pin/lock_screen.dart
```diff
@@ -43,50 +40,27 @@ class _AppLockScreenState extends State<AppLockScreen> {
...
+      _key.currentState.setErrorMessage("Incorrect PIN");
```
roeierez: Also maybe the callback `onPinEntered` can throw an exception in case it desires to change the error message.
It will save us using the global key to manipulate the `PinCodeWidget` state.

PR #162, opened 2019-08-27
Add Moonpay to Add Funds

erdemyerebasmaz:
This PR implements [BU-178](https://breeztech.atlassian.net/browse/BU-178) & [BU-180](https://breeztech.atlassian.net/browse/BU-180).

An [intermediary screen](https://zpl.io/VKykvAq) is implemented to reduce clutter in Add Funds screen.
Users are now able to buy bitcoins using MoonPay vendor. Clicking Buy Bitcoin button routes users to moonpay's website using a Vebview, provided with various parameters to ease the process.
Such as currencyCode, walletAddress, maxQuoteCurrencyAmount etc.

Once the order is complete, it's information is passed to AddFundsBloc. If the address associated with the transaction is in unConfirmedAddresses "Your MoonPay order is being processed." message is shown. Which will expire once the transaction is confirmed or after a set amount of time has passed.

TODO:
 - Determine the expiration time,
 - Set apiKey to publishable key and test on production,
 - ~Determine which error messages to be removed from deposit_to_btc_address_page~


lib/bloc/account/account_model.dart
```diff
@@ -20,34 +20,39 @@ enum SyncUIState {
...
-            showConnectProgress: json["showConnectProgress"] ?? false);
+            showConnectProgress: json["showConnectProgress"] ?? false,
+            moonpayIpCheck: json["moonpayIpCheck"] ?? true,
+            failePaymentBehavior: BugReportBehavior.values[json["failePaymentBehavior"] ?? 0]);
 
   Map<String, dynamic> toJson() {
     return {
       "ignoreWalletBalance": ignoreWalletBalance,
+      "showConnectProgress": showConnectProgress ?? false,
+      "moonpayIpCheck": moonpayIpCheck ?? true,
```
roeierez: We don't need the null fallback here for "`moonpayIpCheck`" and "`showConnectProgress`". They will always have value.

lib/bloc/account/add_funds_bloc.dart
```diff
@@ -11,19 +14,65 @@ class AddFundsBloc {
   final _addFundResponseController = new StreamController<AddFundResponse>();
   Stream<AddFundResponse> get addFundResponseStream => _addFundResponseController.stream;
 
+  final _moonPayOrderController = new StreamController<MoonpayOrder>();
+  Sink<MoonpayOrder> get moonPayOrderSink => _moonPayOrderController.sink;
```
roeierez: Why do we need additional MoonpayOrder controller (sink and stream)?
Can't we use the `_orderController ` also for pending orders?

lib/bloc/account/add_funds_bloc.dart
```diff
@@ -11,19 +14,65 @@ class AddFundsBloc {
...
           .then((reply) => _addFundResponseController.add(new AddFundResponse(reply)))
           .catchError(_addFundResponseController.addError);
       })
       .onDone(_dispose);
+
+    _initializeWithMoonpayOrder(injector);
+    _listenToMoonpayOrder(injector);
+  }
+
+  _initializeWithMoonpayOrder(ServiceInjector injector) {
```
roeierez: I would simplify this function by always adding the last `MoonpayOrder` to the stream and no need to delete it from the storage as well.
Something like:
```suggestion
  var order = preferences.getString("pendingMoonpayOrderAddress");
  if (order != null) {
      _moonPayOrderController.add(MoonpayOrder(
            preferences.getString("pendingMoonpayOrderAddress") ?? null, 
            preferences.getInt("pendingMoonpayOrderTimestamp") ?? null))
   }
```

The UI only needs to filter those expired (already does that).

lib/bloc/account/add_funds_bloc.dart
```diff
@@ -11,19 +14,65 @@ class AddFundsBloc {
...
+    _orderController.stream.listen((order) async {
+      injector.sharedPreferences.then((preferences) {
+        _saveMoonpayOrder(preferences, order);
+        _moonPayOrderController.add(order);
+      });
+    });
+  }
+
+  _saveMoonpayOrder(SharedPreferences preferences, MoonpayOrder order) {
+    preferences.setString("pendingMoonpayOrderAddress", order.address);
```
roeierez: I think it is better that MoonpayOrder class will handle its own json serialization and we can write its value under one key.

lib/routes/user/add_funds/add_funds_page.dart
```diff
@@ -43,71 +56,106 @@ class AddFundsState extends State<AddFundsPage> {
         _accountSubscription.cancel();
       }
     });
+    _settingsSubscription = widget._accountBloc.accountSettingsStream.listen((settings) {
+      isIpAllowed(settings.moonpayIpCheck);
+    });
+    _moonPaySubscription = _addFundsBloc.moonPayOrderStream.listen((order) {
```
roeierez: I don't think we need a subscription here, just use the stream in the build method.

lib/routes/user/add_funds/add_funds_page.dart
```diff
@@ -43,71 +56,106 @@ class AddFundsState extends State<AddFundsPage> {
...
+      setState(() {
+        _moonPayOrder = order;
+      });
+    });
+    _checkMoonpayOrderExpiration();
+  }
+
+  isIpAllowed(bool ipCheckSetting) async {
```
roeierez: all ip checking and the decision if to show moonpay or not should be in the bloc.
Also the moonpay URL should be generated in the bloc.
I think we can add moonpayURL in AddFundsResponse model and show the moonpay menu only if this field has value.
This way we can encapsulate the ip logic in the bloc without making the UI complex.

lib/routes/user/add_funds/add_funds_page.dart
```diff
@@ -120,121 +168,195 @@ class AddFundsState extends State<AddFundsPage> {
...
-          text: hasError ? "RETRY" : "CLOSE",
-          onPressed: () {
-            if (hasError) {
-              _addFundsBloc.addFundRequestSink.add(null);
-            } else {
-              Navigator.of(context).pop();
-            }
-          });
+  Widget _buildMoonpayButton(AddFundResponse response) {
+    String baseUrl = "https://buy-staging.moonpay.io";
```
roeierez: The moonpay url should be generated in the bloc. Also the `apiKey` should be fetched from a config file.

lib/routes/user/add_funds/add_funds_page.dart
```diff
@@ -120,121 +168,195 @@ class AddFundsState extends State<AddFundsPage> {
...
+    String apiKey = "pk_test_AZskxvTXb0rpsI7o2GCdmzs8jeST9d";
+    String currencyCode = "btc";
+    String walletAddress = "n4VQ5YdHf7hLQ2gWQYYrcxoE5B7nWuDFNF";
```
roeierez: Hard coded address...

lib/routes/user/marketplace/vendor_webview.dart
```diff
@@ -62,53 +64,85 @@ class VendorWebViewPageState extends State<VendorWebViewPage> {
...
                 });
               }
-            });            
-        }
-      });
+            });
+          }
+        });
+      } else {
+        String loadedURL;
```
roeierez: I don't think the internals of moonpay logic belongs here as they are not related to VendorWebView.
I also see that most of the logic in this class is not relevant to moonpay as well so I think it is better to create a separate widget "MoonpayWebview" with the specific logic.

lib/routes/user/marketplace/vendor_webview.dart
```diff
@@ -62,53 +64,85 @@ class VendorWebViewPageState extends State<VendorWebViewPage> {
...
+        _widgetWebview.onStateChanged.listen((state) async {
+          if (state.type == WebViewState.finishLoad && loadedURL != state.url) {
+            loadedURL = state.url;
+            var script = "function getParameterByName(name) {    \n" +
```
roeierez: The script should be in a resource file.

lib/routes/user/add_funds/add_funds_page.dart
erdemyerebasmaz: This is hardcoded for test mode, will be changed to current address when we switch to production.

lib/routes/user/marketplace/vendor_webview.dart
erdemyerebasmaz: [Fixed](https://github.com/breez/breezmobile/pull/162/commits/ec80a3164e05ea547c90c5d39dd519e0563b431e)

lib/bloc/account/account_model.dart
erdemyerebasmaz: [Fixed](https://github.com/breez/breezmobile/pull/162/commits/9bb1eabe77c2447ccb86553ccd0231efbb76737f)

lib/routes/user/marketplace/vendor_webview.dart
erdemyerebasmaz: [Fixed](https://github.com/breez/breezmobile/pull/162/commits/52282b8d816a7acb283a541d8d78a892bdb550ff)

lib/routes/user/add_funds/add_funds_page.dart
erdemyerebasmaz: Fixed
[1](https://github.com/breez/breezmobile/pull/162/commits/e2cbce256ee9aeba8b56b024e9eb6c650dbbb913)
[2](https://github.com/breez/breezmobile/pull/162/commits/2c95697004ad429b7857223898348aa9ab1ec3c0)

lib/bloc/account/add_funds_bloc.dart
erdemyerebasmaz: [Fixed](https://github.com/breez/breezmobile/pull/162/commits/96486619e86debc7d6514689a54adbb2e76a63bd)

lib/bloc/account/add_funds_bloc.dart
erdemyerebasmaz: [Fixed](https://github.com/breez/breezmobile/pull/162/commits/8f9ff8cb02826657a80bf08ddb67b2dd7453a0f4)

lib/bloc/account/add_funds_bloc.dart
erdemyerebasmaz: It's removed in [9648661](https://github.com/breez/breezmobile/pull/162/commits/96486619e86debc7d6514689a54adbb2e76a63bd)

lib/routes/user/add_funds/add_funds_page.dart
erdemyerebasmaz: [Removed in](https://github.com/breez/breezmobile/pull/162/commits/e575a05e69ee9d79c65ba65bc54d6f1b73380a1c)

lib/routes/user/add_funds/add_funds_page.dart
erdemyerebasmaz: [Fixed](https://github.com/breez/breezmobile/pull/162/commits/e575a05e69ee9d79c65ba65bc54d6f1b73380a1c)

lib/bloc/account/add_funds_bloc.dart
```diff
@@ -1,29 +1,107 @@
...
+      Map<String, dynamic> settings = json.decode(accountSettings);
+      if (settings["moonpayIpCheck"]) {
+        _isIpAllowed().then((ipIsAllowed) {
+          if (ipIsAllowed) _createMoonpayUrl();
+        });
+      }
+    }
+  }
+
+  Future<bool> _isIpAllowed() async {
```
roeierez: NIT: rename to `_isIPMoonpayAllowed` ?

lib/bloc/account/add_funds_bloc.dart
```diff
@@ -1,29 +1,107 @@
...
+    }).onDone(_dispose);
+    _listenAccountSettings(injector);
+    _handleMoonpayOrders(injector);
+  }
+
+  String get moonPayURL => _moonPayURL;
+
+  _listenAccountSettings(ServiceInjector injector) async {
+    var preferences = await injector.sharedPreferences;
+    var accountSettings = preferences.getString(ACCOUNT_SETTINGS_PREFERENCES_KEY);
```
roeierez: Ideally this value should be saved at the same place it is read from.
I would move the save action from the AccountBloc to the AddFundsBloc.

lib/routes/user/add_funds/add_funds_page.dart
```diff
@@ -43,71 +51,89 @@ class AddFundsState extends State<AddFundsPage> {
...
-    } else if (unconfirmedTxID?.isNotEmpty == true ||
-        account.processingWithdrawal) {
+      errorMessage = 'You\'d be able to add funds after Breez is finished bootstrapping.';
+    } else if (unconfirmedTxID?.isNotEmpty == true || account.processingWithdrawal) {
       errorMessage =
           'Breez is processing your previous ${waitingDepositConfirmation || account.processingBreezConnection ? "deposit" : "withdrawal"}. You will be able to add more funds once this operation is completed.';
     } else if (response != null && response.errorMessage.isNotEmpty) {
       errorMessage = response.errorMessage;
+    } else if (_moonPayOrder?.address != null &&
+        account?.addedFundsReply?.unConfirmedAddresses
```
roeierez: Although the `addedFundsReply` is public field it wasn't the intention as it is a protobuf structure.
The access should be via the `SwapFundStatus` that wraps it.
I would add a getter there for `unconfirmedAddresses` same as the `refundableAddresses` that exists there.

lib/routes/user/add_funds/add_funds_page.dart
```diff
@@ -120,121 +146,185 @@ class AddFundsState extends State<AddFundsPage> {
...
-          onPressed: () {
-            if (hasError) {
-              _addFundsBloc.addFundRequestSink.add(null);
-            } else {
-              Navigator.of(context).pop();
-            }
-          });
+  Widget _buildMoonpayButton(AddFundResponse response) {
+    String walletAddress = "n4VQ5YdHf7hLQ2gWQYYrcxoE5B7nWuDFNF"; // Will switch to response?.address when we use public apiKey
+    String moonPayURL = _addFundsBloc.moonPayURL + "&walletAddress=$walletAddress";
```
roeierez: It is not a practice in our blocs to expose public getter. Instead the blocs provide the data via their stream interface.

lib/bloc/account/add_funds_bloc.dart
erdemyerebasmaz: [Renamed to _isIPMoonpayAllowed](https://github.com/breez/breezmobile/pull/162/commits/55c902c244bd6e62994f90a9df275198276d0b73)

lib/routes/user/add_funds/add_funds_page.dart
erdemyerebasmaz: https://github.com/breez/breezmobile/pull/162/commits/71dc2301a1f97ae791929615644062c444d2ab5b

lib/routes/user/add_funds/add_funds_page.dart
erdemyerebasmaz: https://github.com/breez/breezmobile/pull/162/commits/eb57e859db19800afe4054b9d63ef1bcb0617671

PR #173, opened 2019-09-11
Backup Phrase

erdemyerebasmaz:
This PR implements [BU-186](https://breeztech.atlassian.net/browse/BU-186)

Use PIN for Backup/Restore option in Security page has been removed and replaced with Generate Backup Phrase. Users can secure their backup with  BIP39 mnemonic through this option. This change is backward compatible, existing users that uses pin for backup will still be able to use that option.

The generation flow consists of three pages,
- Confirmation, an informative text is displayed and user confirms to their responsibilities
- Generation, where the bip39 mnemonic will be generated and shown to the user
- Verification, user is asked to input 3 of those words

If the user has set a backup phrase, the user will be redirected to enter backup phrase page and asked to input all 24 words. If the phrases match, the restore process will continue.


erdemyerebasmaz:
Auto-complete feature has been added.

erdemyerebasmaz:
- 'Generate Backup Phrase' is renamed to 'Use Backup Phrase'.
- Backup Phrase is decoupled from PIN, users can use backup phrase without having set a PIN.
- Removing backup phrase triggers backup.
- Instruction text's font size is increased on Confirmation page and checbox label is fixed for larger fonts.
- Going back from Let's Verify screen generates a new seed and a new set of numbers is selected for verification.
- Instruction text on Let's Verify screen is centered between form and bottom button.
- Form validation errors are replaced with an error message on Let's Verify screen. Restore error messages are added to Enter Backup Phrase screen.
- Error message is displayed right under form for Let's Verify screen and for Enter Backup Phrase screen, it's displayed between form an bottom button.
- Changed wording of Let's Verify instruction text.

lib/routes/shared/security_pin/backup_phrase/verify_backup_phrase_page.dart
```diff
@@ -0,0 +1,220 @@
...
+      _randomlySelectedIndexes.add(list[i]);
+    }
+    _randomlySelectedIndexes.sort();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    UserProfileBloc userProfileBloc = AppBlocsProvider.of<UserProfileBloc>(context);
+    BackupBloc backupBloc = AppBlocsProvider.of<BackupBloc>(context);
+    return WillPopScope(
```
roeierez: This way seems a bit odd and I think the requirements were misunderstood.
When user goes back from "Let's Verify" we need to show the previous screens as is, not regenerate the mnemonics at this stage.
If the user then enters the "Let's Verify" again (and each time he enters this screen) we need to pick newly random words.

lib/routes/shared/security_pin/backup_phrase/verify_backup_phrase_page.dart
```diff
@@ -0,0 +1,220 @@
...
+        body: Column(
+          mainAxisAlignment: MainAxisAlignment.spaceBetween,
+          children: <Widget>[
+            _buildForm(),
+            _buildInstructions(),
+            StreamBuilder<BreezUserModel>(
+              stream: userProfileBloc.userStream,
+              builder: (context, snapshot) {
+                if (!snapshot.hasData) {
+                  return Container(padding: EdgeInsets.only(bottom: 88));
```
roeierez:
```suggestion
                  return SizedBox();
```

lib/routes/shared/security_pin/backup_phrase/verify_backup_phrase_page.dart
```diff
@@ -158,7 +167,7 @@ class VerifyBackupPhrasePageState extends State<VerifyBackupPhrasePage> {
             elevation: 0.0,
             shape: const StadiumBorder(),
             onPressed: () {
-              if (_formKey.currentState.validate()) {
+              if (_formKey.currentState.validate() && !_hasError) {
```
roeierez: Shouldn't we reset` _hasError = false` before calling `_formKey.currentState.validate()`

lib/bloc/user_profile/user_profile_bloc.dart
```diff
@@ -188,20 +207,39 @@ class UserProfileBloc {
...
 
+  Future _createBackupPhrase(CreateBackupPhrase action) async {
+    await _secureStorage.write(key: 'backupKey', value: bip39.mnemonicToEntropy(action.backupPhrase));
+    action.resolve(null);
+  }
+
+  Future _validateBackupPhrase(ValidateBackupPhrase action) async {
+    var backupPhrase = await _secureStorage.read(key: 'backupKey');
+    String enteredBackupPhrase;
+    try {
```
roeierez: We don't need try/catch here. This handler is invoked within an action and better to only throw exception here which will make the action resolve to error.
In the UI we can catch error to handle validation failed.

lib/routes/shared/security_pin/backup_phrase/generate_backup_phrase_page.dart
```diff
@@ -0,0 +1,143 @@
...
+              style: theme.appBarTextStyle,
+            ),
+            elevation: 0.0),
+        body: _buildMnemonicSeedList(),
+        bottomNavigationBar: _buildNextBtn(),
+      ),
+    );
+  }
+
+  _onWillPop(BuildContext context) {
```
roeierez: Isn't it better to push each page to the stack and let the back button and native back behaves naturally?

lib/routes/shared/initial_walkthrough.dart
```diff
@@ -63,16 +70,23 @@ class InitialWalkthroughPageState extends State<InitialWalkthroughPage>
...
       if (toRestore != null) {
         if (toRestore.encrypted) {
-          restorePIN = await getRestorePIN();
-          if (restorePIN == null) {
+          if (toRestore.encryptionType == "Mnemonics") {
+            restoreKey = await getBackupPhrase();
+            if (restoreKey != null) _updateSecurityModelAction = UpdateSecurityModel(widget._user.securityModel.copyWith(backupKeyType: BackupKeyType.PHRASE));
+          } else if (toRestore.encryptionType == "Pin") {
+            restoreKey = await getRestorePIN();
+            if (restoreKey != null) _updateSecurityModelAction = UpdateSecurityModel(widget._user.securityModel.copyWith(backupKeyType: BackupKeyType.PIN));
```
roeierez: At both cases where the restoreKey is given the `backupKeyType` is changes correctly but the key is not saved in the secure storage.

lib/bloc/user_profile/security_model.dart
```diff
@@ -1,28 +1,34 @@
...
   static const int _defaultLockInterval = 120;
 
-  final bool requiresPin;  
-  final bool secureBackupWithPin;
+  final bool requiresPin;
+  final BackupKeyType backupKeyType;
   final int automaticallyLockInterval;
 
-  SecurityModel({this.requiresPin, this.secureBackupWithPin, this.automaticallyLockInterval});
+  SecurityModel({this.requiresPin, this.backupKeyType, this.automaticallyLockInterval});
```
roeierez: Better makes this private constructor.

erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/173/commits/1aac4e35785b3b6da90e92c822696485bf9953e8)

lib/routes/shared/security_pin/backup_phrase/verify_backup_phrase_page.dart
erdemyerebasmaz: [We do](https://github.com/breez/breezmobile/commit/a3e77fbe8f9ddbdba67d41599da0179db000d0ef), this file seems outdated.

lib/routes/shared/security_pin/backup_phrase/verify_backup_phrase_page.dart
erdemyerebasmaz: Reverted the changes [here](https://github.com/breez/breezmobile/pull/173/commits/87f74be999f3fb63ef0f74eaa9a93f858acd5498)
_(There's minor formatting changes)_

lib/routes/shared/initial_walkthrough.dart
```diff
@@ -217,7 +242,10 @@ class InitialWalkthroughPageState extends State<InitialWalkthroughPage>
                               return BetaWarningDialog();
                             }).then((approved) {
                           if (approved) {
-                            _proceedToRegister();
+                            widget._registrationBloc.userActionsSink.add(_updateSecurityModelAction);
```
roeierez: Better not to save `_updateSecurityModelAction ` as class variable. Just create an action here for resetting the key.

lib/routes/shared/initial_walkthrough.dart
```diff
@@ -43,6 +49,7 @@ class InitialWalkthroughPageState extends State<InitialWalkthroughPage>
     _instructions = widget._isPos ?
     "The simplest, fastest & safest way\nto earn bitcoin" :
     "The simplest, fastest & safest way\nto spend your bitcoins";
+    _updateSecurityModelAction = UpdateSecurityModel(widget._user.securityModel.copyWith(backupKeyType: BackupKeyType.NONE));
```
roeierez: Better not save this as class variable, just create it on `initState` when needed.

lib/routes/shared/initial_walkthrough.dart
```diff
@@ -89,7 +103,10 @@ class InitialWalkthroughPageState extends State<InitialWalkthroughPage>
...
       popToWalkthrough();
       if (restored) {
-        _proceedToRegister();
+        widget._registrationBloc.userActionsSink.add(_updateSecurityModelAction);
+        _updateSecurityModelAction.future.then((_) {
+          _proceedToRegister();
+        });
       }
     }, onError: (error) {
       popToWalkthrough(
```
roeierez: When en error happens, for example when the phrase wasn't correct, the user would have to choose the restore snapshot again and re-enter the phrase again, which is tedious and bad experience.
Better NOT to pop the phrase/pin screens when restoring and push the loader route on top of them.
On error just pop the loader route and let the user change the pin/phrase or go back and start all over again.

lib/routes/shared/initial_walkthrough.dart
erdemyerebasmaz: [Fixed](https://github.com/breez/breezmobile/pull/173/commits/17d400a70e8decc8aada4810b6d092c33b2f8054)

lib/bloc/user_profile/user_profile_bloc.dart
erdemyerebasmaz: This method is now obsolete and removed [here](https://github.com/breez/breezmobile/pull/173/commits/d8b131b0ffa327af188f8ce98a6581d2305d12ec)

lib/routes/shared/initial_walkthrough.dart
erdemyerebasmaz: It's used on multiple streams. It's value is being set on multipleRestoreStream and used after successful restore on restoreFinishedStream.

lib/routes/shared/initial_walkthrough.dart
erdemyerebasmaz: Removed the usage of _updateSecurityModelAction variable [here](https://github.com/breez/breezmobile/pull/173/commits/4c12fa91203d1b146b9a1ad1678c2ebafec1a1e2)

lib/routes/shared/initial_walkthrough.dart
```diff
@@ -63,16 +71,26 @@ class InitialWalkthroughPageState extends State<InitialWalkthroughPage>
...
-      String restorePIN;
+      String restoreKey;
       if (toRestore != null) {
         if (toRestore.encrypted) {
-          restorePIN = await getRestorePIN();
-          if (restorePIN == null) {
+          if (toRestore.encryptionType == "Mnemonics") {
+            restoreKey = await getBackupPhrase();
+            if (restoreKey != null) {
+              _createBackupPhrase(restoreKey);
```
roeierez: missing `await` here?

lib/routes/shared/initial_walkthrough.dart
```diff
@@ -108,6 +131,22 @@ class InitialWalkthroughPageState extends State<InitialWalkthroughPage>
...
+      builder: (BuildContext context) {
+        return EnterBackupPhrasePage();
+      },
+    ));
+  }
+
+  Future _createBackupPhrase(String _mnemonics) async {
+    var createBackupPhraseAction = CreateBackupPhrase(_mnemonics);
+    widget._registrationBloc.userActionsSink.add(createBackupPhraseAction);
+    createBackupPhraseAction.future.catchError((err) {
```
roeierez: missing `return` here?

lib/routes/shared/security_pin/backup_phrase/verify_backup_phrase_page.dart
```diff
@@ -0,0 +1,199 @@
...
+    _randomlySelectedIndexes.sort();
+  }
+
+  Future _updateSecurityModel(
+      SecurityModel oldModel, SecurityModel newModel, UserProfileBloc userProfileBloc, BackupBloc backupBloc) async {
+    var action = UpdateSecurityModel(newModel);
+    userProfileBloc.userActionsSink.add(action);
+    Navigator.popUntil(context, ModalRoute.withName("/security"));
+    action.future.then((_) {
+      backupBloc.backupNowSink.add(true);
```
roeierez: The following logic somehow looks duplicate as it exist in the parent page `SecurityPage`.
Can we just pop here and return the security model to the parent page and from there use the regular flow of updating the model, backup, etc... ?

lib/routes/shared/initial_walkthrough.dart
```diff
@@ -70,31 +71,40 @@ class InitialWalkthroughPageState extends State<InitialWalkthroughPage>
...
+              var updateAction = UpdateSecurityModel(widget._user.securityModel.copyWith(backupKeyType: BackupKeyType.PHRASE));
+              widget._registrationBloc.userActionsSink.add(updateAction);
+              restore(toRestore, entrophy);
+            });
             return;
           }
+          
+          if (toRestore.encryptionType == "Pin") {
+            restoreUsingPIN((pin) async {                            
+              var updateAction = UpdateSecurityModel(widget._user.securityModel.copyWith(backupKeyType: BackupKeyType.NONE));
```
erdemyerebasmaz: Shouldn't backupKeyType be BackupKeyType.PIN?

erdemyerebasmaz: @roeierez Shouldn't backupKeyType be BackupKeyType.PIN?

roeierez: In case of PIN restore we don't want to update the backup encryption type as this is legacy.

PR #191, opened 2019-10-10
Dark theme progress indicator

erdemyerebasmaz:
A separate gif is used for progress indicator on dark theme.


erdemyerebasmaz:
> Also here I think a better solution is to implement a custom map in the Theme (call it customData) where we can put the custom theme data on these edge cases. Here the code can be something like this:
> `colorBlendMode: Themes.customData[currentTheme].loaderColorBlendMode ?? <default_blend_mode>`

We need access to currentTheme globally.

roeierez:
> First, we need access to currentTheme globally. Currently we either need to listen userStream or pass the user itself to children to know the themeId.

That's right.
On Pages getting the theme from the userprofile looks good to me.
On widgets it is better to pass them the theme name.

erdemyerebasmaz:
theme_data and themes files are joined together.
themeId global variable is introduced(as a  workaround) to access current themeId where necessary.
On Breez Date Picker, current day was black on dark theme which was not very visible. It has been changed to white.

Issue #194, opened 2019-10-12
Minor UI problem on dark mode


bitcoinuser:
Font on top (battery, clock...) should be white color like Telegram app and not white. See the difference:

![photo4976591396744046628](https://user-images.githubusercontent.com/1178856/66706044-b55e0280-ed04-11e9-8787-3cc3a6a4f8d0.jpg)

![photo4976773761055434847](https://user-images.githubusercontent.com/1178856/66706057-df172980-ed04-11e9-9941-09f07229cf47.jpg)

kingonly:
Fixed

PR #195, opened 2019-10-12
Change status bar icon and text color to white

erdemyerebasmaz:
This PR fixes status bar color issue described on https://github.com/breez/breezmobile/issues/194

Both themes use white colored status bar icon and text instead of the previous black value.

### DIFF
--- a/lib/routes/user/home/home_page.dart
+++ b/lib/routes/user/home/home_page.dart
@@ -13,6 +13,7 @@ import 'package:breez/routes/user/connect_to_pay/connect_to_pay_page.dart';
 import 'package:breez/routes/user/ctp_join_session_handler.dart';
 import 'package:breez/routes/user/received_invoice_notification.dart';
 import 'package:breez/routes/user/showPinHandler.dart';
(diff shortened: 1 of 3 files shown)


lib/routes/user/home/home_page.dart
```diff
@@ -134,6 +135,7 @@ class HomeState extends State<Home> {
         child: new Scaffold(
             key: _scaffoldKey,
             appBar: new AppBar(
+              brightness: theme.themeId == "BLUE" ? Brightness.light : Theme.of(context).appBarTheme.brightness,
```
roeierez: Why instead not assigning `Brightness.light` to the dark theme?
Also according to documentation we don't need to `assign AppBar.brightness` explicitly as it is taken from the theme automatically: https://api.flutter.dev/flutter/material/AppBar/brightness.html

erdemyerebasmaz: This is a special case, on blue theme the wallet dashboard's background is white and status bar text & icon would not be visible with brightness defined. in app theme.

PR #210, opened 2019-11-15
Add biometrics support

erdemyerebasmaz:
This PR addresses [BU-197](https://breeztech.atlassian.net/browse/BU-197).

Users will now be able to use enrolled biometrics on their device to unlock the app.
If biometrics auth is enabled, the user will be prompted to authenticate with biometrics on lock screens. If the user cancels, they can access the biometric auth via the button on right bottom corner of the numpad, which will convert to erase button if the user starts typing their pin.

TODO:
- Check enrolled biometrics onResume and update users security model if biometrics have been removed.  This should also be checked when initializing the user on startup, rather than pin code widget's initState.
- Test on iOS devices, preferably both Touch ID and Face ID


lib/widgets/pin_code_widget.dart
```diff
@@ -22,13 +24,38 @@ class PinCodeWidgetState extends State<PinCodeWidget> {
...
+    if (widget.onFingerprintEntered != null) {
+      _availableBiometric = "";
+      _initLocalAuthService();
+    }
     _enteredPinCode = "";
     _errorMessage = "";
   }
 
+  Future _initLocalAuthService() async {
+    _localAuthService = LocalAuthenticationService(context);
```
roeierez: `LocalAuthenticationService` should be exposed only via ServiceInjector and all the biometric authentication logic, including the use of this service, should be inside the bloc that may expose a specific action for that.

lib/bloc/user_profile/user_profile_bloc.dart
```diff
@@ -160,6 +163,22 @@ class UserProfileBloc {
...
+    _deviceService.eventStream.listen((e){
+      if (e == NotificationType.RESUME) {
+        _checkBiometrics();
+      }
+    });
+  }
+
+  Future _checkBiometrics() async {
+    String enrolledBiometrics = await _localAuthService.enrolledBiometrics;
+    _updateSecurityModel(UpdateSecurityModel(_currentUser.securityModel.copyWith(
```
roeierez: Just to make sure this means that we reset the persistent user selection to "not use fingerprint" in case the device settings has changed. Is this the requirement?

erdemyerebasmaz: Yes, this resets the user setting.

It prevents users to bypass app security by defining a new fingerprint.

1- User A has fingerprint auth enabled on app
2- User A deletes pin/fingerprint from device(for some reason)
3- User A's device gets into hand of a person with malicious intent, User B
4- User B opens Breez App and is required to enter pin 

If we didn't reset this setting, User B could define a new fingerprint and unlock the Breez App with that newly defined fingerprint.

roeierez: I actually don't read the code this way. The code resets the setting in case there is no fingerprint defined in the device.
In the case you wrote the requirement was to reset in case there was a **change** in the fingerprint but it is still defined (by user B).

erdemyerebasmaz: We are resetting the setting on stage 4, if we didn't the user could define a new fingerprint and bypass app security.
It is still possible to bypass app security by defining the new fingerprint before opening the app but we will detect the change in fingerprint and prevent this altogether.

lib/bloc/user_profile/user_profile_bloc.dart
```diff
@@ -201,6 +220,14 @@ class UserProfileBloc {
     return action.newTheme;
   }
 
+  Future _getEnrolledBiometrics(GetEnrolledBiometrics action) async {
```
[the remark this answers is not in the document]
erdemyerebasmaz: [Removed](https://github.com/breez/breezmobile/pull/210/commits/a33a10afe25343d6978b664de215b5198adb6726)

lib/widgets/pin_code_widget.dart
```diff
@@ -141,21 +159,51 @@ class PinCodeWidgetState extends State<PinCodeWidget> {
...
-        )
+          widget.onFingerprintEntered == null || ((widget.onFingerprintEntered != null) && _enteredPinCode.length > 0)
+              ? _buildEraseButton()
+              : StreamBuilder<BreezUserModel>(
+                  stream: widget.userProfileBloc.userStream,
+                  builder: (context, snapshot) {
+                    if (!snapshot.hasData || snapshot.data.securityModel.enrolledBiometrics == "") {
+                      return _buildEraseButton();
+                    } else {
+                      if (!_isInit && widget.onFingerprintEntered != null) {
```
roeierez: This should't be called from the build method which should only render widgets. initState is the right place.

erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/210/commits/015c2dfdc9c7eb1f90f158184895e30bc8a3efb2)

lib/widgets/pin_code_widget.dart
```diff
@@ -27,9 +34,24 @@ class PinCodeWidgetState extends State<PinCodeWidget> {
     super.initState();
     _enteredPinCode = "";
     _errorMessage = "";
+    if (widget.onFingerprintEntered != null) {
+      widget.userProfileBloc.userStream.firstWhere((u) => u.securityModel.enrolledBiometrics != "").then((_) {
+        _validateBiometrics();
```
roeierez: Look like this is triggered on init and not cancelled. I can see some weird behavior when someone open this screen, then close it and the biometric validation is still in progress...
Later on it could trigger a flush bar on error or something unexpected.

lib/bloc/user_profile/security_model.dart
```diff
@@ -2,27 +2,33 @@ class SecurityModel {
...
   SecurityModel.fromJson(Map<String, dynamic> json)
       : requiresPin = json['requiresPin'] ?? false,
+        isFingerprintEnabled = json['isFingerprintEnabled'] ?? false,
+        enrolledBiometrics = json['enrolledBiometrics'] ?? "",
         automaticallyLockInterval = json['automaticallyLockInterval'] ?? _defaultLockInterval;
 
   Map<String, dynamic> toJson() => {
+        'requiresPin': requiresPin,
+        'isFingerprintEnabled': isFingerprintEnabled,
+        'enrolledBiometrics': enrolledBiometrics,
```
roeierez: Any reason we should persist this field?

lib/bloc/user_profile/user_profile_bloc.dart
```diff
@@ -160,6 +162,23 @@ class UserProfileBloc {
...
+      }
+    });
+  }
+
+  Future _checkBiometrics() async {
+    String enrolledBiometrics = await _localAuthService.enrolledBiometrics;
+    if (enrolledBiometrics == "" || enrolledBiometrics != _currentUser.securityModel.enrolledBiometrics) {
+      _updateSecurityModel(UpdateSecurityModel(_currentUser.securityModel.copyWith(isFingerprintEnabled: false)));
+    }
+    _updateSecurityModel(UpdateSecurityModel(_currentUser.securityModel.copyWith(enrolledBiometrics: enrolledBiometrics)));
```
roeierez: We don't need to take care of biometric settings changes.
Maybe this is better?
```
bool isFingerprintEnabled = _currentUser.securityModel.isFingerprintEnabled && enrolledBiometrics.isNotEmpty
_updateSecurityModel(UpdateSecurityModel(_currentUser.securityModel.copyWith(enrolledBiometrics: enrolledBiometrics, isFingerprintEnabled: isFingerprintEnabled)));
```

lib/widgets/pin_code_widget.dart
roeierez: We also need to put some delay here since the screen is shown with fade transition and the touch id is shown immediately halting the UI causes the screen to be  half transparent.

roeierez: I pushed this change.

erdemyerebasmaz: Looks good

lib/bloc/user_profile/security_model.dart
erdemyerebasmaz: No, it's not required to persist thi data as it's being set on each startup & onResume.

lib/bloc/user_profile/user_profile_bloc.dart
erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/210/commits/7efadec40ee2b25b11fb4791048cee014edf998c)

lib/bloc/user_profile/security_model.dart
erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/210/commits/7efadec40ee2b25b11fb4791048cee014edf998c)

lib/widgets/pin_code_widget.dart
erdemyerebasmaz: [We listen to first item available to avoid unwanted behaviour on Security & Backup page](https://github.com/breez/breezmobile/pull/210/commits/5b360eeda4c134c5b432eaa2aa0bd11d6b8c8014)

PR #225, opened 2019-12-02
Set enrolled biometrics when initializing user

erdemyerebasmaz:
This PR follows up on [Add biometrics support ](https://github.com/breez/breezmobile/pull/210), [Update biometrics](https://github.com/breez/breezmobile/pull/223) and fixes issues:
- There's no fingertip menu after activating the pin right after fresh install [BU-208](https://breeztech.atlassian.net/browse/BU-208)
- The fingerprint menu doesn't prompt automatically [BU-209](https://breeztech.atlassian.net/browse/BU-209)

These issues emerged because [this commit](https://github.com/breez/breezmobile/commit/2ab78b5bfa79b17a8e9bb3c9f675dc0004982885#diff-942b641f4ad9455c1856ac3fa442d243) was overlooked during merge.

### DIFF
--- a/lib/bloc/user_profile/user_profile_bloc.dart
+++ b/lib/bloc/user_profile/user_profile_bloc.dart
@@ -151,30 +151,41 @@ class UserProfileBloc {
             color: randomName[0],
             animal: randomName[1]);
       }
-
       user = user.copyWith(locked: user.securityModel.requiresPin);
+      await _checkBiometrics().then(
+        (enrolledBiometrics) {
+          bool isFingerprintEnabled =
+              (user.securityModel.isFingerprintEnabled &&
+                  enrolledBiometrics.isNotEmpty);
+          user = user.copyWith(
+              securityModel: user.securityModel.copyWith(
+                  isFingerprintEnabled: isFingerprintEnabled,
+                  enrolledBiometrics: enrolledBiometrics));
+        },
+      );
       _publishUser(user);
     });
   }
 
   _updateBiometricsSettings() {
-    _checkBiometrics();
     _deviceService.eventStream.listen((e) {
       if (e == NotificationType.RESUME) {
-        _checkBiometrics();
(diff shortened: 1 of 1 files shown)


lib/bloc/user_profile/user_profile_bloc.dart
```diff
@@ -151,30 +151,41 @@ class UserProfileBloc {
             color: randomName[0],
             animal: randomName[1]);
       }
-
       user = user.copyWith(locked: user.securityModel.requiresPin);
+      await _checkBiometrics().then(
```
roeierez: If using await then it is better to do:
`enrolledBiometrics  = await _checkBiometrics();`

lib/bloc/user_profile/user_profile_bloc.dart
```diff
@@ -151,30 +151,41 @@ class UserProfileBloc {
...
-  Future _checkBiometrics() async {
-    String enrolledBiometrics = await _localAuthService.enrolledBiometrics;
-    bool isFingerprintEnabled =
-        _currentUser.securityModel.isFingerprintEnabled &&
-            enrolledBiometrics.isNotEmpty;
-    _updateSecurityModel(UpdateSecurityModel(_currentUser.securityModel
-        .copyWith(
-            enrolledBiometrics: enrolledBiometrics,
-            isFingerprintEnabled: isFingerprintEnabled)));
+  Future<String> _checkBiometrics() async {
```
roeierez: NIT: Looks like we don't need this function?

lib/bloc/user_profile/user_profile_bloc.dart
erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/225/commits/deda31cf880894cfd5741e2c9cb8cd0b11fdc52f)

PR #230, opened 2019-12-04
Prevent drag gesture on biometrics switch

erdemyerebasmaz:
This PR addresses [BU-205](https://breeztech.atlassian.net/browse/BU-205).

Switch widget -most likely- has an unresolved bug at the moment. I will open an issue to flutter team and take action after their respond. For now, drag gesture is prevented on this specific switch.

The issue: [#46046](https://github.com/flutter/flutter/issues/46046) _(Feel free to simplify the wording...)_

### DIFF
--- a/lib/routes/shared/security_pin/security_pin_page.dart
+++ b/lib/routes/shared/security_pin/security_pin_page.dart
@@ -297,7 +297,7 @@ class SecurityPageState extends State<SecurityPage> {
         group: securityModel.requiresPin ? _autoSizeGroup : null,
       ),
       trailing: Switch(
-        key: Key("switch $_renderIndex"),
+        key: Key(_renderIndex.toString()),
         value: securityModel.isFingerprintEnabled,
         onChanged: (value) {
           if (this.mounted) {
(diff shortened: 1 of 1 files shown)


kingonly:
Why not setting it manually to 'off' when the user cancel the auth dialog?

erdemyerebasmaz:
https://github.com/breez/breezmobile/pull/230#issuecomment-561464436

@kingonly because it already is off when cancelled, just not visually... the switch's state does not update even after a hot reload. I explained the problem in more detail on the issue I opened on Flutter repo.

kingonly:
I saw your issue. That's why I think we could find a better workaround... setting it manually at some point or rendering one control and then replacing it with another after the auth dialog.

erdemyerebasmaz:
https://github.com/breez/breezmobile/pull/230#issuecomment-561467961

True, I had a similar idea in mind. I didn't want to introduce spaghetti code before I verify if this is the expected behaviour.

kingonly:
Let's wait for Roei's inpout

roeierez:
@erdemyerebasmaz I see the render bug in flutter.
You can work-around it by forcing render on the switch using a key that changes on every render call:

_key: Key("switch ${renerIndex++}")_

It will force the switch to re-render the correct value.

erdemyerebasmaz:
> @erdemyerebasmaz I see the render bug in flutter.
> You can work-around it by forcing render on the switch using a key that changes on every render call:
> 
> _key: Key("switch ${renerIndex++}")_
> 
> It will force the switch to re-render the correct value.

[Done](https://github.com/breez/breezmobile/pull/230/commits/e7db33d02626be7a0bee15470ecb5cdaa24ab1c2)

PR #231, opened 2019-12-04
Biometric prompt visual improvements

erdemyerebasmaz:
This PR introduces several visual improvements to biometric prompt flow.

 - Biometric prompt is shown faster.
 - On lockscreen, blank circles are filled to indicate that the fingerprint was entered.
 - Lockscreen waits for biometric prompt to disappear before popping.

### DIFF
--- a/lib/widgets/pin_code_widget.dart
+++ b/lib/widgets/pin_code_widget.dart
@@ -52,21 +52,12 @@ class PinCodeWidgetState extends State<PinCodeWidget> {
         ValidateBiometrics(localizedReason: widget.localizedReason);
     widget.userProfileBloc.userActionsSink.add(validateBiometricsAction);
     validateBiometricsAction.future.then((isValid) async {
-      if (isValid) {
-        do {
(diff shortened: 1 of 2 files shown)


lib/widgets/pin_code_widget.dart
```diff
@@ -39,7 +39,7 @@ class PinCodeWidgetState extends State<PinCodeWidget> {
       widget.userProfileBloc.userStream.first.then(
         (user) async {
           if (user.securityModel.enrolledBiometrics != "") {
-            await Future.delayed(Duration(milliseconds: 500));
+            await Future.delayed(Duration(milliseconds: 100));
```
roeierez: This interval should be more than the fade transition interval in FadeInRoute otherwise in iOS the page would be seen semi transparent.
This happens because the ui freezes immediately once the biometrics prompt is shown.

erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/231/commits/cb2c537092b947c249de3102aa89562edb1da91e)

PR #232, opened 2019-12-11
Do not reset enrolledBiometrics before proceeding to register

erdemyerebasmaz:
This issue addresses [BU-208](https://breeztech.atlassian.net/browse/BU-208).

SecurityModel was being reverted to it's original state before proceeding to register, which caused  fingertip menu to not show up on fresh installs unless users restarted or resumed the app.

enrolledBiometrics is now updated before processing register requests.

### DIFF
--- a/lib/bloc/user_profile/user_profile_bloc.dart
+++ b/lib/bloc/user_profile/user_profile_bloc.dart
@@ -4,6 +4,7 @@ import 'dart:io';
 
 import 'package:breez/bloc/async_action.dart';
 import 'package:breez/bloc/user_profile/breez_user_model.dart';
+import 'package:breez/bloc/user_profile/security_model.dart';
 import 'package:breez/bloc/user_profile/currency.dart';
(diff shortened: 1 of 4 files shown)


lib/bloc/user_profile/user_profile_bloc.dart
```diff
@@ -240,6 +245,9 @@ class UserProfileBloc {
 
   void _listenRegistrationRequests(ServiceInjector injector) {
     _registrationController.stream.listen((request) async {
+      await _updateEnrolledBiometrics();
+      _updateSecurityModel(UpdateSecurityModel(SecurityModel.initial().copyWith(
```
roeierez: This is problematic.
Moving the reset security model to the registration point creates a side effect where it is also happens after restore. It will reset the PIN and backup mnemonics that have been assigned right after the restore.
We should keep the reset only from the "Let's Breez" button. If necessary we can implement an action only for that (ResetSecurityModel).

PR #241, opened 2019-12-22
Track closed channels

roeierez:
This PR follows https://github.com/breez/breez/pull/103 and adds the UI section for the closed channels list items.
In addition to the list item, in the case the channel is pending we show a  warning icon with sweep estimation time.


lib/widgets/payment_details_dialog.dart
```diff
@@ -312,3 +337,77 @@ _buildSnackBar(String item) {
...
+  const ClosedChannelPaymentDetails({Key key, this.closedChannel})
+      : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    if (!closedChannel.pending) {
+      return RichText(
+          text: TextSpan(
+              style: Theme.of(context).dialogTheme.contentTextStyle,
+              text: "Trasnfer to local wallet due to ",
```
fiatjaf: Typo here.

roeierez: Thanks. fixed.

PR #254, opened 2020-01-06
Get enrolled biometrics on device on demand

erdemyerebasmaz:
We retrieve enrolled biometrics on device on demand rather than once every start up and each on resume. This reduces load on computing overhead and ensures we have the latest information.

enrolledBiometrics is removed from SecurityModel and is now being updated on onResume, on UI level.

### DIFF
--- a/lib/widgets/pin_code_widget.dart
+++ b/lib/widgets/pin_code_widget.dart
@@ -55,8 +55,8 @@ class PinCodeWidgetState extends State<PinCodeWidget>
 
   @override
   void dispose() {
-    super.dispose();
     WidgetsBinding.instance.removeObserver(this);
+    super.dispose();
   }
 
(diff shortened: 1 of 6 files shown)


lib/widgets/pin_code_widget.dart
```diff
@@ -26,27 +26,53 @@ class PinCodeWidget extends StatefulWidget {
...
-        },
-      );
+      _promptBiometrics();
     }
   }
 
+  @override
+  void didChangeAppLifecycleState(AppLifecycleState state) {
+    if (state == AppLifecycleState.resumed) {
+      if (widget.onFingerprintEntered != null) {
```
roeierez: `onFingerprintEntered` is true when the settings are true but the check for if the biometrics are enables is missing here, isn't it?

Issue #243, opened 2019-12-22
Minor UI issue in "Redeem FastBitcoins Voucher" screen


bolatovumar:
Currency dropdown is misaligned with the "Voucher Value" field. It's not a big deal but annoys me.

![Screenshot_20191222-105950](https://user-images.githubusercontent.com/1934678/71327329-008e5180-24bc-11ea-98ed-bdd5bb5fe799.png)

erdemyerebasmaz:
Issue persists on large screen devices ![image](https://user-images.githubusercontent.com/4012752/74238344-a903a980-4ce6-11ea-9e11-c1b693628247.png)

kingonly:
Fixed in latest

PR #267, opened 2020-02-11
Fix misalignment between Voucher Value and Currency dropdown

erdemyerebasmaz:
This PR addresses https://github.com/breez/breezmobile/issues/243

Currency dropdown is now aligned with Voucher Value field.


erdemyerebasmaz:
Closes #243

PR #132, opened 2019-08-02
POS: Add fiat support

icota:
- Add fiat currencies to the dropdown list
- Show decimals depending on `currencyData.fractionSize`
- Convert to last used bitcoin unit and display on the charge button
- Refactor duplicate code into `updateAmountControllers` and `clearCurrentAmounts`

PR #268, opened 2020-02-11
POS: Add fiat support 

erdemyerebasmaz:
Moved https://github.com/breez/breezmobile/pull/132 into this branch. (_because I messed up resolving conflicts and wanted to save time rather than resolve it on that branch_)

- Add fiat currencies to the dropdown list
- Show decimals depending on `currencyData.fractionSize`
- Convert to last used bitcoin unit and display on the charge button
- Refactor duplicate code into `updateAmountControllers` and `clearCurrentAmounts`

- Recognize "sats" as Sat currency using Currency.fromSymbol().

### DIFF
--- a/lib/bloc/app_blocs.dart
+++ b/lib/bloc/app_blocs.dart
@@ -46,7 +46,7 @@ class AppBlocs {
     AccountBloc accountBloc =
         _registerBloc(AccountBloc(userProfileBloc.userStream), blocsByType);
     InvoiceBloc invoicesBloc =
-        _registerBloc(InvoiceBloc(userProfileBloc), blocsByType);
+        _registerBloc(InvoiceBloc(), blocsByType);
     ConnectPayBloc connectPayBloc = _registerBloc(
         ConnectPayBloc(userProfileBloc.userStream, accountBloc.accountStream,
             accountBloc.userActionsSink),
--- a/lib/bloc/invoice/actions.dart
+++ b/lib/bloc/invoice/actions.dart
@@ -1,0 +1,9 @@
+import 'package:breez/bloc/async_action.dart';
+
+import 'invoice_model.dart';
+
+class NewInvoice extends AsyncAction {
+  final InvoiceRequestModel request;
+
+  NewInvoice(this.request);
+}
--- a/lib/bloc/invoice/invoice_bloc.dart
+++ b/lib/bloc/invoice/invoice_bloc.dart
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'package:breez/bloc/invoice/actions.dart';
 import 'package:breez/bloc/invoice/invoice_model.dart';
 import 'package:breez/bloc/user_profile/user_profile_bloc.dart';
 import 'package:breez/services/injector.dart';
@@ -14,7 +15,9 @@ import 'package:fixnum/fixnum.dart';
 import 'package:breez/services/lightning_links.dart';
 import 'package:breez/services/device.dart';
 
-class InvoiceBloc {
+import '../async_actions_handler.dart';
+
+class InvoiceBloc with AsyncActionsHandler {
   final _newInvoiceRequestController = StreamController<InvoiceRequestModel>();
   Sink<InvoiceRequestModel> get newInvoiceRequestSink =>
       _newInvoiceRequestController.sink;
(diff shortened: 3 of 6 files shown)


lib/routes/pos/home/pos_invoice.dart
```diff
@@ -390,56 +434,87 @@ class POSInvoiceState extends State<POSInvoice> {
...
   onNumButtonPressed(String numberText) {
     setState(() {
-      _currentAmount = (_currentAmount * 10) + int.parse(numberText);
-      _amountController.text = _currency.format((Int64(_currentAmount)),
-          fixedDecimals: true, includeSymbol: false);
-      _chargeAmountController.text = _currency.format(
-          (Int64(_totalAmount + _currentAmount)),
-          fixedDecimals: true,
-          includeSymbol: true);
+      if (!_usingFiat) {
```
roeierez: I wonder if we can maintain only the string representation in a class variable and implement both _currentFiatAmount & _currentAmount as calculated properties.

lib/routes/pos/home/pos_invoice.dart
```diff
@@ -390,56 +434,87 @@ class POSInvoiceState extends State<POSInvoice> {
...
   changeCurrency(value) {
     setState(() {
-      _currency = Currency.fromSymbol(value);
-      _userProfileBloc.currencySink
-          .add(Currency.currencies[Currency.currencies.indexOf(_currency)]);
+      bool usingFiat = _fiatCurrencyConversion != null ? true : false;
+      Currency.currencies.forEach((currency) {
+        if (currency.displayName == value) {
+          usingFiat = false;
+          _currency = Currency.fromSymbol(value);
```
roeierez:
```suggestion
          _currency = currency;
```

lib/routes/pos/home/pos_invoice.dart
```diff
@@ -390,56 +434,87 @@ class POSInvoiceState extends State<POSInvoice> {
...
+          if (_usingFiat) {
+            // We are switching back from fiat
+            _usingFiat = false;
+            _clearCurrentAmounts();
+            _totalAmount = 0;
+          }
+          _userProfileBloc.currencySink
+              .add(Currency.currencies[Currency.currencies.indexOf(_currency)]);
```
roeierez:
```suggestion
              .add(_currency);
```

lib/routes/pos/home/pos_invoice.dart
```diff
@@ -259,43 +256,90 @@ class POSInvoiceState extends State<POSInvoice> {
...
+                                                          color:
+                                                              Theme.of(context)
+                                                                  .textTheme
+                                                                  .headline
+                                                                  .color),
+                                                  items: Currency.currencies
+                                                      .map((Currency value) {
+                                                    return DropdownMenuItem<
+                                                        String>(
+                                                      value: value.displayName,
```
roeierez:
```suggestion
                                                      value: value.symbol,
```

lib/routes/pos/home/pos_invoice.dart
```diff
@@ -390,56 +434,87 @@ class POSInvoiceState extends State<POSInvoice> {
...
+      _updateAmountControllers();
     });
   }
 
   changeCurrency(value) {
     setState(() {
-      _currency = Currency.fromSymbol(value);
-      _userProfileBloc.currencySink
-          .add(Currency.currencies[Currency.currencies.indexOf(_currency)]);
+      bool usingFiat = _fiatCurrencyConversion != null ? true : false;
```
roeierez: Could this be null? Because we don't check it in other places at this file.

lib/routes/pos/home/pos_invoice.dart
```diff
@@ -390,56 +434,87 @@ class POSInvoiceState extends State<POSInvoice> {
...
+      Currency.currencies.forEach((currency) {
```
roeierez: If we pass the symbol from the dropdown, we don't have to iterate all.

erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/268/commits/d0ffcf54a357ddac6ea495a979e616f2024361df)

lib/routes/pos/home/pos_invoice.dart
erdemyerebasmaz: It can be null if getExchangeRates fails.
[Removed that logic](https://github.com/breez/breezmobile/pull/268/commits/4cfbca6e14e6a9038239d47a42bf0721ef0049ac)

lib/routes/pos/home/pos_invoice.dart
```diff
@@ -517,3 +578,46 @@ class POSInvoiceState extends State<POSInvoice> {
         ]).toList());
   }
 }
+
+class Amount {
```
roeierez: I am OK with have the parsing logic encapsulated in a separate class but I think this class should hold only:
1. The text entered by the user (as string variable)
2. The fiat conversion or the bitcoin currency.

It should provide calculated property for the text we show the user (after parsing).
This will simplify a lot of code at the widget level and put all the parsing logic here.

PR #277, opened 2020-02-23
Navigation Drawer flavor changes

erdemyerebasmaz:
This PR addresses [POS-13](https://breeztech.atlassian.net/browse/POS-13) and [POS-27](https://breeztech.atlassian.net/browse/POS-27)

Removed Marketplace from drawer in POS mode.
Switch between flavors from a menu item  with a toggle in Advanced tab.


lib/widgets/navigation_drawer.dart
```diff
@@ -18,9 +18,14 @@ class DrawerItemConfig {
...
   final bool disabled;
   final void Function(String name) onItemSelected;
+  final void Function() onToggle;
+  final toggleValue;
 
   DrawerItemConfig(this.name, this.title, this.icon,
-      {this.onItemSelected, this.disabled = false});
+      {this.onItemSelected,
+      this.disabled = false,
+      this.onToggle,
```
roeierez: Isn't it better to pass a builder function (that builds widget) here instead of _onToggle_ & _toggleValue_?

erdemyerebasmaz: [You are right](https://github.com/breez/breezmobile/pull/277/commits/cabf313c64cff64ba70ac765fdacd7d132f217ba)

PR #278, opened 2020-02-23
Pos consolidation regressions

erdemyerebasmaz:
This PR addresses regressions caused by #273

Regressions
- Disable charge button when entering invoice note.
- Do not convert currencies when flipping fiat vice versa.
- Show currency conversion of entered fiat amount on Charge button

Fixes
- Change 'SAT' to 'SATS' in the Charge [POS-28](https://breeztech.atlassian.net/browse/POS-28)
- Dark theme settings are applied to pos buttons.

### DIFF
--- a/lib/routes/charge/pos_invoice.dart
+++ b/lib/routes/charge/pos_invoice.dart
@@ -518,22 +518,18 @@ class POSInvoiceState extends State<POSInvoice> {
   }
 
   String _formattedCurrentCharge(AccountModel acc) {
-    return _formattedCharge(acc, currentAmount);
+    if (_useFiat) {
+      return (currentAmount)
+          .toStringAsFixed(acc.fiatCurrency.currencyData.fractionSize);
+    }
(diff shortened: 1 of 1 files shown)


lib/routes/charge/pos_invoice.dart
```diff
@@ -294,6 +300,12 @@ class POSInvoiceState extends State<POSInvoice> {
     );
   }
 
+  _onOnFocusNodeEvent() {
+    setState(() {
+      _isButtonDisabled = true;
```
roeierez: When the `_isButtonDisabled` going back to false?

erdemyerebasmaz: Whenever you click outside the keyboard.
```
Scaffold(
      resizeToAvoidBottomPadding: false,
      body: new GestureDetector(
        onTap: () {
          // call this method here to hide soft keyboard
          FocusScope.of(context).requestFocus(new FocusNode());
          setState(() {
            _isButtonDisabled = false;
          });
        },
```

PR #279, opened 2020-02-23
Payment request dialog icon

erdemyerebasmaz:
This PR addresses [POS-31](https://breeztech.atlassian.net/browse/POS-31)

The payeeImageURL was not recognised for certain cases such as a POS invoice. The icon widget is replaced with BreezAvatar so the payeeImageURL is recognised and the proper widget is selected for rendering the icon.

The maxHeight is increased after taking _buildErrorMessage into consideration.

### DIFF
--- a/lib/widgets/payment_request_info_dialog.dart
+++ b/lib/widgets/payment_request_info_dialog.dart
@@ -71,7 +71,7 @@ class PaymentRequestInfoDialogState extends State<PaymentRequestInfoDialog> {
       child: Container(
           key: _dialogKey,
           width: MediaQuery.of(context).size.width,
-          constraints: BoxConstraints(minHeight: 220.0, maxHeight: 320.0),
+          constraints: BoxConstraints(minHeight: 220.0, maxHeight: 350.0),
           child: Column(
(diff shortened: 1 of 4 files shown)


lib/widgets/payment_request_info_dialog.dart
```diff
@@ -82,34 +82,13 @@ class PaymentRequestInfoDialogState extends State<PaymentRequestInfoDialog> {
...
-                      height: 64.0,
-                      placeholder: MemoryImage(widget._transparentImage),
-                      image: AdvancedNetworkImage(widget.invoice.payeeImageURL,
-                          useDiskCache: true),
-                      fadeOutDuration: Duration(milliseconds: 200),
-                      fadeInDuration: Duration(milliseconds: 200)),
-                )),
-              ],
-            ));
+        : Padding(
```
roeierez: Don't we want the loader experience as was before? Or the BreezAvatar already have this effect?

erdemyerebasmaz: BreezAvatar has this effect.

PR #280, opened 2020-02-23
POS dark mode adjustments

erdemyerebasmaz:
This PR addresses [POS-30](https://breeztech.atlassian.net/browse/POS-30)

Payment Approved icon is now white with blue tick on dark mode.

### DIFF
--- a/lib/theme_data.dart
+++ b/lib/theme_data.dart
@@ -87,7 +87,7 @@ final ThemeData blueTheme = ThemeData(
         height: 1.28,
         fontFamily: 'IBMPlexSansMedium'),
     body1: TextStyle(
-        color: Color.fromRGBO(0, 133, 251, 1.0),
+        color: BreezColors.blue[900],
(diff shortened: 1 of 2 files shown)


lib/theme_data.dart
```diff
@@ -87,7 +87,7 @@ final ThemeData blueTheme = ThemeData(
         height: 1.28,
         fontFamily: 'IBMPlexSansMedium'),
     body1: TextStyle(
-        color: BreezColors.blue[900],
+        color: Color.fromRGBO(0, 133, 251, 1.0),
```
roeierez: Is this something requires going over all the UI? How can we make sure the body1 value is OK for other widgets? or if even used?

erdemyerebasmaz: Changed to 
`theme.themeId == "BLUE"
                                ? Color.fromRGBO(0, 133, 251, 1.0)
                                : Colors.white`

PR #282, opened 2020-02-24
POS Currency conversion

erdemyerebasmaz:
This PR addresses [POS-34](https://breeztech.atlassian.net/browse/POS-34)

Switching currencies clears the current input and charge amount is converted.

`accountModel.fiatConversionList.firstWhere((f) => f.currencyData.shortName == value)` is used when converting from fiat -> fiat and sat -> fiat because of the delay in `userProfileBloc.fiatConversionSink.add(value);` to be reflected to the account model.

### DIFF
--- a/lib/routes/charge/pos_invoice.dart
+++ b/lib/routes/charge/pos_invoice.dart
@@ -389,12 +389,9 @@ class POSInvoiceState extends State<POSInvoice> {
       AccountModel accountModel, value, UserProfileBloc userProfileBloc) {
     setState(() {
       Currency currency = Currency.fromSymbol(value);
-      FiatConversion newFiatCurrency;
       if (currency != null) {
         userProfileBloc.currencySink.add(currency);
       } else {
-        newFiatCurrency = accountModel.fiatConversionList
-            .firstWhere((f) => f.currencyData.shortName == value);
         userProfileBloc.fiatConversionSink.add(value);
       }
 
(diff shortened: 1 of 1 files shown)


lib/routes/charge/pos_invoice.dart
```diff
@@ -396,20 +396,22 @@ class POSInvoiceState extends State<POSInvoice> {
       }
 
       bool flipFiat = _useFiat == (currency != null);
+      Int64 oldSatAmount = _satAmount(accountModel, amount);
       if (flipFiat) {
         _useFiat = !_useFiat;
-        _clearAmounts(clearTotal: true);
+        _clearAmounts();
       }
+      amount = (flipFiat && _useFiat || _useFiat) // For Sat->Fiat or Fiat->Fiat
```
roeierez: I suggest to replace with this:

```
amount = oldSatAmount.toDouble();

// We need to convert only in case we use fiat.
if (_useFiat) {
    amount = accountModel.fiatCurrency.satToFiat(oldSatAmount);
}
```

erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/282/commits/28d751d7fb4f91cac5ddca6bd005a3214879aab6)
I kept `accountModel.fiatConversionList.firstWhere((f) => f.currencyData.shortName == value)` as accountModel.fiatCurrency is not updated till we reach here.

PR #283, opened 2020-02-24
Avatar picker bug

erdemyerebasmaz:
This PR addresses [POS-38](https://breeztech.atlassian.net/browse/POS-38).

- Choosing a picture often doesn’t work
    - Out of scope. 
   This is the error related to this problem: ”gRPC Error (8, /breez.Pos/UploadLogo is rejected by ratelimit, please retry later.)”
- Choosing a custom picture causing the circle to look elliptic 
   - Set an aspect ratio of 1:1
- When you choose a picture and cancel, when you open a again, the cancelled data remains in the dialog
   - Profile picture now uploads to server when the user saves, not after selecting the image from gallery
- We need to add loader when choosing a custom picture
   - This is not needed anymore as we pass the preview images bytes directly to BreezAvatar

### DIFF
--- a/lib/widgets/breez_avatar.dart
+++ b/lib/widgets/breez_avatar.dart
@@ -47,7 +47,7 @@ class BreezAvatar extends StatelessWidget {
         this.backgroundColor ?? theme.sessionAvatarBackgroundColor;
 
     if (avatarURL != null && avatarURL.isNotEmpty) {
-      if (avatarURL.startsWith("/data")) {
+      if (avatarURL.contains("/image_cropper")) {
         return _FileImageAvatar(radius, avatarURL);
       }
 
--- a/lib/widgets/breez_avatar.dart
+++ b/lib/widgets/breez_avatar.dart
@@ -1,8 +1,10 @@
-import 'package:flutter_advanced_networkimage/provider.dart';
-import 'package:flutter/material.dart';
-import 'package:breez/theme_data.dart' as theme;
+import 'dart:io';
+
 import 'package:breez/bloc/user_profile/default_profile_generator.dart'
     as generator;
+import 'package:breez/theme_data.dart' as theme;
+import 'package:flutter/material.dart';
+import 'package:flutter_advanced_networkimage/provider.dart';
 
 final _breezAvatarColors = {
   "salmon": Color(0xFFFA8072),
@@ -36,21 +38,19 @@ class BreezAvatar extends StatelessWidget {
   final String avatarURL;
   final double radius;
   final Color backgroundColor;
-  final List<int> bytes;
 
-  BreezAvatar(this.avatarURL,
-      {this.radius = 20.0, this.backgroundColor, this.bytes});
+  BreezAvatar(this.avatarURL, {this.radius = 20.0, this.backgroundColor});
 
   @override
   Widget build(BuildContext context) {
     Color avatarBgColor =
         this.backgroundColor ?? theme.sessionAvatarBackgroundColor;
 
-    if (bytes != null) {
-      return _MemoryImageAvatar(radius, bytes);
-    }
-
     if (avatarURL != null && avatarURL.isNotEmpty) {
+      if (avatarURL.startsWith("/data")) {
+        return _FileImageAvatar(radius, avatarURL);
+      }
+
       if (avatarURL.startsWith("breez://profile_image?")) {
         var queryParams = Uri.parse(avatarURL).queryParameters;
         return _GeneratedAvatar(
@@ -116,11 +116,11 @@ class _GeneratedAvatar extends StatelessWidget {
   }
 }
 
-class _MemoryImageAvatar extends StatelessWidget {
+class _FileImageAvatar extends StatelessWidget {
   final double radius;
-  final List<int> bytes;
+  final String filePath;
 
-  _MemoryImageAvatar(this.radius, this.bytes);
+  _FileImageAvatar(this.radius, this.filePath);
 
   @override
   Widget build(BuildContext context) {
@@ -130,7 +130,9 @@ class _MemoryImageAvatar extends StatelessWidget {
       decoration: new BoxDecoration(
         shape: BoxShape.circle,
         image: DecorationImage(
-          image: MemoryImage(bytes),
+          image: FileImage(
+            File(filePath),
+          ),
         ),
       ),
     );
--- a/lib/widgets/breez_avatar.dart
+++ b/lib/widgets/breez_avatar.dart
@@ -57,7 +57,7 @@ class BreezAvatar extends StatelessWidget {
         return _VendorAvatar(radius, avatarURL);
       }
 
-      if (Uri.parse(avatarURL).scheme.startsWith("http")) {
+      if (Uri.tryParse(avatarURL)?.scheme?.startsWith("http") ?? false) {
         return _NetworkImageAvatar(avatarURL, radius);
       }
 
--- a/lib/widgets/breez_avatar.dart
+++ b/lib/widgets/breez_avatar.dart
@@ -47,10 +47,6 @@ class BreezAvatar extends StatelessWidget {
         this.backgroundColor ?? theme.sessionAvatarBackgroundColor;
 
     if (avatarURL != null && avatarURL.isNotEmpty) {
-      if (avatarURL.contains("/image_cropper")) {
-        return _FileImageAvatar(radius, avatarURL);
-      }
-
       if (avatarURL.startsWith("breez://profile_image?")) {
         var queryParams = Uri.parse(avatarURL).queryParameters;
         return _GeneratedAvatar(
@@ -61,7 +57,11 @@ class BreezAvatar extends StatelessWidget {
         return _VendorAvatar(radius, avatarURL);
       }
 
-      return _NetworkImageAvatar(avatarURL, radius);
+      if (Uri.parse(avatarURL).scheme.startsWith("http")) {
+        return _NetworkImageAvatar(avatarURL, radius);
+      }
+
+      return _FileImageAvatar(radius, avatarURL);
     }
 
     return _UnknownAvatar(radius, avatarBgColor);
--- a/lib/widgets/breez_avatar.dart
+++ b/lib/widgets/breez_avatar.dart
@@ -36,13 +36,20 @@ class BreezAvatar extends StatelessWidget {
   final String avatarURL;
   final double radius;
   final Color backgroundColor;
+  final List<int> bytes;
 
-  BreezAvatar(this.avatarURL, {this.radius = 20.0, this.backgroundColor});
+  BreezAvatar(this.avatarURL,
+      {this.radius = 20.0, this.backgroundColor, this.bytes});
 
   @override
   Widget build(BuildContext context) {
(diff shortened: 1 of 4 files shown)


lib/widgets/breez_avatar_dialog.dart
```diff
@@ -11,147 +12,180 @@ import 'package:image/image.dart' as DartImage;
...
 
-  Future _pickImage(BuildContext context) async {
+  Future _pickImage(BuildContext context, setState) async {
     return ImagePicker.pickImage(source: ImageSource.gallery).then((file) {
       ImageCropper.cropImage(
         sourcePath: file.path,
         aspectRatio: CropAspectRatio(ratioX: 1.0, ratioY: 1.0),
       ).then((file) {
         if (file != null) {
           file.readAsBytes().then(scaleAndFormatPNG).then((image) {
```
roeierez: Instead of reading the file to memory isn't it better to use the file url instead?
This way we can use this url in the BreezAvatar instead of adding an extra "imageData" parameter.

lib/widgets/breez_avatar_dialog.dart
```diff
@@ -11,147 +12,180 @@ import 'package:image/image.dart' as DartImage;
...
+  List<int> _imageData;
 
   final _nameInputController = TextEditingController();
   userBloc.userPreviewStream.listen((user) {
     _nameInputController.text = user.name;
     _currentSettings = user;
   });
 
-  Future _pickImage(BuildContext context) async {
+  Future _pickImage(BuildContext context, setState) async {
```
roeierez: Instead of passing `setState` why not return the file url as the future result? Mutating the state can be done by the caller.

lib/widgets/breez_avatar_dialog.dart
```diff
@@ -11,147 +12,180 @@ import 'package:image/image.dart' as DartImage;
...
+          ),
+          FlatButton(
+            child:
+                Text('SAVE', style: Theme.of(context).primaryTextTheme.button),
+            onPressed: () async {
+              if (_imageData != null) {
+                var action = UploadProfilePicture(_imageData);
+                userBloc.userActionsSink.add(action);
+                await action.future.then((_) async {
+                  Navigator.of(context).pop();
```
roeierez: This block is a bit tricky since lots of things can happen during the upload (like user cancel the dialog), the upload will continue in the background and when completed the navigator will pop the current route without checking it is the right route.
Also I suggest to add some UI indication (loading) during the upload.

erdemyerebasmaz: [Prevented user to close dialog or cancel during upload](https://github.com/breez/breezmobile/pull/283/commits/fcb442c9e84dc2aeb449fc677c833418f9eb73b3)

lib/widgets/breez_avatar_dialog.dart
erdemyerebasmaz: [Returned the file and passed the url to BreezAvatar, it's bytes to UploadProfilePicture user action](https://github.com/breez/breezmobile/pull/283/commits/890899a2ba0a5eae80bfe2c3d5f5391d43b6cf25)

lib/widgets/breez_avatar_dialog.dart
```diff
@@ -11,150 +13,249 @@ import 'package:image/image.dart' as DartImage;
...
+      await ImagePicker.pickImage(source: ImageSource.gallery)
+          .then((file) async {
+        await ImageCropper.cropImage(
+          sourcePath: file.path,
+          aspectRatio: CropAspectRatio(ratioX: 1.0, ratioY: 1.0),
+        ).then((file) {
+          if (file != null) {
+            image = file;
+          }
+        });
```
roeierez: The whole function can be simplified:

```suggestion
  Future<File> _pickImage() async {
  return ImagePicker.pickImage(source: ImageSource.gallery).then((file) {
    return ImageCropper.cropImage(
        sourcePath: file.path,
        aspectRatio: CropAspectRatio(ratioX: 1.0, ratioY: 1.0));
  });
}
```

lib/widgets/breez_avatar.dart
```diff
@@ -43,7 +45,12 @@ class BreezAvatar extends StatelessWidget {
   Widget build(BuildContext context) {
     Color avatarBgColor =
         this.backgroundColor ?? theme.sessionAvatarBackgroundColor;
+
     if (avatarURL != null && avatarURL.isNotEmpty) {
+      if (avatarURL.startsWith("/data")) {
```
roeierez: We shouldn't assume "/data" prefix on file urls.
A simple solution is to check first for "breez://profile_image" prefix and if not assume it is file.

lib/widgets/breez_avatar_dialog.dart
```diff
@@ -11,150 +13,249 @@ import 'package:image/image.dart' as DartImage;
...
+                  : () {
+                      Navigator.of(context).pop();
+                    },
+            ),
+            FlatButton(
+              child: Text('SAVE',
+                  style: Theme.of(context).primaryTextTheme.button),
+              onPressed: () async {
+                if (_pickedImage != null) {
+                  await _pickedImage
```
roeierez: It will be nice to refactor the  upload to its own function that returns a future and encapsulate all the details (reading, scaling).
It currently hard to follow the code and I think there are some bugs like if the upload fails it still update the user name, etc...

erdemyerebasmaz: It was deliberate but I guess you're right, it should save the changes as a whole.

erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/283/commits/40f99d4f9d8709f7de4dfee8872f13189efa6093)

lib/widgets/breez_avatar.dart
erdemyerebasmaz: I believe uploaded image takes precedence over "breez://profile_image".
Also, we can't assume it's file, if it does not start with "breez://profile_image" because there are several other cases.
image

How about
```suggestion
      if (avatarURL.contains("/image_cropper")) {
```

roeierez: I don't think it is robust enough.
We can check by the following sequence:
* "breez://profile_image" prefix.
* "src/icon/vendors"
* check for http url (Uri.parse and make sure the scheme starts with http).
* fallback to file.

lib/widgets/breez_avatar_dialog.dart
```diff
@@ -195,35 +188,25 @@ Widget breezAvatarDialog(BuildContext context, UserProfileBloc userBloc) {
               child: Text('SAVE',
                   style: Theme.of(context).primaryTextTheme.button),
               onPressed: () async {
```
roeierez: The button should be disabled during upload or when there is no profile image chosen.
```suggestion
              onPressed: _pickedImage == null || _isUploading ? null : () async {
```

lib/widgets/breez_avatar_dialog.dart
```diff
@@ -236,12 +219,23 @@ Widget breezAvatarDialog(BuildContext context, UserProfileBloc userBloc) {
   );
 }
 
-// Do not pop dialog if there's a upload in progress
-Future<bool> _onWillPop(UPLOAD_STATE _state) async {
-  if (_state == UPLOAD_STATE.UPLOAD_IN_PROGRESS) {
-    return false;
+Future<bool> _uploadImage(File _pickedImage, UserProfileBloc userBloc) async {
+  if (_pickedImage != null) {
```
roeierez: We don't need this condition, we can verify this in the caller scope.

lib/widgets/breez_avatar_dialog.dart
```diff
@@ -236,12 +219,23 @@ Widget breezAvatarDialog(BuildContext context, UserProfileBloc userBloc) {
   );
 }
 
-// Do not pop dialog if there's a upload in progress
-Future<bool> _onWillPop(UPLOAD_STATE _state) async {
-  if (_state == UPLOAD_STATE.UPLOAD_IN_PROGRESS) {
-    return false;
+Future<bool> _uploadImage(File _pickedImage, UserProfileBloc userBloc) async {
```
roeierez: I think it is better not to return bool but rather just return error from the future in case the upload failed.
It will simplify the function:

lib/widgets/breez_avatar_dialog.dart
```diff
@@ -195,35 +188,25 @@ Widget breezAvatarDialog(BuildContext context, UserProfileBloc userBloc) {
...
-                    }, onError: (error) {
-                      setState(() {
-                        _pickedImage = null;
-                        _state = null;
-                      });
-                      return showFlushbar(
+                setState(() {
+                  _isUploading = true;
+                });
+                await _uploadImage(_pickedImage, userBloc).then((uploaded) {
```
roeierez: We don't need the await here.

lib/widgets/breez_avatar_dialog.dart
erdemyerebasmaz: It should be enabled if there's no profile image chosen because the user could only be changing their name or generated an avatar using "Random" button.

lib/widgets/breez_avatar.dart
erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/283/commits/9d49f38a42ccdaa6615cc70699655d7295bbb639)

lib/widgets/breez_avatar_dialog.dart
erdemyerebasmaz: [Removed unnecessary async/await keyword](https://github.com/breez/breezmobile/pull/283/commits/3614a9bb9cc19102f816de2947ca1c160f674a18)

lib/widgets/breez_avatar_dialog.dart
roeierez: OK, then in the case there is no _pickedImage just don't call to uploadImage.

lib/widgets/breez_avatar_dialog.dart
erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/283/commits/44aefdb0abd3388ff3b6b1b9b7bd6b220941602b)

lib/widgets/breez_avatar.dart
```diff
@@ -54,7 +57,11 @@ class BreezAvatar extends StatelessWidget {
         return _VendorAvatar(radius, avatarURL);
       }
 
-      return _NetworkImageAvatar(avatarURL, radius);
+      if (Uri.parse(avatarURL).scheme.startsWith("http")) {
```
roeierez: This will throw a FormatException if it is not a valid URI (or file path).

erdemyerebasmaz:
```suggestion
      var uri = Uri.tryParse(avatarURL);
      if (uri != null && uri.scheme.startsWith("http")) {
        return _NetworkImageAvatar(avatarURL, radius);
      }
```

roeierez: Change looks good.

PR #289, opened 2020-03-05
POS Items View - Initial Draft

erdemyerebasmaz:
This PR addresses [POS-20](https://breeztech.atlassian.net/browse/POS-20).

 - Switch between keypad and items view with the help of single button
 - Show every item on items view
 - Add a floating action button to create new items
 - Slide left to show delete button
   - Deleting does not reset amount added to total charge for that item, no certain way to know at the moment
 - Each items price should be displayed in the currency it’s created in
 - Clicking an item should convert it's price to current currency and add it to the total charge amount

PR #290, opened 2020-03-06
POS Items View - UI Fixes

erdemyerebasmaz:
This PR follows up on #289 and addresses the feedbacks given in  [POS-48](https://breeztech.atlassian.net/browse/POS-48)
## UI Changes

**Dashboard:**

- Remove the 'Add Note' (till the minimize task)
- Move the switcher button under the Charge button and add captions 'Keypad | Items'. Take the entire row, align to center. Make it behave like two buttons (only click on the inactive view switches view).

**Catalog View:**
- “No items to display. Add items to this view using the '+' button.”
- Remove the current amount (right to the switcher) in catalog view
- Use a regular icon for an item (as a reference)

**Slidable:**
- No need for the ‘Delete’ caption, icon is enough
- Background color of the slide should be the same as a button color. Foreground color should be white.

**Create Item Page:**
- Change the fields to ‘Item Name’/'Item Price' and currency should look like currency field in ‘Redeem Voucher’. 
- The Add Item form should look like  ‘Redeem Voucher’ in terms of style. The padding, fonts look much better in the redeem form. 

## UI Fixes:
- No visible separator in dark mode in keypad-items switcher. 
- Change ‘Create Item’ to ‘Add Item’
- Remove the flushbar when deleting an item.
- Items price should be displayed w/o trailing zeros
- Use only currency symbols in catalog
- Add more padding between the items
- Change 'sats' back to 'SAT'
- Make the switch buttons easier to click
- Change alignment of charge button and viewswitch
#

 - _Add 'Item SKU' optional field to the Add Item form._ Part of another PR

### DIFF
--- a/lib/routes/charge/items/catalog_item.dart
+++ b/lib/routes/charge/items/catalog_item.dart
@@ -25,7 +25,7 @@ class CatalogItem extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     return Padding(
-      padding: const EdgeInsets.all(8.0),
+      padding: const EdgeInsets.fromLTRB(0, 8, 8, 8),
       child: Stack(alignment: Alignment.bottomCenter, children: <Widget>[
         Slidable(
           actionPane: SlidableDrawerActionPane(),
--- a/lib/routes/charge/items/catalog_item.dart
+++ b/lib/routes/charge/items/catalog_item.dart
@@ -36,8 +36,7 @@ class CatalogItem extends StatelessWidget {
             onTap: () {
               DeleteItem deleteItem = DeleteItem(_itemInfo.id);
               posCatalogBloc.actionsSink.add(deleteItem);
-              deleteItem.future.then((_) {},
-                  onError: (err) => showFlushbar(context,
+              deleteItem.future.catchError((err) => showFlushbar(context,
                       message: "Failed to delete ${_itemInfo.name}"));
             },
           ),
--- a/lib/routes/charge/items/catalog_item.dart
+++ b/lib/routes/charge/items/catalog_item.dart
@@ -24,26 +24,27 @@ class CatalogItem extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return Padding(
-      padding: const EdgeInsets.fromLTRB(0, 8, 8, 8),
-      child: Stack(alignment: Alignment.bottomCenter, children: <Widget>[
-        Slidable(
-          actionPane: SlidableDrawerActionPane(),
-          actionExtentRatio: 0.25,
-          secondaryActions: <Widget>[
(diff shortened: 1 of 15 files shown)


lib/routes/charge/items/catalog_item.dart
```diff
@@ -36,9 +36,7 @@ class CatalogItem extends StatelessWidget {
             onTap: () {
               DeleteItem deleteItem = DeleteItem(_itemInfo.id);
               posCatalogBloc.actionsSink.add(deleteItem);
-              deleteItem.future.then(
-                  (_) => showFlushbar(context,
-                      message: "${_itemInfo.name} is successfully deleted"),
+              deleteItem.future.then((_) {},
```
roeierez: Instead can use catchError directly:

`deleteItem.future.catchError( (err){})`

lib/routes/charge/items/create_item.dart
```diff
@@ -160,7 +160,7 @@ class CreateItemPageState extends State<CreateItemPage> {
                                                 return DropdownMenuItem<String>(
                                                   value: value.symbol,
                                                   child: Text(
-                                                    value.displayName,
+                                                    value.symbol.toUpperCase(),
```
roeierez: Maybe instead of `toUpperCase()` we should define the symbol as SAT ?

erdemyerebasmaz: This was causibg issues recognizing older data created with "Sat".
We can do as you say, if we are sure no-one has used this feature yet or we can add code to handle that older data set.

erdemyerebasmaz: This was causing issues recognizing older data created with "Sat".
We can do as you say, if we are sure no-one has used this feature yet or we can add code to handle that older data set.

lib/routes/charge/items/catalog_item.dart
erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/290/commits/f19da1be4e7935fab1fe488c3ff5dcae3d62b38f)

lib/routes/charge/items/create_item.dart
erdemyerebasmaz: Suggestion:
image

roeierez: I think the suggestion is reasonable. 
Maybe even consider:
`c.symbol.topUpperCase() == symbol.toUpperCase()`
To be on the safe side.

erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/290/commits/7d1fed83dfdf843dd890f7aac4483f8601bf5cf5)

PR #294, opened 2020-03-11
Sales view

roeierez:
This PR adds initial implementation of Sale view applied in the POS mode.

### DIFF
--- a/lib/routes/charge/sale_view.dart
+++ b/lib/routes/charge/sale_view.dart
@@ -116,9 +116,10 @@ class _TotalSaleCharge extends StatelessWidget {
     var totalSats = currentSale.totalChargeSat;
     String formattedCharge;
     if (useFiat) {
-      formattedCharge = accountModel.fiatCurrency.format(totalSats);
+      var fiatValue = accountModel.fiatCurrency.satToFiat(totalSats);
+      formattedCharge = accountModel.fiatCurrency.formatFiat(fiatValue, allowBelowMin: true, removeTrailingZeros: true);
     } else {
-      formattedCharge = accountModel.currency.format(totalSats);
+      formattedCharge = accountModel.currency.format(totalSats, removeTrailingZeros: true);
     }
     return Text(
       "Total: $formattedCharge",
@@ -146,6 +147,10 @@ class SaleLinesList extends StatelessWidget {
           return Column(
             children: [
               SaleLineWidget(
+                  onDelete: (){
+                    var newSale = currentSale.copyWith(saleLines: currentSale.saleLines..removeAt(index));
+                    posCatalogBloc.actionsSink.add(SetCurrentSale(newSale));
+                  },
                   onChangeQuantity: (int delta) {
                     var saleLine = currentSale.saleLines[index];
                     var newSale = currentSale.incrementQuantity(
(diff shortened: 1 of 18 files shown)


lib/routes/dev/dev.dart
```diff
@@ -418,8 +419,15 @@ class DevViewState extends State<DevView> {
...
         */
-
+    choices.add(Choice(
+        title: 'Reset POS DB',
+        icon: Icons.phone_android,
+        function: () {
+          PosCatalogBloc bloc = AppBlocsProvider.of<PosCatalogBloc>(context);
+          bloc.resetDB();
+        }));
     return choices;
```
erdemyerebasmaz: I know most items on 'Developers' screen is not for users or not ready for their use yet but I think this option does not belong here. Since 'Developers' is not accesible from POS mode. I believe "Reset POS DB" should be moved to POS Settings. It should have several warning/confirmation dialogs. Enter Manager Password, PIN to confirm etc. 
At the same time, I believe we should wait till we can import items from any source first before making this visible to users..

lib/routes/charge/sale_view.dart
```diff
@@ -0,0 +1,212 @@
...
+                      currentSale: currentSale,
+                      useFiat: widget.useFiat,
+                    ),
+                    actions: <Widget>[
+                      IconButton(
+                        icon: Icon(
+                          Icons.note_add,
+                          color: Theme.of(context).iconTheme.color,
+                        ),
+                        onPressed: () {},
```
erdemyerebasmaz: Comment out note button till Note PR, unless this was added for testing purposes.

lib/widgets/badge.dart
```diff
@@ -0,0 +1,27 @@
+import 'package:badges/badges.dart';
+import 'package:flutter/material.dart';
+
+class BadgeIcon extends StatelessWidget {
```
erdemyerebasmaz: I don't see this being used anywhere.
Will delete item, add/remove quantity icons be replaced by this widget?

lib/routes/charge/sale_view.dart
```diff
@@ -0,0 +1,212 @@
...
+  @override
+  Widget build(BuildContext context) {
+    var totalSats = currentSale.totalChargeSat;
+    String formattedCharge;
+    if (useFiat) {
+      formattedCharge = accountModel.fiatCurrency.format(totalSats);
+    } else {
+      formattedCharge = accountModel.currency.format(totalSats);
+    }
+    return Text(
```
erdemyerebasmaz: Return empty string for 0 total charge to prevent unwanted behaviour* on deleting last item.

*: Displaying "< €0.00" right before popping the widget.

lib/routes/dev/dev.dart
roeierez: I only did this to be used internally because I changed the db scheme. It is not meant for uses and I will remove it once the POS is ready.

lib/routes/charge/sale_view.dart
roeierez: Let's leave it so I can get feedback on the UI.

lib/widgets/badge.dart
roeierez: It is being used in pos_invoice.dart.

lib/routes/charge/sale_view.dart
roeierez: Good feedback. I changed to format correctly.

erdemyerebasmaz:
LGTM

PR #296, opened 2020-03-12
Add 'Item SKU' optional field to the Add Item form and create Edit Item page

erdemyerebasmaz:
This PR addresses [POS-47](https://breeztech.atlassian.net/browse/POS-47).

- Item SKU optional field is added to Add Item form.
- "Create Item" is renamed to "Add Item" across the app
- Added Edit Item option to Slidable


lib/routes/charge/items/add_item_page.dart
```diff
@@ -263,6 +275,7 @@ class CreateItemPageState extends State<CreateItemPage> {
                   name: _nameController.text.trimRight(),
                   currency: _currencySymbol(),
                   price: double.parse(_formattedPrice()),
+                  sku: _skuController.text,
```
roeierez: Shouldn't it be null in case empty?

lib/routes/charge/items/edit_item_page.dart
```diff
@@ -0,0 +1,316 @@
...
+import 'package:breez/bloc/pos_catalog/model.dart';
+import 'package:breez/bloc/user_profile/currency.dart';
+import 'package:breez/theme_data.dart' as theme;
+import 'package:breez/widgets/back_button.dart' as backBtn;
+import 'package:breez/widgets/flushbar.dart';
+import 'package:breez/widgets/single_button_bottom_bar.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+
+class EditItemPage extends StatefulWidget {
```
roeierez: What are the differences between AddItemPage and EditItemPage ?
It seems that by passing a model to the the AddItemPage you can differentiate the case and support edit with the same codebase.
Thoughts?

lib/routes/charge/items/add_item_page.dart
erdemyerebasmaz:
```suggestion
               sku:  _skuController.text.isNotEmpty ? _skuController.text : null,
```
What are the differences between
`sku: null `versus `sku: ""` ?

lib/routes/charge/items/edit_item_page.dart
erdemyerebasmaz: I agree, we should reduce it to one page. If an item + model is passed to the page, it should be in edit mode(initialize the fields and use UpdateItem action).

erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/296/commits/593e9e2cdf47bad76806d4d97fd740ca12b1caff )
[bugfix](https://github.com/breez/breezmobile/pull/296/commits/dd8bf2e210865627140b00cbfbe93ec07dfa1219)

erdemyerebasmaz: @roeierez Replaced _selectedCurrency & _selectedFiatCurrency with [CurrencyWrapper](https://github.com/breez/breezmobile/pull/296/commits/01547fd5640ce7fb6c51732c6560b38bb11e4277)

lib/routes/charge/items/item_page.dart
```diff
@@ -14,29 +14,53 @@ import 'package:breez/widgets/single_button_bottom_bar.dart';
...
-class CreateItemPage extends StatefulWidget {
+import '../currency_wrapper.dart';
+
+class ItemPage extends StatefulWidget {
+  final Item item;
+  final AccountModel accountModel;
   final PosCatalogBloc _posCatalogBloc;
 
-  CreateItemPage(this._posCatalogBloc);
+  ItemPage(this._posCatalogBloc, {this.item, this.accountModel});
```
roeierez: I see we need the accountModel in both new/edit flows. If that is so, it shouldn't be optional.

lib/routes/charge/items/item_page.dart
```diff
@@ -14,29 +14,53 @@ import 'package:breez/widgets/single_button_bottom_bar.dart';
...
   TextEditingController _nameController = TextEditingController();
   TextEditingController _priceController = TextEditingController();
+  TextEditingController _skuController = TextEditingController();
   AccountBloc _accountBloc;
   bool _isInit = false;
-  bool _isFiat = false;
-  Currency _selectedCurrency = Currency.BTC;
-  FiatConversion _selectedFiatCurrency;
-  String currency;
+  bool _inEditMode = false;
```
roeierez: can be calculated property.

lib/routes/charge/items/item_page.dart
```diff
@@ -14,29 +14,53 @@ import 'package:breez/widgets/single_button_bottom_bar.dart';
...
+  String _titlePrefix = "Add";
+  CurrencyWrapper _selectedCurrency = CurrencyWrapper.fromBTC(Currency.BTC);
+
+  @override
+  void initState() {
+    super.initState();
+    _inEditMode = widget.item != null;
+    if (_inEditMode) _initializeFields();
```
roeierez: Maybe initialize the fields anyway? In the "new" mode we need to change the title anyway (I think it is better to have  a variable "title" instead of "titlePrefix").
We can embed here the initialization without needing another function.

lib/routes/charge/items/item_page.dart
```diff
@@ -93,30 +119,25 @@ class CreateItemPageState extends State<CreateItemPage> {
                           child: TextFormField(
                               keyboardType: TextInputType.numberWithOptions(
                                   decimal: true),
-                              inputFormatters: _isFiat
+                              inputFormatters: _selectedCurrency.chargeSuffix ==
+                                      "BTC"
```
roeierez: Can we add the regex as property on currency? Call it "inputFormatter"?

lib/routes/charge/items/item_page.dart
```diff
@@ -246,29 +261,48 @@ class CreateItemPageState extends State<CreateItemPage> {
...
-                  price: double.parse(_formattedPrice()),
-                ),
-              );
-              widget._posCatalogBloc.actionsSink.add(addItem);
-              addItem.future.then((_) {
-                Navigator.pop(context);
-              });
+              if (_inEditMode) {
+                UpdateItem updateItem = UpdateItem(
+                  Item(
```
roeierez: I think it is cleaner to merge changes with the current edited item.
Something like:
`widget.item.copyWith(name: ..., price: ..., )`

lib/routes/charge/items/item_page.dart
```diff
@@ -246,29 +261,48 @@ class CreateItemPageState extends State<CreateItemPage> {
...
+                    id: widget.item.id,
+                    name: _nameController.text.trimRight(),
+                    currency: _selectedCurrency.shortName,
+                    price: double.parse(
+                        _formattedPrice(double.parse(_priceController.text))),
+                    sku: _skuController.text,
```
roeierez: Should be  null if empty.
Same bellow.

lib/routes/charge/items/item_page.dart
```diff
@@ -246,29 +261,48 @@ class CreateItemPageState extends State<CreateItemPage> {
...
-                Navigator.pop(context);
-              });
+              if (_inEditMode) {
+                UpdateItem updateItem = UpdateItem(
+                  Item(
+                    id: widget.item.id,
+                    name: _nameController.text.trimRight(),
+                    currency: _selectedCurrency.shortName,
+                    price: double.parse(
+                        _formattedPrice(double.parse(_priceController.text))),
```
roeierez: why the roundtrip? Can we just do `double.parse` ?
same bellow?

lib/routes/charge/items/item_page.dart
erdemyerebasmaz: True, I kept this to make the code readable. I'll add a comment explaining this instead.

lib/routes/charge/items/add_item_page.dart
roeierez: The difference is "no value" vs "empty string value".
At the repository level  (SQL for example) it is important to have those as null values.

erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/296/commits/3c5761d4edc0927fba540c2fb49ba332fdd6ade2)

lib/routes/charge/items/item_page.dart
erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/296/commits/de801c8a9c28fc66abb0abcce56d96aca4d6349f)

lib/routes/charge/items/item_page.dart
erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/296/commits/ad6d6e0ee983b18b1315650feaa6802af10babac)

lib/routes/charge/items/item_page.dart
erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/296/commits/fc5073de055417289cde5536d54b7c0e7c1b4b67)

lib/routes/charge/items/item_page.dart
erdemyerebasmaz: AccountModel is removed and it's being listened once on didChangeDependencies.
[680c91](https://github.com/breez/breezmobile/pull/296/commits/680c91ca295661f5e9c2ba5a5572a18568a3c193)

lib/routes/charge/items/item_page.dart
erdemyerebasmaz: Title changes:
3d59c89
5a7e6f5

lib/routes/charge/currency_wrapper.dart
```diff
@@ -48,6 +49,18 @@ class CurrencyWrapper {
     return fiat.exchangeRate / 100000000;
   }
 
+  List<TextInputFormatter> get inputFormatters {
```
roeierez: We should return here only the Regex. The TextInputFormatter is a UI element.
Also better to add it on Currency and FiatConversion and call them from here.

lib/routes/charge/items/item_page.dart
```diff
@@ -0,0 +1,275 @@
...
+            _priceController.text = _formattedPrice(widget.item.price);
+          });
+        });
+      }
+
+      fetchRatesAction.future.catchError((err) {
+        if (this.mounted) {
+          setState(() {
+            showFlushbar(context,
+                message: "Failed to retrieve BTC exchange rate.");
```
roeierez: Are we OK to continue with this form in a state without the conversion rates.

lib/routes/charge/items/item_page.dart
```diff
@@ -0,0 +1,275 @@
...
+                              inputFormatters:
+                                  _selectedCurrency.inputFormatters,
+                              controller: _priceController,
+                              decoration: InputDecoration(
+                                  labelText: "Item Price",
+                                  hintText: "Enter an item price",
+                                  border: UnderlineInputBorder()),
+                              style: theme.FieldTextStyle.textStyle,
+                              validator: (value) {
+                                if (value.length == 0) {
```
roeierez: Do we need to validate number here for the case the user paste? or not?

lib/routes/charge/currency_wrapper.dart
erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/296/commits/890d18227d192ef7b4d060055869bc53b48cca1e)

lib/routes/charge/items/item_page.dart
erdemyerebasmaz: I allowed the users to continue without a fiat currency list so that they can at least create items with BTC/SAT.

PR #302, opened 2020-03-15
Consider dynamic fee changes on sale charge.

roeierez:
This PR fixes the wrong offset sometimes seen in the sale total charge. There were two issues that have been caused this weird behavior:
1. The conversion to sats and backwards wasn't accurate enough and lost precision.
2. The rates were changing and the sales view was not updated accordingly.

### DIFF
--- a/lib/bloc/pos_catalog/bloc.dart
+++ b/lib/bloc/pos_catalog/bloc.dart
@@ -1,9 +1,11 @@
 import 'dart:async';
 
+import 'package:breez/bloc/account/account_model.dart';
 import 'package:breez/bloc/async_actions_handler.dart';
 import 'package:breez/bloc/pos_catalog/actions.dart';
 import 'package:breez/bloc/pos_catalog/repository.dart';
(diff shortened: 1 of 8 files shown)


lib/bloc/pos_catalog/bloc.dart
```diff
@@ -32,6 +34,31 @@ class PosCatalogBloc with AsyncActionsHandler {
...
+
+      // In case the price is locked we don't calculate the charge.
+      if (currentSale.priceLocked) {
+        return;
+      }
+      _currentSaleController.add(currentSale.copyWith(
+        saleLines: currentSale.saleLines.map(
+          (sl){
+            double rate = sl.satConversionRate;
+            Currency curr = Currency.fromTickerSymbol(sl.currency);
```
erdemyerebasmaz: CurrencyWrapper.fromShortName(sl.currency, acc) can be used here.

roeierez: Right now we can't since currently the CurrencyWrapper is only at the UI level. Once we move it into the bloc we can use it here.

PR #308, opened 2020-03-18
Formatting fiat with blank group separator for POS

erdemyerebasmaz:
This PR addresses [POS-57](https://breeztech.atlassian.net/browse/POS-57)

The amounts are now displayed in 3-digit grouping with a blank " " group separator for POS views.

12345.67 -> 12 345.67

The affected fields are current amount, charge button, price field on items list and their counterparts on current sale view.


lib/bloc/account/fiat_conversion.dart
```diff
@@ -53,11 +58,35 @@ class FiatConversion {
     // if conversion result is less than the minimum it doesn't make sense to display
     // it.
     if (!allowBelowMin && fiatAmount < minimumAmount) {
-      formattedAmount += minimumAmount.toStringAsFixed(fractionSize);
+      formattedAmount = minimumAmount.toStringAsFixed(fractionSize);
       prefix = '< ' + prefix;
     } else {
-      // Otherwise just show the formatted value.
-      formattedAmount += fiatAmount.toStringAsFixed(fractionSize);
+      if (useBlankGroupSeparator) {
```
roeierez: What is the reason we need a parameter? Shouldn't we be consistent and add the blank separator on all formatting except for user input (like sats) ?

erdemyerebasmaz: There were elements not mentioned in the PR, such as exchange rate label on currency converter label. I assumed these changes only applied to POS fields.
Removed this parameter so the formatting is consistent throughout the app.

lib/bloc/account/fiat_conversion.dart
```diff
@@ -53,11 +59,12 @@ class FiatConversion {
...
     // it.
     if (!allowBelowMin && fiatAmount < minimumAmount) {
-      formattedAmount += minimumAmount.toStringAsFixed(fractionSize);
+      formattedAmount = minimumAmount.toStringAsFixed(fractionSize);
       prefix = '< ' + prefix;
     } else {
-      // Otherwise just show the formatted value.
-      formattedAmount += fiatAmount.toStringAsFixed(fractionSize);
+      final formatter = CurrencyFormatter.formatter;
+      formatter.minimumFractionDigits = fractionSize;
```
roeierez: This will change the global formatter that affects Sats.

roeierez: My mistake, I see they are two different instances.

PR #309, opened 2020-03-19
POS currency persistence

erdemyerebasmaz:
This PR addresses [POS-45](https://breeztech.atlassian.net/browse/POS-45).

Selected currency now persist :
- between user and pos mode,
- and after restarts*

*: Unless the app fails to retrieve exchange rates, which in that case the UI renders with BTC. 

If the user does not change to another currency at this point, the stored value for selected currency does not change. Meaning, restarting the app again and getting exchange rates without an error would display the selected currency.

### DIFF
--- a/lib/bloc/account/account_bloc.dart
+++ b/lib/bloc/account/account_bloc.dart
@@ -401,12 +401,10 @@ class AccountBloc {
       _currentUser = user;
       log.info("account: got new user $user");
       //convert currency.
-      _accountController
-          .add(_accountController.value.copyWith(currency: user.currency));
-      _accountController.add(
-          _accountController.value.copyWith(fiatShortName: user.fiatCurrency));
-      _accountController.add(
-          _accountController.value.copyWith(posCurrencyShortName: user.posCurrencyShortName));
+      _accountController.add(_accountController.value.copyWith(
+          currency: user.currency,
+          fiatShortName: user.fiatCurrency,
+          posCurrencyShortName: user.posCurrencyShortName));
       var updatedPayments = _paymentsController.value.copyWith(
         nonFilteredItems: _paymentsController.value.nonFilteredItems
             .map((p) => p.copyWith(_accountController.value))
--- a/lib/bloc/account/account_bloc.dart
+++ b/lib/bloc/account/account_bloc.dart
@@ -62,9 +62,6 @@ class AccountBloc {
   Sink<AccountSettings> get accountSettingsSink =>
       _accountSettingsController.sink;
 
-  final _posCurrencyController = BehaviorSubject<String>();
-  Sink<String> get posCurrencySink => _posCurrencyController.sink;
-
   final _routingNodeConnectionController = BehaviorSubject<bool>();
   Stream<bool> get routingNodeConnectionStream =>
       _routingNodeConnectionController.stream;
@@ -160,7 +157,6 @@ class AccountBloc {
       //listen streams
       _listenAccountActions();
       _hanleAccountSettings();
-      _listenPosCurrencyChanges();
       _listenUserChanges(userProfileStream);
       _listenFilterChanges();
       _listenAccountChanges();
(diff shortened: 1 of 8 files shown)


lib/bloc/account/account_model.dart
```diff
@@ -218,6 +224,8 @@ class AccountModel {
       (f) => f.currencyData.shortName == _fiatShortName,
       orElse: () => null);
   List<FiatConversion> get fiatConversionList => _fiatConversionList;
+  String get posCurrencyShortName => _posCurrencyShortName;
+  CurrencyWrapper get posCurrency => CurrencyWrapper.fromShortName(_posCurrencyShortName, this);
```
roeierez: The CurrencyWrapper is still a UI element, we can't use it in the bloc.
Later when the pos stabilizes I think we should create a unified interface for btc currencies and fiat currencies.

lib/bloc/user_profile/user_profile_bloc.dart
```diff
@@ -329,6 +333,14 @@ class UserProfileBloc {
     });
   }
 
+  _listenPosCurrencyChanges(ServiceInjector injector) {
```
roeierez: Instead of introducing another sink, isn't it better to use an AsyncAction?

lib/routes/charge/pos_invoice.dart
```diff
@@ -72,11 +73,22 @@ class POSInvoiceState extends State<POSInvoice> with TickerProviderStateMixin {
...
+        currentCurrency = acc.posCurrency;
+      });
+      FetchRates fetchRatesAction = FetchRates();
+      accountBloc.userActionsSink.add(fetchRatesAction);
+      fetchRatesAction.future.catchError((err) {
+        if (this.mounted) {
+          UserProfileBloc userProfileBloc =
+              AppBlocsProvider.of<UserProfileBloc>(context);
+          setState(() {
+            userProfileBloc.posCurrencySink.add("BTC");
```
roeierez: Why do we need this? We don't need to manipulate the user choice. If we failed to retrieve the rates, let's just render the default (BTC? Sats?).

lib/routes/charge/pos_invoice.dart
```diff
@@ -843,21 +859,19 @@ class POSInvoiceState extends State<POSInvoice> with TickerProviderStateMixin {
...
-      if (flipFiat) {
-        _useFiat = !_useFiat;
-      }
+      CurrencyWrapper currency =
+          CurrencyWrapper.fromShortName(value, accountModel);
       _clearAmounts(currentSale);
 
-      if (currency != null) {
-        userProfileBloc.currencySink.add(currency);
+      if (currency.btc != null) {
```
roeierez: btc property is exposed by mistake, we shouldn't expose it. Use "isFiat" instead.

lib/bloc/account/account_bloc.dart
```diff
@@ -405,6 +405,8 @@ class AccountBloc {
           .add(_accountController.value.copyWith(currency: user.currency));
       _accountController.add(
           _accountController.value.copyWith(fiatShortName: user.fiatCurrency));
+      _accountController.add(
```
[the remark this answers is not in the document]
erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/309/commits/a3cd0f93859005687c142dabbde340afc6a8b879)

lib/bloc/account/account_model.dart
erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/309/commits/c421d202e4f3bf9da390a0f2adea75cd720eadfa)
I agree.

lib/routes/charge/pos_invoice.dart
erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/309/commits/2634aa19b1891d79709f56eaff5292aeea591d8f)

lib/routes/charge/pos_invoice.dart
erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/309/commits/e7b2dfd7680ea9b850ed03f26e753a42e3a37c28)

lib/bloc/user_profile/user_profile_bloc.dart
erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/309/commits/a01c04f3e980c13d1ddf5ae97b9f99fd76f70c13)

PR #310, opened 2020-03-19
Sale currency

roeierez:
This PR adds the price conversion to the current sale currency for every sale line.
In case the Sale is read-only (executed sale) we show the conversion to Sats.


lib/routes/charge/sale_view.dart
```diff
@@ -379,9 +396,7 @@ class SaleLineWidget extends StatelessWidget {
             //style: TextStyle(fontWeight: FontWeight.bold),
           ),
           subtitle: Text(
-              currrency.symbol +
```
erdemyerebasmaz: Typo

PR #318, opened 2020-03-24
Item avatar picker

erdemyerebasmaz:
This PR addresses [POS-42](https://breeztech.atlassian.net/browse/POS-42)

- Users can now select an image for their items from a range of [Material icons](https://material.io/resources/icons/?style=baseline).
- The icon changes are reflected instantly, there are no 'save changes' prompts during selection.
- If no image is selected, the avatar defaults to showing first available two letters of the item name with an underline.
image
- There's a search bar to find the desired icon easily.
image
- The icon changes can be reverted by pressing the trashcan badge on icon selection screen.
image
- The icons have no decoration on catalog.
- Icons are differentiated by 'icon:' prefix.


lib/routes/charge/items/item_avatar.dart
```diff
@@ -68,6 +75,46 @@ Color fromHex(String hexString) {
...
+  final double radius;
+  final String iconName;
+  final bool useDecoration;
+
+  _IconAvatar(this.radius, this.iconName, this.useDecoration);
+
+  @override
+  Widget build(BuildContext context) {
+    Map<String, IconData> iconMapping = IconMap().iconMap;
+    String iconName = this.iconName.substring(5);
```
roeierez: It is better to hide our internal protocol at this level and pass the icon name to this class.
The caller is aware of the protocol and should strip the prefix.

lib/routes/charge/items/item_avatar.dart
```diff
@@ -68,6 +75,46 @@ Color fromHex(String hexString) {
...
+                  fit: BoxFit.cover))
+          : null,
+      child: Column(
+        mainAxisSize: MainAxisSize.max,
+        mainAxisAlignment: MainAxisAlignment.center,
+        crossAxisAlignment: CrossAxisAlignment.center,
+        children: <Widget>[
+          Icon(
+            iconMapping[iconName],
+            size: useDecoration ? radius : radius * 1.5,
```
roeierez: Why do we duplicate the radius at 1.5 ?

lib/routes/charge/items/item_avatar.dart
```diff
@@ -91,17 +138,67 @@ class _FileImageAvatar extends StatelessWidget {
...
+              child: Text(
+              _getFirstTwoLetters(),
+              style: TextStyle(
+                  fontSize: useDecoration ? 48 : radius,
+                  color: Color.fromRGBO(255, 255, 255, 0.88),
+                  decoration: TextDecoration.underline,
+                  letterSpacing: 0.0,
+                  fontFamily: "IBMPlexSans"),
+            ))
+          : Column(
```
roeierez: It doesn't look like the "Select Image" case belongs here. Instead at a higher level we should decide if to show an avatar (in case we have one) or to render a "Select Image" widget.

lib/routes/charge/items/item_avatar_picker.dart
```diff
@@ -0,0 +1,167 @@
...
+  ItemAvatarPicker(this.itemImage, this.onImageSelected, {this.itemName});
+
+  @override
+  State<StatefulWidget> createState() {
+    return ItemAvatarPickerState();
+  }
+}
+
+class ItemAvatarPickerState extends State<ItemAvatarPicker> {
+  Map<String, IconData> iconMap = IconMap().iconMap;
```
roeierez: Can we do this static? so it won't be loaded every time?

lib/routes/charge/items/item_avatar_picker.dart
```diff
@@ -0,0 +1,167 @@
...
+          ],
+        ),
+      ),
+    );
+  }
+
+  Center _buildItemAvatar(BuildContext context) {
+    return Center(
+        child: Badge(
+      showBadge: _selectedImage != "",
```
roeierez: alignment issue?

lib/routes/charge/items/item_page.dart
```diff
@@ -262,6 +296,9 @@ class ItemPageState extends State<ItemPage> {
               } else {
                 AddItem addItem = AddItem(
                   Item(
+                      imageURL: _itemImage != null && _itemImage.isNotEmpty
```
roeierez: why not just _itemImage as in the update case above?

erdemyerebasmaz:
Feedback will be handled in another PR

lib/routes/charge/items/item_avatar.dart
erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/319/commits/022093abc1a629a528afb157ec290d18e185ad4b)

lib/routes/charge/items/item_avatar.dart
erdemyerebasmaz: We want the icon to look bigger on catalog list without making the container larger.

lib/routes/charge/items/item_avatar.dart
erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/319/commits/cefa0a16444237d5af2f2afbf11291f438b2353b)

lib/routes/charge/items/item_avatar_picker.dart
erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/319/commits/e7f34f15e5c04d82fe1331ebe3d39564a487dd68)

lib/routes/charge/items/item_avatar_picker.dart
erdemyerebasmaz: [Fixed](https://github.com/breez/breezmobile/pull/319/commits/ac1487f7c6e7f9575d1e3edfd4d532d217214be9)

lib/routes/charge/items/item_page.dart
erdemyerebasmaz: Update case was faulty and saved null value as empty string.
[Fixed](https://github.com/breez/breezmobile/pull/319/commits/257a6f0237bc1914343dc19e3cd22418c1af1367)

PR #328, opened 2020-04-19
Node ID(pub key) support to QR Scan

erdemyerebasmaz:
This PR addresses [BU-222](https://breeztech.atlassian.net/browse/BU-222)


lib/routes/home/floating_actions_bar.dart
```diff
@@ -142,14 +143,21 @@ class FloatingActionsBar extends StatelessWidget {
                                 ));
                                 return;
                               }
+                              if (!isValidNodeId(scannedString)) {
```
roeierez, line 146: Shouldn't it be the opposite condition?
```suggestion
                              if (isValidNodeId(scannedString)) {
```

PR #332, opened 2020-04-27
Add a menu entry in the Developers menu to prune nodes not supporting TLV

yaslama:


lib/routes/dev/dev.dart
```diff
@@ -482,6 +486,16 @@ class DevViewState extends State<DevView> {
     });
   }
 
+  void _refreshGraph() async {
+    Navigator.push(
+            context,
+            createLoaderRoute(context,
+                message: "Pruning nodes without TLV support...", opacity: 0.8));
+    widget._breezBridge.deleteNonTLVNodesFromGraph().whenComplete(() {
+      _promptForRestart();
```
roeierez, line 495: Need to pop the loader route here as well.

PR #338, opened 2020-06-10
Add fiat value instead of balance

erdemyerebasmaz:
This PR addresses [BU-228](https://breeztech.atlassian.net/browse/BU-228).

- Balance title is removed from wallet dashboard.
- Fiat equivalent of balance is shown under balance and users can switch between fiat currencies by clicking on this field.
- This field is only shown if the amount is bigger than minimal value, e.g. "< $0.01"
- Clicking on the fiat value will look up the next fiat currency that's above minimal value.

### DIFF
--- a/lib/routes/home/wallet_dashboard.dart
+++ b/lib/routes/home/wallet_dashboard.dart
@@ -69,37 +69,6 @@ class WalletDashboardState extends State<WalletDashboard> {
                   child: StatusIndicator(context, widget._accountModel))
               : SizedBox(),
           Positioned(
-              top: 70 - BALANCE_OFFSET_TRANSITION * widget._offsetFactor,
-              child: Center(
-                child: widget._accountModel != null &&
-                        !widget._accountModel.initial &&
-                        (widget._accountModel.fiatConversionList.isNotEmpty ||
-                            widget._accountModel.fiatCurrency != null) &&
-                        isAboveMinAmount(widget._accountModel.fiatCurrency)
-                    ? FlatButton(
-                        onPressed: () {
-                          var nextFiatCurrencyIndex = widget
-                                  ._accountModel.fiatConversionList
-                                  .indexOf(widget._accountModel.fiatCurrency) +
-                              1;
-                          findNextValidFiatConversion(nextFiatCurrencyIndex %
-                              widget._accountModel.fiatConversionList.length);
-                        },
-                        child: Text(
-                            "${widget._accountModel.formattedFiatBalance}",
-                            style: Theme.of(context)
-                                .textTheme
-                                .subtitle2
-                                .copyWith(
-                                    color: Theme.of(context)
-                                        .textTheme
-                                        .subtitle2
-                                        .color
-                                        .withOpacity(pow(
-                                            0.95 - widget._offsetFactor, 8)))))
-                    : SizedBox(),
-              )),
-          Positioned(
             top: 30 - BALANCE_OFFSET_TRANSITION * widget._offsetFactor,
             child: Center(
               child: widget._accountModel != null &&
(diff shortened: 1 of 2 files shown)


lib/routes/home/wallet_dashboard.dart
```diff
@@ -114,7 +102,38 @@ class WalletDashboardState extends State<WalletDashboard> {
...
+                        (widget._accountModel.fiatConversionList.isNotEmpty ||
+                            widget._accountModel.fiatCurrency != null) &&
+                        isAboveMinAmount(widget._accountModel.fiatCurrency)
+                    ? FlatButton(
+                        onPressed: () {
+                          var nextFiatCurrencyIndex = widget
+                                  ._accountModel.fiatConversionList
+                                  .indexOf(widget._accountModel.fiatCurrency) +
+                              1;
+                          findNextValidFiatConversion(nextFiatCurrencyIndex %
```
roeierez, line 120: The name of `findNextValidFiatConversion` implies read-only. Better to keep it read-only and call the `_onFiatCurrencyChange` callback here.

lib/routes/home/wallet_dashboard.dart
```diff
@@ -114,7 +102,38 @@ class WalletDashboardState extends State<WalletDashboard> {
...
-          )
+          ),
+          Positioned(
+              top: 70 - BALANCE_OFFSET_TRANSITION * widget._offsetFactor,
+              child: Center(
+                child: widget._accountModel != null &&
+                        !widget._accountModel.initial &&
+                        (widget._accountModel.fiatConversionList.isNotEmpty ||
+                            widget._accountModel.fiatCurrency != null) &&
+                        isAboveMinAmount(widget._accountModel.fiatCurrency)
```
roeierez, line 113:
```suggestion
                        isAboveMinAmount(widget._accountModel?.fiatCurrency)
```

lib/routes/home/wallet_dashboard.dart
```diff
@@ -114,7 +102,38 @@ class WalletDashboardState extends State<WalletDashboard> {
...
+                    ? FlatButton(
+                        onPressed: () {
+                          var nextFiatCurrencyIndex = widget
+                                  ._accountModel.fiatConversionList
+                                  .indexOf(widget._accountModel.fiatCurrency) +
+                              1;
```
roeierez, line 119: findNextValidFiatConversion is a method of this class and has access to the widget._accountModel. I think it is better to move all the logic there and keep this UI section simpler.

lib/routes/home/wallet_dashboard.dart
```diff
@@ -130,4 +149,23 @@ class WalletDashboardState extends State<WalletDashboard> {
       },
     );
   }
+
+  void findNextValidFiatConversion(int nextIndex, {int retries = 0}) {
```
roeierez, line 153: Maybe a for-loop will read simpler than a recursive function here?

erdemyerebasmaz: https://github.com/breez/breezmobile/pull/338/commits/eb9e3c34ea346715bf0efa1572f535f5428933ec

lib/routes/home/wallet_dashboard.dart
erdemyerebasmaz: This logic is moved to findNextValidFiatConversion(nextValidFiatConversion) method https://github.com/breez/breezmobile/pull/338/commits/eb9e3c34ea346715bf0efa1572f535f5428933ec

lib/routes/home/wallet_dashboard.dart
erdemyerebasmaz: Done https://github.com/breez/breezmobile/pull/338/commits/334e7b2861d9a43a51c30ad86a730685a70249f0

PR #340, opened 2020-06-17
Support lightning: protocol via NFC

erdemyerebasmaz:
This PR addresses [BU-232](https://breeztech.atlassian.net/browse/BU-232)

Registered Breez app to lightning deep links read through NFC.


android/app/src/main/java/com/breez/client/NfcHandler.java
```diff
@@ -159,6 +159,23 @@ public void handleIntent(Intent intent) {
                             Log.e(TAG, "UnsupportedEncodingException while reading Breez card", exc);
                         }
                     }
+                } else {
+                    try {
```
roeierez, line 163: Instead of using native code to get the nfc data it is better to use the plugin for that. This will have the cross platform advantage.

erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/340/commits/00c56612f1ca9b46c1691fd9d0735b87bfd5eabe)

PR #341, opened 2020-06-19
Create loader on receiving lnlink via NFC

erdemyerebasmaz:
Show a loader while lnlink is being handled, much like receiving an invoice.

### DIFF
--- a/lib/bloc/lnurl/lnurl_bloc.dart
+++ b/lib/bloc/lnurl/lnurl_bloc.dart
@@ -56,7 +56,7 @@ class LNUrlBloc with AsyncActionsHandler {
         }
       }).handleError((error) {
         _lnUrlStreamController.add(fetchLNUrlState.completed);
-        //_lnUrlStreamController.add(error.toString());
+        _lnUrlStreamController.addError(error.toString());
       }).listen((_) {});
     }
     return _lnUrlStreamController.stream;
--- a/lib/bloc/lnurl/lnurl_bloc.dart
+++ b/lib/bloc/lnurl/lnurl_bloc.dart
@@ -15,13 +15,12 @@ enum fetchLNUrlState { started, completed }
 class LNUrlBloc with AsyncActionsHandler {
   BreezBridge _breezLib;
 
-  final _lnUrlStreamController = StreamController.broadcast();
+  StreamController _lnUrlStreamController;
 
   LNUrlBloc() {
     ServiceInjector injector = ServiceInjector();
     _breezLib = injector.breezBridge;
 
-    listenLNUrl();
     registerAsyncHandlers({
       Fetch: _fetch,
       Withdraw: _withdraw,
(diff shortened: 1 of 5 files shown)


lib/handlers/lnurl_handler.dart
```diff
@@ -12,72 +13,104 @@ import '../routes/create_invoice/create_invoice_page.dart';
...
-    Navigator.of(context).push(FadeInRoute(
-      builder: (_) => CreateInvoicePage(lnurlWithdraw: response),
-    ));
-  } else {
-    throw "Unsupported LNUrl";
+      Navigator.of(context).push(FadeInRoute(
+        builder: (_) => CreateInvoicePage(lnurlWithdraw: response),
+      ));
+    } else {
+      _setLoading(false);
```
erdemyerebasmaz, line 48: .

lib/handlers/lnurl_handler.dart
```diff
@@ -12,72 +13,104 @@ import '../routes/create_invoice/create_invoice_page.dart';
...
-  if (response.runtimeType == ChannelFetchResponse) {
-    _openLNURLChannel(context, lnurlBloc, response);
-  } else if (response.runtimeType == WithdrawFetchResponse) {
-    Navigator.popUntil(context, (route) {
+  void executeLNURLResponse(
+      BuildContext context, LNUrlBloc lnurlBloc, dynamic response) {
+    if (response.runtimeType == ChannelFetchResponse) {
+      _openLNURLChannel(context, lnurlBloc, response);
+    } else if (response.runtimeType == WithdrawFetchResponse) {
+      _setLoading(false);
```
erdemyerebasmaz, line 39: .

lib/handlers/lnurl_handler.dart
```diff
@@ -12,72 +13,104 @@ import '../routes/create_invoice/create_invoice_page.dart';
...
+      ));
+    } else {
+      _setLoading(false);
+      throw "Unsupported LNUrl";
+    }
+  }
+
+  void _openLNURLChannel(BuildContext context, LNUrlBloc lnurlBloc,
+      ChannelFetchResponse response) {
+    _setLoading(false);
```
erdemyerebasmaz, line 55: .

roeierez:
@erdemyerebasmaz Why do we need a loader here? What is the reason for the delay?

lib/bloc/lnurl/lnurl_bloc.dart
```diff
@@ -29,23 +27,32 @@ class LNUrlBloc with AsyncActionsHandler {
...
+      if (response.hasWithdraw()) {
+        lnUrlStreamController.add(WithdrawFetchResponse(response.withdraw));
+      }
+      if (response.hasChannel()) {
+        lnUrlStreamController.add(ChannelFetchResponse(response.channel));
+      }
 
-        return Future.error("Unsupported lnurl");
-      });
+      lnUrlStreamController.add(Future.error("Unsupported lnurl"));
```
roeierez, line 51: missing else?

lib/handlers/lnurl_handler.dart
```diff
@@ -24,60 +29,83 @@ class LNURLHandler {
...
+          action.future.catchError((err) {
+            promptError(context, "Open Channel Error",
+                Text("Failed to open channel.\n" + err.toString()));
+          }).whenComplete(() => Navigator.of(context).removeRoute(loaderRoute));
+        }
       }
+    });
+  }
+
+  void _listenLnLinks() {
```
roeierez, line 90: I think it is better to add the loader action to the main stream rather here.
It will be easier to understand.

erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/341/commits/e9fe825edf983b805ed1503de579786af09b6ddc)

lib/services/nfc.dart
```diff
@@ -132,7 +135,10 @@ class NFCService {
     } on PlatformException {
       print('Method "NFC event started with" exception was thrown');
     }
+    _checkNfcStartedWithTimer?.cancel();
```
roeierez, line 138: shouldn't we cancel only when we get a valid payload?

erdemyerebasmaz: The timer is to resolve the delay bug, we need to check for deep link once, because the app is either started with one or not.

lib/services/nfc.dart
```diff
@@ -91,7 +92,18 @@ class NFCService {
   }
 
   NFCService() {
-    Timer(Duration(seconds: 2), _listenLnLinks);
+    NfcPlugin nfcPlugin = NfcPlugin();
+    int fnCalls = 0;
+    _checkNfcStartedWithTimer =
+        Timer.periodic(Duration(milliseconds: 100), (Timer t) {
+      if (fnCalls == 5) {
+        _checkNfcStartedWithTimer.cancel();
```
roeierez, line 100: We should also cancel the timer when we get a valid payload otherwise risking in duplicate events.

erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/341/commits/32f45fe70cdca469dddb48e164a0d41162b05271)

PR #345, opened 2020-06-20
Adding splice in and splice out to the features list

fresheneesz:


kingonly:
Not relevant at the moment for the foreseeable future.

fresheneesz:
Could you elaborate on this? It it simply not necessary because breez has other mechanisms through which people can pay on chain?

kingonly:
There's no way to implement splicing at the moment.
With MPP/AMP and zero-conf channels, not sure it's needed.

fresheneesz:
What do you mean there's no way to implement it? I think there very much is a way to implement splicing. Is there something about Breez software that makes this particularly difficult? 

> With MPP/AMP and zero-conf channels, not sure it's needed.

AMP and zero conf channels are in no way a replacement for splicing. What do you mean by that? Why would AMP and zero conf channels make splicing "not needed"? This doesn't make sense to me.

kingonly:
So go ahead and fork. Splicing is not implement in any of the LN implementations.
Splicing is needed to dynamically increase the capacity of a channel. You don't need a large capacity channel when you can run and pay from multiple channels.

fresheneesz:
I can see that you're getting frustrated with me. I want Breez to succeed. All I'm trying to do is help, and understand. I usually see github issues on an open source project as a place where the community can discuss improvements to the project and come to a consensus on things. However, it seems like you're the only person from Breez on here, and you're not really willing to spend time discussing things. Is Breez just using these issues as a feedback form rather than as a place to discuss improvements to the product?

kingonly:
Not frustrated at all... I tried to explain why this PR is not relevant 🙂  If you wish to give arguments that supports your PR - go ahead, but you didn't respond to my input. Splicing is definitely not in the horizon and no major LN implementation is working on it. Maybe if/when it will be implemented, we'll support it, but with MPP/zero-conf I doubt there will be a need.

Breez is improving immensely from the feedback of the community. Just today we've committed fixes to a bug (Receive via Invoice screen in small screens), suggestion (lnurl string change) and a new feature (NFC support for lightning links)  - all initiated by Breez users. I suggest you take a look at the 40+ releases we've issues since we started our beta.

Btw, ad hominem certainly doesn't help.

fresheneesz:
I didn't respond to your input because it seemed like you didn't want to hear it. But since it seems you are open to it, I'll address your input. 

> With MPP/AMP and zero-conf channels, not sure it's needed.

> You don't need a large capacity channel when you can run and pay from multiple channels.

I see your point that with AMP, instead of adding funds to an existing channel, you could simply create a new channel and use all your channels in concert. However submarine swaps are what are currently used in place of splice out. Implementing a proper splice out eliminates the need for any intermediary. Surely this is a worthy benefit? 

> Breez is improving immensely from the feedback of the community

I see that. I never said anything that would imply otherwise.

> ad hominem certainly doesn't help.

Please point out my ad hominem. I think you're misusing the term. Looking back, I see nothing that I said that was an ad hominem. If you're trying to de-escalate the obvious contention between us, as I'm trying to do, accusing me of ad hominem is the opposite of effective in that respect.

kingonly:
We don't know what's a 'proper splice out' means yet simply because it's not implement. The end user doesn't care about the implementation details. The major benefit of splicing (in theory) is in dynamically increasing the channel capacity. If it would benefit in other user cases - we'll see.

Ad hominem is to say "I'm getting frustrated" instead of responding to the substance of my argument. Also, in your response now, your response was partial. Splicing is definitely not on the horizon and no major LN implementation is working on it. You should open an issue to lnd/c-lightning/Eclair if you'd like to see it get implement. It's not Breez feature. We could possibly use it to deliver a feature like dynamically increasing capacity or lowering on-chain fees.

fresheneesz:
> simply because it's not implement

I see, this node relies on LND features, and this is not one that's available yet. Fair enough.

> Ad hominem is to say "I'm getting frustrated" instead of responding to the substance of my argument.

That is not what ad hominem usually means. Ad hominem is when I attack your character as a way to discredit your argument, which is not at all what I was doing. What I was doing was trying to de-escalate what I saw was a potentially inflammatory situation that I was part of causing. I was neither attacking your character, nor using that to argue against what you were saying. I was simply trying to avoid potentially continuing to offend you, since that would make it harder to come to an understanding.

PR #348, opened 2020-06-29
Handle bitcoin QR string with amount parameter

erdemyerebasmaz:
This PR addresses [BU-236](https://breeztech.atlassian.net/browse/BU-236)

If QR string contains parameters, it's stripped of it's parameters before it's validated in breez library then the ReverseSwapPage is initialized with amount parameter if it's available and in a valid format.

Assumptions made: 
- amount field is always in BTC
- other use cases does not have parameters


lib/routes/home/floating_actions_bar.dart
```diff
@@ -140,8 +156,8 @@ class FloatingActionsBar extends StatelessWidget {
                               if (await _isBTCAddress(scannedString)) {
```
roeierez, line 156: Since this feature only applicable to btc address, I think the change should be here.
We can rename the `_isBTCAddress` to `parseBTCAddress` that would return an object with the address and amount parameter if exists. Also the parsing should be there instead in the QRCode scanner.
Parsing should use URL class and not regex, see `extractBolt11FromBip21` for example.

lib/routes/withdraw_funds/withdraw_funds_page.dart
```diff
@@ -277,12 +278,25 @@ class WithdrawFundsPageState extends State<WithdrawFundsPage> {
     });
   }
 
-  Future _scanBarcode() async {
+  Future _scanBarcode(AccountModel account) async {
     try {
       FocusScope.of(context).requestFocus(FocusNode());
       String barcode = await QRScanner.scan();
+      String requestAmount;
+      if (barcode.contains("?")) {
```
roeierez, line 286: Here I would use as well the `parseBTCAddress` that I suggested to create previously.

lib/routes/home/floating_actions_bar.dart
erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/348/commits/94c5a8c818751cbba4cba568668a20eae9bf7489)

lib/utils/btc_address.dart
```diff
@@ -0,0 +1,19 @@
+import 'package:breez/bloc/user_profile/currency.dart';
+import 'package:fixnum/fixnum.dart';
+
+BTCAddressInfo parseBTCAddress(String scannedString) {
+  Uri uri = Uri.parse(scannedString);
+  String address = uri.path;
+  Int64 satAmount;
+  if (uri.queryParameters["amount"] != null)
+    satAmount =
+        Currency.BTC.toSats(double.parse(uri.queryParameters["amount"]));
```
roeierez, line 10: We need to be more robust here. `tryParse`?

erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/348/commits/109375e373044e758405ad7f6da0c93af370a922)

PR #356, opened 2020-07-15
Open input dialog for our payments in clipboard

erdemyerebasmaz:
Clicking "Paste Invoice or Node ID" with our payments in clipboard did nothing as it was getting filtered out in invoice bloc.

The payment is filtered out before the invoice is added to decodeInvoice sink and if it's recognized as our own, enter payment info dialog shows up.

### DIFF
--- a/lib/bloc/invoice/invoice_bloc.dart
+++ b/lib/bloc/invoice/invoice_bloc.dart
@@ -63,7 +63,7 @@ class InvoiceBloc with AsyncActionsHandler {
     listenActions();
   }
 
-  Stream get decodedClipboardStream =>
+  Stream<Future<DecodedClipboardData>> get decodedClipboardStream =>
       device.rawClipboardStream.map((clipboardData) async {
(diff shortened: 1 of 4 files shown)


erdemyerebasmaz:
> @erdemyerebasmaz the access to the services layer should be from the bloc. Currently it is from the UI.
> The problem is that both of the following streams:
> `widget.invoiceBloc.clipboardInvoiceStream`
> `widget.invoiceBloc.clipboardNodeIdStream`
> returns string instead of more structured data.
> The change I would go for is to make them returned data that implies on the type as well, e.g:
> 
> ```
> {
>    data: "string",
>    type: oneOf("invoice", "nodeID")
> }
> ```
> 
> If the type is null it means it is not identified and we should show the dialog.
> The `clipboardInvoiceStream` can now check in the bloc if this is our invoice and return null in the type.

Streams are reduced into one, both nodeId and invoice decoding handled in this stream.
The stream now returns Map<String, String> and the check for if the invoice is ours or not is handled inside this stream on the bloc.

PR #357, opened 2020-07-16
Export and Import POS items

erdemyerebasmaz:
This PR addresses [POS-96](https://breeztech.atlassian.net/browse/POS-96).

"Items list" tile with a pop up menu is added to POS Settings page for import and export features.

- "Export to CSV"
Exports all items on catalog to a .csv file and opens up the share plugin. 
If the list is empty, an error "There are no items to export." is displayed via flushbar.

- "Import from CSV" 
Prompts the user with a dialog informing the user this process will overwrite the current item list.
After the user agrees to this dialog they will pick the file to be imported.
Once a valid file is selected, items are retrieved from that file and db is overwritten with those items.


lib/bloc/pos_catalog/bloc.dart
```diff
@@ -103,6 +108,30 @@ class PosCatalogBloc with AsyncActionsHandler {
...
+      throw Exception("EMPTY_LIST");
+    }
+  }
+
+  _importItems(ImportItems action) async {
+    List<Item> itemList =
+        await PosCsvUtils().retrieveItemListFromCSV(action.importFile);
+    // We should try to restore old db if import fails.
+    // var backupDB = await _repository.fetchItems();
+    await resetDB();
```
roeierez, line 125: Resetting the db and importing the new items should be executed in one db transaction otherwise we may end with empty database.

lib/bloc/pos_catalog/pos_csv_utils.dart
```diff
@@ -0,0 +1,108 @@
...
+  List _generateList() {
+    log.info("generating payment list started");
+    List<List<dynamic>> paymentList =
+        List.generate(this.itemList.length, (index) {
+      List paymentItem = List();
+      Item paymentInfo = this.itemList.elementAt(index);
+      paymentItem.add(paymentInfo.id.toString());
+      paymentItem.add(paymentInfo.name);
+      paymentItem.add(paymentInfo.sku);
+      paymentItem.add(paymentInfo.imageURL);
```
roeierez, line 32: Currently null values will be printed as 'null' in the CSV. Isn't it better to print empty strings instead?

lib/routes/settings/pos_settings_page.dart
```diff
@@ -159,6 +171,119 @@ class PosSettingsPageState extends State<_PosSettingsPage> {
...
+            PopupMenuItem(
+              height: 36,
+              value: Choice(() => _exportItems(context, posCatalogBloc)),
+              child: Text('Export to CSV',
+                  style: Theme.of(context).textTheme.button),
+            ),
+          ],
+        ),
+      ),
+      onTap: () => _onChangeAdminPasswordSelected(),
```
roeierez, line 213: Why changing password?

lib/routes/settings/pos_settings_page.dart
```diff
@@ -159,6 +171,119 @@ class PosSettingsPageState extends State<_PosSettingsPage> {
...
+          Navigator.of(context).push(loaderRoute);
+          action.future.then((_) {
+            Navigator.of(context).removeRoute(loaderRoute);
+            Navigator.of(context).pop();
+            showFlushbar(context, message: "Items were successfully imported.");
+          }).catchError((err) {
+            Navigator.of(context).removeRoute(loaderRoute);
+            var errorMessage = "Failed to import POS items.";
+            switch (err.toString()) {
+              case "INVALID_FILE":
```
roeierez, line 251: Better to have these strings as constants on the action. Or even better we can use customize errors as constants so we compare errors and not strings.

lib/routes/settings/pos_settings_page.dart
erdemyerebasmaz: I was reusing password tile...
[Removed](https://github.com/breez/breezmobile/pull/357/commits/8dfa357acfe9ace72c0555dcc058f1b65abb028c)

lib/bloc/pos_catalog/bloc.dart
erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/357/commits/875ac92106e860ae72da9b32100f4797ffadfdbc)

lib/routes/settings/pos_settings_page.dart
erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/357/commits/a662057afe38100bb96e5b5456ca5716ef2b4abb)

erdemyerebasmaz:
> Only suggest to rename the "resetDB" to "importItems" and good to go.

[Renamed](https://github.com/breez/breezmobile/pull/357/commits/8ad6b8d237458a2639ee8e2811a77bd0423ea3a1)

PR #372, opened 2020-10-04
Home Page Reworked UI

erdemyerebasmaz:
This PR addresses [BU-244](https://breeztech.atlassian.net/browse/BU-244)

Home page UI has been reworked using the designs on: [Figma](https://www.figma.com/proto/pNMrKD37uR7aklFrcInird/Breez?node-id=102%3A9911&scaling=min-zoom)

- Floating Actions Bar is replaced with Bottom Actions Bar and QrActionButton.
- Long press behavior on balance is removed and now fiat conversion is shown below balance if applicable(above min amount etc.)
- Removed waves images from home page.
- Applied the new text styles and color theme.
- Fixed hitbox of fullscreen loader buttons.


roeierez (approved):
We need to test the home screen really good.

PR #377, opened 2020-10-13
Payment list flashing bug

erdemyerebasmaz:
This PR addresses [BU-256](https://breeztech.atlassian.net/browse/BU-256)

I couldn't reproduce the bug on any physical or virtual device so we decided to go with this workaround where the avatar fades away as it's leaving the viewport.

 - Title, amount text and bottom divider fades away too.

Concerns: Using Opacity is costly, we may revert to using this trick for avatar only if performance issues arises for our users. I had the same frame rate while scrolling on my tests with and without this workaround.

### DIFF
--- a/lib/routes/home/account_page.dart
+++ b/lib/routes/home/account_page.dart
@@ -130,8 +130,11 @@ class AccountPageState extends State<AccountPage>
     }
 
     if (paymentsModel.nonFilteredItems.length > 0) {
-      slivers.add(PaymentsList(paymentsModel.paymentsList,
-          PAYMENT_LIST_ITEM_HEIGHT, widget.firstPaymentItemKey));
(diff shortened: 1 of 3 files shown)


roeierez:
@erdemyerebasmaz  in general I understand we need an optimisation here and OK with the approach which reduces rendering load on non-visible items.
Currently though, this is handled inside the PaymentItem which requires passing it with the scroll position so it can make a decision of doing light weight rendering.
Two questions:
1. Did you try to handle the rendering decision at the list level where the scroll position is already accessible?
2. Instead of making the specific elements within the payment item transparent, did you try make the whole item transparent? or just render an empty SizedBox? I think this will make the code more maintainable.

roeierez (approved):

PR #394, opened 2020-10-28
Fiat currency settings

erdemyerebasmaz:
This PR addresses [BU-191](https://breeztech.atlassian.net/browse/BU-191?atlOrigin=eyJpIjoiNDUzMTU2YTlmOWM1NGM1NjgxNDc1OWI2NzE4NDM4YTAiLCJwIjoiaiJ9)

With the addition of new fiat currencies, we're adding fiat currency settings page where the users can select & order their preferred fiat currencies.


lib/routes/fiat_currencies/fiat_currency_settings.dart
```diff
@@ -0,0 +1,274 @@
...
+
+  void _getUserFiatCurrencyPreferences() {
+    _userProfileBloc.userStream.firstWhere((user) => user != null).then((user) {
+      _preferredFiatCurrencies =
+          List.from(user.fiatCurrencyPreferences.preferredFiatCurrencies);
+    });
+  }
+
+  void _initializeFiatCurrencyList() {
+    _accountBloc.accountStream
```
roeierez, line 78: Seems better to use Stream builder instead of fetching and sorting on init.

lib/routes/fiat_currencies/fiat_currency_settings.dart
```diff
@@ -0,0 +1,274 @@
...
+                canBeDraggedTo: (from, to) =>
+                    to < _selectedFiatConversions.length,
+                dragElevation: 8.0,
+              );
+            }),
+      ),
+    );
+  }
+
+  _onWillPop() async {
```
roeierez, line 157: Usually users either have "Save" button to save changes or the changes are saved as they edit.
Here the save happens implicitly when user click "Back". I think it is better to save changes on each change and remove this code.
One side effect with the current code is that user gets different behaviour when pressing the back menu and when pressing the back android button.

lib/widgets/currency_converter_dialog.dart
```diff
@@ -123,6 +131,11 @@ class CurrencyConverterDialogState extends State<CurrencyConverterDialog>
               .exchangeRate;
           _updateExchangeLabel(exchangeRate);
 
+          List<FiatConversion> usersFiatConversions = List.from(account
```
roeierez, line 134: This patterns repeats itself and I think the account model can encapsulate that. 
The Account bloc listens for user changes already and push a new AccountModel with each change so we can add a property on the AccountModel preferredFiatConversions that is already the list you need here.
We will save all the previous line of codes as well.
Also a better way is to use the StreamBuilder instead of this initialisation code.

roeierez (changes requested):
Initial pass.

lib/home_page.dart
```diff
@@ -389,6 +390,19 @@ class HomeState extends State<Home> {
                                                       ...posItem,
                                                       DrawerItemConfigGroup(
                                                           _filterItems([
+                                                            DrawerItemConfig(
+                                                                "",
+                                                                "Fiat Currencies",
+                                                                "src/icon/fiat_currencies.png",
+                                                                onItemSelected: (_) => showDialog(
```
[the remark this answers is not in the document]
erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/394/commits/c84ff006d02c3ffa5de45b8313288c975f0fd1b1)
FiatCurrencySettings was initially created as FiatCurrencySettingsDialog.

lib/widgets/currency_converter_dialog.dart
erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/394/commits/b735bf0282c1cdff6ac9266c28c9d2c6cd23e13c)

lib/routes/fiat_currencies/fiat_currency_settings.dart
erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/394/commits/1381bcf8bcb3e0a35d8f3893f8d8057de37a08bc)

lib/routes/fiat_currencies/fiat_currency_settings.dart
erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/394/commits/d6c942f95931ca79fd135292915b3a08c07d56aa)

Sort logic is moved to bloc.

roeierez (approved):

PR #400, opened 2020-11-02
Add Print button to Successful Payment dialog and valid Transactions

erdemyerebasmaz:
Print button is added to Successful Payment dialog and sale views of valid transactions on transaction list.

Pressing print creates a pdf with following layout:
- Business Name
- Business Address
- Sale Table(similar formatting to Sale View)


roeierez (approved):
@erdemyerebasmaz looks good.
Just make sure you check the PDF with RTL currency and I think this is good to go.

PR #405, opened 2020-11-05
Print Payment as PDF

erdemyerebasmaz:
This PR addresses [BU-267](https://breeztech.atlassian.net/browse/BU-267).
Reviewed Parts: https://github.com/breez/breezmobile/pull/399 and https://github.com/breez/breezmobile/pull/400

With this feature users will be able to print and share their invoices upon successful payment or by accessing the invoice from old transactions.

- Arabic characters are now supported. Updated widgets to support text direction.
- Transaction Time, Payment Preimage and Sale Note informations are added.


lib/routes/charge/pos_invoice.dart
```diff
@@ -763,7 +769,12 @@ class POSInvoiceState extends State<POSInvoice> with TickerProviderStateMixin {
         }).then((res) {
       if (res?.paid == true) {
         Navigator.of(context).push(TransparentPageRoute((context) {
-          return SuccessfulPaymentRoute();
+          return SuccessfulPaymentRoute(
+            currentUser: user,
```
roeierez, line 774: All these additional parameters only passed from POS flow complicates the widget and all its children hierarchy.
We should find a. better and simple way to separate the flows such as  pass a widget for the final desired "SuccesfullPaymentMessage" to override the default behaviour.

lib/routes/charge/sale_view.dart
```diff
@@ -235,6 +226,49 @@ class SaleViewState extends State<SaleView> {
           );
         });
   }
+
+  _buildActions(AccountModel account) {
+    if (account != null) {
```
roeierez, line 233: Doesn't seem to be an intuitive way to distinguish between read only and in progress sales view. No reason it shouldn't be explicit.
I suggest to separate to two functions one for readonly and one for the editable view.

roeierez (changes requested):

lib/routes/charge/sale_view.dart
erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/405/commits/68e5e421f479aac77208714bcb43fb65226b4e2b)

lib/routes/charge/pos_invoice.dart
erdemyerebasmaz: [Something like this?](https://github.com/breez/breezmobile/pull/405/commits/522e4b2b5a838796bc79ca1ce63225468ea0bf1d)

erdemyerebasmaz: I don't see why we have two classes here instead of a method that builds SuccesfullPaymentMessage.

roeierez: @erdemyerebasmaz  this is better than before although I think it is better to only pass a function (print()) instead of print parameters that needs to go all the way down.

roeierez (commented):
@erdemyerebasmaz added one comment regarding the PrintParameters to your consideration.
Rational is to keep the SuccesfullPaymentRoute compact and with minimal dependancies. 
Once this is resolved we can push this in.

lib/routes/charge/pos_invoice.dart
erdemyerebasmaz: [Done](https://github.com/breez/breezmobile/pull/405/commits/56ab8d522d3ef67a2f3d91f3adbe8a2c114fe776)

roeierez (approved):

PR #412, opened 2020-12-17
Update webview plugin

erdemyerebasmaz:
We are replacing our unofficial webview fork with flutter's official webview plugin as it's being maintained regularly and it now solves the issues we have solved on our fork.

Now that webview is a part of the app lifecycle, widgets such as lockscreen and dialog can be shown on top of webview without the hide/show & takeScreenshot logic.
WebLN and post message logic is moved to bidirectional JavaScript channel.
NavigationDelegate is used to prevent redirects to "lightning" scheme url's.

### DIFF
--- a/src/scripts/lightningLinkInterceptor.js
+++ b/src/scripts/lightningLinkInterceptor.js
@@ -11,10 +11,6 @@ function interceptClickEvent(e) {
     }
     if (target.tagName === 'A') {
         href = target.getAttribute('href');
-    } else if (target.tagName === 'svg' && target.parentElement && target.parentElement.tagName === 'A') {
-        href = target.parentElement.getAttribute('href');
-    } else if (target.tagName === 'rect' && target.parentElement && target.parentElement.tagName === 'svg' && target.parentElement.parentElement && target.parentElement.parentElement.tagName === 'A') {
-        href = target.parentElement.parentElement.getAttribute('href');
     }
     console.log("interceptClickEvent href=" + href);
     if (href && href.startsWith("lightning:")) {
@@ -29,3 +25,10 @@ if (document.addEventListener) {
 } else if (document.attachEvent) {
     document.attachEvent('onclick', interceptClickEvent);
 }
+
+// disable clicks on SVG elements
+((d,e)=>{
+  e = d.createElement("style");
+  e.innerHTML = "svg{pointer-events: none;}";
+  d.head.appendChild(e)
+})(document)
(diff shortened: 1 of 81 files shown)


src/scripts/lightningLinkInterceptor.js
```diff
@@ -9,13 +9,17 @@ function interceptClickEvent(e) {
...
         }
     }
-    // if (target.tagName === 'A') {
-    //     href = target.getAttribute('href');
-    //     console.log("interceptClickEvent href=" + href);
-    //     if (href && href.startsWith("lightning:")) {
-    //         window.postMessage(JSON.stringify({"lightningLink": href}), "*");
-    //     }
-    // }
+    if (target.tagName === 'A') {
```
roeierez, line 12: Can we simplify this method by just querying 2-3 levels for parents that contains lightning link? like a simple loop instead of these conditions? Maybe also we don't care if the child is 'svg' or 'rect'?

roeierez (commented):
@erdemyerebasmaz first pass looks good.
I still need to do some testing and another pass afterwards.

src/scripts/lightningLinkInterceptor.js
erdemyerebasmaz: We can replace it with this:
```
target.closest("a")
```
but it's likely to prompt wrong/unwanted invoices as it can go above 2-3 levels.

We can run this for 3 layers with a counter:
```     
var counter = 0;
while(target.parentElement && counter < 2) {
    target = target.parentElement;
    if(target.hasAttribute('href')) {
        href = target.getAttribute('href');
        break;
    }
    counter++;
}
```
or we can inject css to disable clicks on SVG
```
((d,e)=>{
  e = d.createElement("style");
  e.innerHTML = "svg{pointer-events: none;}";
  d.head.appendChild(e)
})(document)
```

roeierez: @erdemyerebasmaz disabling svg may disable non lightning links as well right?

roeierez (changes requested):
@erdemyerebasmaz that looks good. Please see my comment regarding lightning links handling.

src/scripts/lightningLinkInterceptor.js
erdemyerebasmaz: @roeierez Not likely, we are disabling clicks on svg element, if it is surrounded by an element that has a link( <a href=""> or <button> with onclick etc.) it will follow through that link.

PR #416, opened 2021-01-21
Bumped uni_links version

jamesonev:
uni_links is outdated, but the new version doesn't contain any [breaking changes](https://pub.dev/packages/uni_links/changelog) and removes several compiler warnings around nullable pointers.


roeierez (approved):
Thanks for the PR @jamesonev!

jamesonev:
I misread the digits in spec and didn't realize that this would resolve anyway; I'm trying to get the repo up and running locally and saw an error here but it was a red herring. I'll try and submit a PR shortly with the dependencies that I've actually updated.

PR #422, opened 2021-02-02
Podcast theme regression

erdemyerebasmaz:
Transformed withPodcastTheme to a wrapping widget for theme wrapping.

 - Applied our dialog theme to podcast theme.
 - Added overridePodcastTheme variable to override podcast theme for child widgets, which is used on pages under Preferences and their children.

### DIFF
--- a/lib/home_page.dart
+++ b/lib/home_page.dart
@@ -1,6 +1,6 @@
 import 'dart:async';
 import 'dart:io';
-import 'package:breez/routes/podcast/theme.dart';
+
 import 'package:anytime/ui/anytime_podcast_app.dart';
 import 'package:breez/bloc/account/account_actions.dart';
 import 'package:breez/bloc/account/account_bloc.dart';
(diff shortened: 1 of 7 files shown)


roeierez (commented):
@erdemyerebasmaz if we are going this way I would prefer to use two functions instead of a flag "overridePodcastTheme"

withPodcastTheme - to wrap with podcast theme
withBreezTheme - to wrap with breez original theme.

erdemyerebasmaz:
> @erdemyerebasmaz if we are going this way I would prefer to use two functions instead of a flag "overridePodcastTheme"
> 
> withPodcastTheme - to wrap with podcast theme
> withBreezTheme - to wrap with breez original theme.

[Done](https://github.com/breez/breezmobile/pull/422/commits/ab9b7c87e7467eab2f0fb717f218962215023aaa)

roeierez (approved):

PR #470, opened 2021-04-22
Add ability to share podcasts

erdemyerebasmaz:
This PR addresses [BU-321](https://breeztech.atlassian.net/browse/BU-321)

Users will be able to share podcasts via share button that's added to every podcasts detail page.
Clicking this share button will create a Firebase dynamic link that has feedURL info of the podcast. This link will use the existing deeplink entry point for C2P links.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -92,6 +92,14 @@
                 <data android:scheme="breez"/>
             </intent-filter>
 
+            <intent-filter android:autoVerify="true">
+                <action android:name="android.intent.action.VIEW"/>
+                <category android:name="android.intent.category.DEFAULT"/>
+                <category android:name="android.intent.category.BROWSABLE"/>
+                <data android:host="www.breez.link" android:scheme="http"/>
+                <data android:host="www.breez.link" android:scheme="https"/>
+            </intent-filter>
+
             <intent-filter>
                 <action android:name="FLUTTER_NOTIFICATION_CLICK" />
                 <category android:name="android.intent.category.DEFAULT" />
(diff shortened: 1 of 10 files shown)


roeierez:
@erdemyerebasmaz Did an initial pass and overall it looks good.
Still missing the logic to wire the specific episode, I will wait for that on my second review.

roeierez (changes requested):
@erdemyerebasmaz using the url change event seems a bit problematic to me such as:
1. It requires the podcast page to be rendered (links are not opened if we are in balanace view).
2. It doesn't open the link twice in a row.
In addition the handleDeepLink method is implemented in anytime and I think by running it in breez side we will get some advantages such as:
1. We control the podcast and episode construction so we can later build them using podcast index API and not RSS.
2. We control the loading time and do not need to add a loader inside anytime project.
3. Being able to push the "podcast details view" and "episode view" from outside anytime seems like a well defined contract to me.

I have implemented that in the last commit so you can review and let me know what you think.
If you agree please remove the redundant code in anytime.

erdemyerebasmaz:
> @erdemyerebasmaz using the url change event seems a bit problematic to me such as:
> 
> 1. It requires the podcast page to be rendered (links are not opened if we are in balanace view).
> 2. It doesn't open the link twice in a row.
>    In addition the handleDeepLink method is implemented in anytime and I think by running it in breez side we will get some advantages such as:
> 3. We control the podcast and episode construction so we can later build them using podcast index API and not RSS.
> 4. We control the loading time and do not need to add a loader inside anytime project.
> 5. Being able to push the "podcast details view" and "episode view" from outside anytime seems like a well defined contract to me.
> 
> I have implemented that in the last commit so you can review and let me know what you think.
> If you agree please remove the redundant code in anytime.

@roeierez  Changes are looking good. I've removed the redundant code in anytime.

roeierez:
> @roeierez Changes are looking good. I've removed the redundant code in anytime.

Thanks @erdemyerebasmaz

PR #482, opened 2021-05-23
Add ability to add custom boost & sats per min amounts

erdemyerebasmaz:
This PR addresses [BU-323](https://breeztech.atlassian.net/browse/BU-323)

Clicking on the boost or sats/min amounts will open a input dialog for users to enter a custom amount.

- The custom amount can’t be lower than minimum preset value (but can be higher than preset max value).
- +/- will set the amount to nearest value from preset values.

- Added information text to tutorials to introduce this behavior.


lib/routes/podcast/boost.dart
```diff
@@ -235,4 +289,15 @@ class _BoostWidgetState extends State<BoostWidget> {
               });
         });
   }
+
+  static int _getClosestBoostAmount(int customAmount, List presetAmountList,
```
roeierez, line 293: Can't we use this function also for the equal case? This way we won't need to split the condition depends on whether the custom amount is taken from the list or not.
Also I would make this function more readable by giving each case its own line, it is OK if it will be a little bit longer IMO.

lib/routes/podcast/payment_adjuster.dart
```diff
@@ -159,4 +203,16 @@ class _PaymentAdjusterState extends State<PaymentAdjuster> {
           );
         });
   }
+
+  static int _getClosestSatsPerMinAmount(
```
roeierez, line 207: Same here regarding reusing for the equal case.

roeierez (changes requested):

lib/routes/podcast/boost.dart
erdemyerebasmaz: We can split this into _getNextAmount and _getPreviousAmount methods and get rid of the condition and index logic.

erdemyerebasmaz:
Will merge persist_custom_payment_amount branch instead.

PR #493, opened 2021-06-03
LNURL-Pay.

nochiel:
This PR depends on [PR #135](https://github.com/breez/breez/pull/135).

This PR is ready for testing. Bugs and missing functionality are listed below.

## TODO

- [ ] Test 
    - [x] <http://bitrefill.com> wallet deposit.
    - [x]  <http://coinos.io> wallet deposit.
    - [x] Bug: <http://zebedee.io> lnurl-pay QR codes aren't handled.
        - Fix: BreezMobile uses qr_code_tools. Upstream [qr_code_tools needs to upgrade its zxing dependency](https://github.com/AifeiI/qr_code_tools/pull/9).
    - [x] lnurl-toolbox.
    - [x] <https://lightning.gifts>
- [x] Handle `SuccessAction`.
   - [x] Handle `SuccessAction` "aes" tag.
       - [ ] We need a way to test decryption of lnurl-Pay messages. I'll configure `LNBits` for this.
- [x] Save `description` from `service`. (description hash is already saved in the invoice `h` tag.)
- [x] Save user's `comment`.
- [ ] Display user's `comment` with transaction details.
    - [ ] Feature; Should users be able to label/comment any transaction (as we do in Bitcon wallets, for coin control?). 
- [x] Code cleanup.
    - [x]  Make sure any new files don't have extraneous dependencies.
    - [x] Remove unnecessary comments. 
- [x] Bug: Errors from rpc server are not being received in pay_invoice_page.dart. Why have errors stopped working?
- [ ] Bug: lnurls in the clipboard should be detected and optionally trigger lnurl-pay flow.


roeierez:
Hi @nochiel Thanks a lot for your progress.
I will do an initial review today so you can move on.

lib/bloc/lnurl/lnurl_bloc.dart
```diff
@@ -17,6 +17,12 @@ class LNUrlBloc with AsyncActionsHandler {
 
   StreamController _lnUrlStreamController;
 
+  final _newPayRequestController = StreamController<PayFetchResponse>();
```
roeierez, line 20: Why do we need this stream controller and its corresponding stream?
As far as I understand the fetch action is handled by the _listenLNUrl_ function that returns a stream directly to the _lnurl_handler.dart_.

roeierez (changes requested):
@nochiel I think you got the idea here, see my comment regarding the new stream, unless I am missing something I don't see it is needed.
Now that you added support for the parsing and handling of LNURL Pay the next thing you should implement is the UI in lnurl_handler.dart.

roeierez:
> * How do I test LNURl-Pay in production?
Can you DM me in telegram ( @roeierez ) ?

lib/bloc/lnurl/lnurl_bloc.dart
nochiel: > Why do we need this stream controller and its corresponding stream?
> As far as I understand the fetch action is handled by the _listenLNUrl_ function that returns a stream directly to the _lnurl_handler.dart_.

`_listenLNURL` captures the first request from client to server.  (Step. 2 and Step 3 in the [spec](https://github.com/fiatjaf/lnurl-rfc/blob/master/lnurl-pay.md).

In step 5 the client makes a second request with an amount to get an invoice from the server.

I'm re-using the PayFetchResponse class and updating its amount for step 5 but I should make a new class to make the difference between step 2 and step 5 clear.

nochiel:
> > * How do I test LNURl-Pay in production?
> 
> Can you DM me in telegram ( @roeierez ) ?

I've sent you a DM. My uid is @Nick0o0o.

nochiel:
> the next thing you should implement is the UI in lnurl_handler.dart.

I've added files to `routes\create_invoice` for the lnurl-pay ui.

lib/bloc/lnurl/lnurl_bloc.dart
roeierez: I see. Makes sense overall.
I will do another review for the new commits later today.

lib/routes/create_invoice/pay_invoice_page.dart
```diff
@@ -0,0 +1,500 @@
...
+         5. LN WALLET makes a GET request using callback with the following query parameters:
+         amount (input) - user specified sum in MilliSatoshi
+         nonce - an optional parameter used to prevent server response caching
+         fromnodes - an optional parameter with value set to comma separated nodeIds if payer wishes a service to provide payment routes starting from specified LN nodeIds
+         comment (input) - an optional parameter to pass the LN WALLET user's comment to LN SERVICE
+
+       */
+
+    _payFetchResponse.amount = account.currency.parse(_amountController.text);
+    lnurlBloc.newPayRequestSink.add(_payFetchResponse);
```
roeierez, line 316: At this point we are requesting the invoice (step 5 in the spec).
An alternative way to create a new payment implementation ( `LNURLPayDialog` ) would be to wait for the result and merge into the current flow of paying an invoice ( `PaymentRequestDialog` ).
To achieve the pattern of waiting we can add an AsyncAction that adds executes the request inside the bloc and return a promise with the result.
What do you think?

roeierez (changes requested):

lib/handlers/lnurl_handler.dart
```diff
@@ -50,7 +50,7 @@ class LNURLHandler {
       });
 
       Navigator.of(context).push(FadeInRoute(
-        builder: (_) => CreateInvoicePage(lnurlWithdraw: response),
+        builder: (_) => PayInvoicePage(lnurlPay: response),
```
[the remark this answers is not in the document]
nochiel: > Why did you change the withdraw handling? The PayFetchResponse doesn't expect `WithdrawFetchResponse` as argument and it will throw an error.

That's a unintentional error. Sorry. Reverted.

lib/routes/create_invoice/pay_invoice_page.dart
nochiel: > At this point we are requesting the invoice (step 5 in the spec).
> An alternative way to create a new payment implementation ( `LNURLPayDialog` ) would be to wait for the result and merge into the current flow of paying an invoice ( `PaymentRequestDialog` ).
> To achieve the pattern of waiting we can add an AsyncAction that adds executes the request inside the bloc and return a promise with the result.

When I obtain the `invoice` does it make sense to add it to add it to  `decodeInvoiceSink`?

  widget.invoiceBloc.decodeInvoiceSink.add(widget.payinfo.invoice)

I was hoping that:
- the invoice is decoded and added to `invoiceBlock.readyInvoicesStream`.
- the decoded invoice and payment flow is handled by `received_invoice_notification.dart`.
- Navigation route pops back to lnurl-pay flow so we can show `successAction`.

Does this make sense?

> What do you think?

I'll implement your suggestion and test. Thanks.

roeierez: Sounds good. That is what was in my mind as well.

lib/routes/create_invoice/pay_invoice_page.dart
```diff
@@ -0,0 +1,393 @@
...
+    action.future.catchError((e) {
+      promptError(
+        context,
+        "An error occurred while attempting to retreive an invoice from ${_payFetchResponse.host}!",
+        Text("Reason: ${e.toString()}",
+            style: Theme.of(context).dialogTheme.contentTextStyle),
+      );
+    }).then((payinfo) {
+      var navigator = Navigator.of(context);
+      navigator.pop();
```
roeierez, line 368: Before popping here, I suggest to check if the user cancelled by closing the dialog (back on android).
One way to do it is to check if the current widget is disposed and continue with the invoice only if it is not.

lib/bloc/lnurl/lnurl_actions.dart
```diff
@@ -28,3 +28,9 @@ class Login extends AsyncAction {
 
   Login(this.response, {this.jwt = false});
 }
+
+class Pay extends AsyncAction {
```
roeierez, line 32: This action is used to fetch the invoice on second stage. Maybe use better naming? "FetchInvoice" or something similar?

roeierez (changes requested):
That looks good and solid! Only success action is missing. Can't wait.

lib/routes/lnurl_fetch_invoice_page.dart
```diff
@@ -0,0 +1,376 @@
...
+
+      /*
+           FIXME Verify that this works when:
+           "Before popping here, I suggest to check if the user cancelled by closing the dialog (back on android).  One way to do it is to check if the current widget is disposed and continue with the invoice only if it is not." (@roeierez) 
+      */
+
+      if (mounted && _currentRoute.isCurrent) {
+        var navigator = Navigator.of(context);
+        navigator.pop();
+      }
```
nochiel, line 364: @roeierez 

re: https://github.com/breez/breezmobile/pull/493/#discussion_r651738611

Is this correct?

roeierez: I think in case the _currentRoute is not _current_ anymore it means that the user has pressed back and probably meant to cancel the flow.
In that case for me it seems more correct to just return and do nothing.
Maybe similar to this:

```
if (!_currentRoute.isCurrent) {
   return;
}
Navigator.of(context).pop();
```

Does it make sense to you as well?

roeierez (changes requested):

roeierez (changes requested):
@nochiel it is great to see how you covered everything. That's impressive.
Regarding the success action I think we can make one change that will make everything a bit more simple.
Eventually the success action should propogate to the payment item in the UI so instead of maintaining the payInfoController stream and then adding a logic to map success actions to payments, I think we can add the success action on the payment item model and do that in breez library (At the same place where you populate the description.
This way it would be available on the paymentItem model eliminating the need for mapping logic.
What do you think?

nochiel:
@roeierez Thank you for all your help and the opportunity to contribute to Breez. I'm deeply grateful.

I'm implementing the small change you suggested.

Can we now discuss what I need to do so that the LNUrl-Pay PRs can be merged? I'm ready to work! I'll add what you want me to do to the checklist and move as fast as I can to get the PRs ready.

roeierez:
> I'm implementing the small change you suggested.

@nochiel  I think maybe I didn't explain myself well.
What I meant is that we can populate the lnurl info on the payment item in breez here: https://github.com/breez/breez/pull/135/files#diff-1fe6cded4ebb9c7862f6f6d062b47a4b41a4e3c339ca953b9b450325b6184332R1110

You already does it for the description so I think it makes sense to populate other related info as well.
This way we will have it automatically on the UI here: https://github.com/breez/breezmobile/pull/493/files#diff-17f8452748e541d07509f77382d2eacff861439647d365a55fbb9606b081fd57R47

Ans we won't need to propagate the `_lnurlPayInfos` all the way down just to search for that info.
We can talk on telegram if that is more convenient.



> Can we now discuss what I need to do so that the LNUrl-Pay PRs can be merged? I'm ready to work! I'll add what you want me to do to the checklist and move as fast as I can to get the PRs ready.

Sure. The way I see it is that apart from the above change I think we only missing showing a dialog after payment succeed with the success action, maybe here after the flushbar? https://github.com/breez/breezmobile/pull/493/files#diff-8f6001a504bced7db052edf9001ad691c1aeddc38228fae1566d87af0bb50c8cR768
After addressing these two items Me and the team will start testing it, clearing the path to merge this great feature.

nochiel:
> @nochiel I think maybe I didn't explain myself well.
> What I meant is that we can populate the lnurl info on the payment item in breez here: https://github.com/breez/breez/pull/135/files#diff-1fe6cded4ebb9c7862f6f6d062b47a4b41a4e3c339ca953b9b450325b6184332R1110

Aha! I understand. Thanks for the clarification. I can make that change quickly.

> The way I see it is that apart from the above change I think we only missing showing a dialog after payment succeed with the success action, maybe here after the flushbar? https://github.com/breez/breezmobile/pull/493/files#diff-8f6001a504bced7db052edf9001ad691c1aeddc38228fae1566d87af0bb50c8cR768

👍🏾

lib/widgets/payment_details_dialog.dart
```diff
@@ -278,6 +284,18 @@ List<Widget> _getStreamedPaymentInfoDetails(StreamedPaymentInfo paymentInfo) {
   }).toList();
 }
 
+List<Widget> _getLNUrlSuccessActionForPayment(
+    BuildContext context, SuccessAction sa) {
+  return <Widget>[
+    if (sa.tag == 'url')
+      ShareablePaymentRow(title: "Description", sharedValue: sa.description),
+    if (sa.tag == 'url')
+      ShareablePaymentRow(title: "URL", sharedValue: sa.url), // TODO Hyperlink.
```
vital-edu, line 293: Why do you need to check twice if `sa.tag == 'url'`?

nochiel: I didn't know how to use spread syntax in conditional list construction. Now I know. Will fix.

PR #494, opened 2021-06-08
Fix ln pizza asset on apps page

ademar111190:
The logo `ln.pizza_logo_lg.png` was misstyped as `lnpizza_logo_lg.png` on `conf/marketplace.conf` file.

how it looks like before and after the fix:

|Before|After|
|---|---|
|photo_2021-06-08 15 02 53|photo_2021-06-08 15 02 51|


roeierez (approved):
Thank you!

PR #495, opened 2021-06-10
Make tests pass again 💪 🚀

ademar111190:
I have been studying the project and I notice the tests through the `flutter test` are not passing, at least not in my machine.

I manage to make the tests pass with those changes.

Please let me know if something I did should not be made or be maybe in another way.


roeierez (approved):
Very nice! Thanks a lot @ademar111190 .

PR #496, opened 2021-06-15
Increase test coverage of sqlite/repository from 81% to 100%

ademar111190:
# Coverage of `sqlite/repository` before and after:

## Before

repository - from

## After

repository - to


roeierez (approved):
Fantastic! Thanks a lot!

PR #497, opened 2021-06-15
Increase test coverage of fastbitcoins from 76% to 100%

ademar111190:
Increase test coverage of `fastbitcoins` group from 76% to 100%

# Before

fastbitcoins - from

# After

fastbitcoins - to


roeierez (approved):
Another great addition @ademar111190 ! LGTM

ademar111190:
Thank you very much @roeierez 🙏 

It would be perfect if the CI ran correctly too.

roeierez:
> It would be perfect if the CI ran correctly too.

I agree. Currently we are running a mobile build and push it to test flight. I need to change that in case of PRs to only run the tests. I will try to get to it soon.

PR #498, opened 2021-06-15
Add tests for lib/utils files

ademar111190:
Increases the coverage from 54% to 90%

# Before

Screenshot 2021-06-15 at 17 00 54

# After

Screenshot 2021-06-15 at 17 30 53

# What is missing

Screenshot 2021-06-15 at 17 31 03

The line `FormatException` is never called because of the check `if (lowerBip21.startsWith("bitcoin:")) {` that causes `Uri.parse` to create a `_SimpleUri`, the exception is thrown on special cases when Uri.parse try to parse a `http` or `file:` uri but because of the check it never happens.

# About formation

I just get the output and set as expected on the tests but I'm not sure if `'0'` is correct instead of `'0.00'` or `'12 345 678.9'` instead of `'12,345,678.90'`

### DIFF
--- a/test/utils/currency_formatter_test.dart
+++ b/test/utils/currency_formatter_test.dart
@@ -1,0 +1,26 @@
+import 'package:breez/utils/currency_formatter.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+CurrencyFormatter make() => CurrencyFormatter();
+
+void main() {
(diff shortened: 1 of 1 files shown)


roeierez (approved):
Thanks for this as well! Great to see such important work keeps coming.

Issue #170, opened 2019-09-07
Suggestion: have spaces between numbers


bitcoinuser:
Hi,

When inserting an amount on Sat amount field, should have spaces like how appear on receive up to. See the image:

![photo5159248921254537267](https://user-images.githubusercontent.com/1178856/64478053-fcdaf700-d178-11e9-9af5-17d948d7e831.jpg)

bitcoinuser:
Same thing on dollars field:

![photo5159248921254537268](https://user-images.githubusercontent.com/1178856/64478068-36136700-d179-11e9-9a09-c162df27146c.jpg)

bitcoinuser:
![photo5042286589438437433](https://user-images.githubusercontent.com/1178856/68089764-1f248480-fe4b-11e9-84c5-2f664406b4ca.jpg)

kingonly:
Fixed in latest

PR #500, opened 2021-06-17
Add satoshi amount form formatter for amount form field

ademar111190:
Based on https://github.com/breez/breezmobile/issues/170 suggestion 

https://user-images.githubusercontent.com/1225438/122422829-d956fa00-cf63-11eb-9951-f5aed9faab09.mp4

### DIFF
--- a/lib/widgets/amount_form_field.dart
+++ b/lib/widgets/amount_form_field.dart
@@ -127,9 +127,23 @@ class SatAmountFormFieldFormatter extends TextInputFormatter {
       includeDisplayName: false,
       includeCurrencySymbol: false,
     );
+
+    var diff = formatted.length - oldValue.text.length;
+    var newOffset = newValue.selection.start;
+    if (formatted != oldValue.text) {
+      if (diff > 1) {
+        newOffset += 1;
+      }
+      if (diff < -1) {
+        newOffset -= 1;
+      }
+    } else {
+      newOffset = oldValue.selection.start;
+    }
+
     return newValue.copyWith(
       text: formatted,
-      selection: TextSelection.collapsed(offset: formatted.length),
+      selection: TextSelection.collapsed(offset: newOffset),
     );
   }
 }
--- a/lib/widgets/amount_form_field.dart
+++ b/lib/widgets/amount_form_field.dart
@@ -70,7 +70,7 @@ class AmountFormField extends TextFormField {
             controller: controller,
             inputFormatters: accountModel.currency != Currency.SAT
                 ? [FilteringTextInputFormatter.allow(RegExp(r'\d+\.?\d*'))]
-                : [FilteringTextInputFormatter.digitsOnly],
+                : [SatAmountFormFieldFormatter()],
             onFieldSubmitted: onFieldSubmitted,
             onSaved: onSaved,
             onChanged: onChanged,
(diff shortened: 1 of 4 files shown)


kingonly:
That's very very cool 👏👏👏

kingonly:
Did you test the BTC use case (format only sats)?

ademar111190:
> Did you test the BTC use case (format only sats)?

Yes but I did not record as I forced the mask to test it.

It kept the previous behaviour because of this https://github.com/breez/breezmobile/pull/500/files#diff-3a1f08529ce271b9478dd03b681a5905833fd1d90bbdc8400cbb1c3494a056baL71 check

lib/widgets/amount_form_field.dart
```diff
@@ -99,3 +99,37 @@ class AmountFormField extends TextFormField {
...
+    } catch (ignored) {
+      value = Int64(0);
+    }
+
+    final formatted = Currency.SAT.format(
+      value,
+      includeDisplayName: false,
+      includeCurrencySymbol: false,
+    );
+    return newValue.copyWith(
```
roeierez, line 130: As it is right now the cursor always jumps to the end of the text even if the user adds a digit in the middle.
I think it is better to preserve the selection and given the fact that a space can be added or removed I think something similar to the following suggestion may work:

```suggestion
    var diff = formatted.length - oldValue.text.length;
    var newOffset = newValue.selection.start;
    if (diff > 1) {
      newOffset += 1;
    }
    if (diff < -1) {
      newOffset -= 1;
    }
    return newValue.copyWith(
      text: formatted,
      selection: TextSelection.collapsed(offset: newOffset),
    );
```

roeierez (changes requested):
Very cool change, see my comment regarding preserving selection.

lib/widgets/amount_form_field.dart
ademar111190: Good catch, the code you suggested work like a charm, I had to add a small check only `if (formatted != oldValue.text)…` and use the old selection as else statement `newOffset = oldValue.selection.start;`. that is to avoid a wrong behaviour when you try to insert an invalid character.

ademar111190: btw How it looks like:

https://user-images.githubusercontent.com/1225438/122784577-f6077080-d288-11eb-9a0c-35640be3ef67.mp4

lib/widgets/amount_form_field.dart
```diff
@@ -70,7 +70,7 @@ class AmountFormField extends TextFormField {
             controller: controller,
             inputFormatters: accountModel.currency != Currency.SAT
                 ? [FilteringTextInputFormatter.allow(RegExp(r'\d+\.?\d*'))]
-                : [FilteringTextInputFormatter.digitsOnly],
+                : [SatAmountFormFieldFormatter()],
```
roeierez, line 73: Now that the formatter is changed the value contains spaces and when the validator tries to validate the form value here: https://github.com/breez/breezmobile/blob/5442a90e03426777acfee380bdf20e92e5303313/lib/widgets/amount_form_field.dart#L87
it will fail. You can reproduce that by pressing "CREATE" on the invoice page.
The failure is since the `currency.parse` expects only numbers.
It seems reasonable to me that we will change the `currency.parse` (for sats) to ignore spaces so the whole behaviour is symmetric.

roeierez (changes requested):
I believe it is my last pass.

lib/widgets/amount_form_field.dart
ademar111190: Don't worry, it is always better to fix things when they are fresh in mind, and you catch really good points as I have not test the rest of the flow 🤦 

Do you think is it ok applying the same regex `RegExp(r'[^\d*]')` on currency parse? So it removes the space and any other possible invalid thing, maybe a `sat` suffix.

roeierez: I tend to be more conservative here as the `parse` is used all over and there are many potential input sources...
Since we only format numbers and spaces it seems reasonable to me on parsing to ignore only spaces and leave the numbers.
Does it make sense to you? Or you have other valid cases in mind?

ademar111190: Makes sense @roeierez I'm applying the changes :)

ademar111190: Just made it, I add some tests to avoid future problems :)

roeierez (approved):

Issue #186, opened 2019-10-05
Bug: Android back button on Breez PIN window does nothing


bitcoinuser:
Bug: Android back button on Breez PIN window does nothing, but should close the app.

bitcoinuser:
Maybe show a message like this:

![photo5014805412993673263](https://user-images.githubusercontent.com/1178856/67611337-591cc900-f76f-11e9-830d-fda30db17cc8.jpg)

kingonly:
Fixed in latest

Issue #461, opened 2021-04-06
Clicking on back button when Breez menu is opened, should close it and not show exit popup


bitcoinuser:
Hi,

Clicking on back button when Breez menu is opened, should close it and not show exit popup.

![photo_2021-04-06_10-56-49](https://user-images.githubusercontent.com/1178856/113722666-061aa580-96c7-11eb-8dc0-1c37d1c68137.jpg)

Related to: https://github.com/breez/breezmobile/issues/448

bitcoinuser:
Fixed. Thanks

PR #502, opened 2021-06-17
Add close confirmation popup on lock screen

ademar111190:
I did apply the suggestion of this thread https://github.com/breez/breezmobile/issues/186

# Before 

https://user-images.githubusercontent.com/1225438/122470520-88132e80-cf94-11eb-9317-bc5afca4d7ee.mp4

# After

https://user-images.githubusercontent.com/1225438/122470545-8e090f80-cf94-11eb-8076-00823a811a79.mp4

### DIFF
--- a/lib/routes/security_pin/lock_screen.dart
+++ b/lib/routes/security_pin/lock_screen.dart
@@ -1,5 +1,6 @@
 import 'package:breez/bloc/user_profile/user_profile_bloc.dart';
 import 'package:breez/widgets/back_button.dart' as backBtn;
+import 'package:breez/widgets/close_popup.dart';
 import 'package:breez/widgets/pin_code_widget.dart';
 import 'package:flutter/material.dart';
 
@@ -30,9 +31,7 @@ class _AppLockScreenState extends State<AppLockScreen> {
   @override
   Widget build(BuildContext context) {
     return WillPopScope(
-      onWillPop: () {
-        return Future.value(widget.canCancel);
-      },
+      onWillPop: willPopCallback(context),
       child: Scaffold(
         appBar: widget.canCancel == true
             ? AppBar(
--- a/lib/routes/security_pin/lock_screen.dart
+++ b/lib/routes/security_pin/lock_screen.dart
@@ -31,7 +31,10 @@ class _AppLockScreenState extends State<AppLockScreen> {
   @override
   Widget build(BuildContext context) {
     return WillPopScope(
-      onWillPop: willPopCallback(context),
+      onWillPop: willPopCallback(
+        context,
+        canCancel: Future.value(widget.canCancel),
+      ),
       child: Scaffold(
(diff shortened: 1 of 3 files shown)


lib/routes/security_pin/lock_screen.dart
```diff
@@ -30,9 +31,7 @@ class _AppLockScreenState extends State<AppLockScreen> {
   @override
   Widget build(BuildContext context) {
     return WillPopScope(
-      onWillPop: () {
-        return Future.value(widget.canCancel);
-      },
+      onWillPop: willPopCallback(context),
```
roeierez, line 34: @ademar111190 we should return true if widget.canCancel is true as there is another flow that uses the lock screen and let the user cancel (go back).

roeierez (changes requested):
Good enhancement, see my comment bellow.

lib/routes/security_pin/lock_screen.dart
ademar111190: @roeierez I did add the `can cancel` clause to the `willPopCallback` function, I guess now it reflects correctly the behaviour.

lib/routes/security_pin/lock_screen.dart
```diff
@@ -30,9 +31,10 @@ class _AppLockScreenState extends State<AppLockScreen> {
   @override
   Widget build(BuildContext context) {
     return WillPopScope(
-      onWillPop: () {
-        return Future.value(widget.canCancel);
-      },
+      onWillPop: willPopCallback(
```
roeierez, line 34: Maybe it is simpler just to set onWillPop to `Future.value(true)` in case widget.canCancel is true? Then we don't need to propagate canCancel.

roeierez (changes requested):

lib/routes/security_pin/lock_screen.dart
ademar111190: I did add the function looking for this issue https://github.com/breez/breezmobile/issues/461 so I can write something like this:

```
canCancel: () {
  if (drawer.isOpen) {
    drawer.close();
    return false;
  } else {
    return true;
  }
```

So both cases will look similar and keep the close logic in the same place.

To get the same but with the check outside of the `willPopCallback` function, I will make something like that:

```
onWillPop: () {
  if (drawer.isOpen) {
    drawer.close();
    return Future.value(false);
  } else {
    return willPopCallback(context).call();
  }
```

To me it is ok despite of being more verbose.

So what do you think @roeierez ? Change the check to be outside the `willPopCallback` function or keep it there?

roeierez: I see, if we have two cases already then encapsulating that the way you did looks good to me.

roeierez (approved):

roeierez (approved):

PR #464, opened 2021-04-11
Going Global

mcroad:
This PR tracks the **on-going** effort to translate Breez. 

Breez has the potential to dominate peer-to-peer payments in developing markets where people are desperate for sound money. This PR addresses that need in two ways: adds an easy-to-contribute internationalization infraestructure, and adds Spanish as the first translation.

Translations are stored as plain JSON files in `assets/locales/{language}.json`. I'm using the [`flutter_i18n`](https://github.com/ilteoood/flutter_i18n) library. `flutter_i18n` is a simple wrapper around `flutter_localizations` and `intl` so I don't have to write all the boilerplate code necessary.

This is a **work in progress**. It is not ready to be merged. It requires more translating and thorough testing. Groundwork and configuration are done. I'll continue going through the app and extracting keys where necessary. 

I am not able to test in IOS. If you are able, please run this fork and let me know if everything works.

PRs to this PR and suggestions welcome.

Related to #430, #441, #209, #217

PS: I have not been able to run the app without commenting out [this line](https://github.com/mcroad/breezmobile/blob/19256afbb3c7080a696d63998a3d557d61f050e0/android/app/build.gradle#L46). Any idea why this is happening?

PR #503, opened 2021-06-18
Change payment timeline date format

ademar111190:
@kingonly shares an idea with me about changing the time format shown on the home payments timeline.

The idea is to use relative time notation such as "1 day ago" to recent date (3 days or less) or "month/day/year" to older dates instead of just "month/day".

How those changes looks like:

# Before

https://user-images.githubusercontent.com/1225438/122615129-60889880-d05e-11eb-8b4b-645804078bd0.mp4

# After

https://user-images.githubusercontent.com/1225438/122615137-64b4b600-d05e-11eb-8f9f-c56afb96e5c3.mp4

# PS

- The library that generates those messages is: https://pub.dev/packages/timeago
- The messages are customisable but for now it is using the defaults for english
- The library supports Spanish and other languages so it is compatible with https://github.com/breez/breezmobile/pull/464


roeierez (approved):
Thanks you!

PR #501, opened 2021-06-17
Add secure flag to main activity

ademar111190:
Based on https://github.com/breez/breezmobile/issues/184 suggestion

# Before

https://user-images.githubusercontent.com/1225438/122447037-b1729100-cf79-11eb-993f-f6662fe7300e.mp4

# After

https://user-images.githubusercontent.com/1225438/122447048-b46d8180-cf79-11eb-98eb-f6373817e16c.mp4|

PR #506, opened 2021-06-21
WIP: Change the flag secure lifecycle moment

ademar111190:
@kingonly @bitcoinuser may you try the flag secure from this branch?

I'm wondering if the reason why it is not working on your devices is related to a lifecycle issue.

The changes I made here make it impossible to take screenshots, so the idea is just to test the lifecycle theory as I think to block screenshots is not desirable.

context: https://github.com/breez/breezmobile/pull/501


kingonly:
I feel it's too much for this issue tbh...
Let's focus on more important stuff and circle back to it.

PR #509, opened 2021-06-22
Close the drawer when back is pressed instead of showing the close dialog

ademar111190:
The idea is to fix the issue https://github.com/breez/breezmobile/issues/461 some context https://github.com/breez/breezmobile/pull/502

When you touch the back button, if the drawer is open the app should close the drawer instead of showing the exit dialog, in another hand if the drawer is already closed it should show the close dialog.

P.S. I had to change the `canCancel` from `Future` to a `Function` to correct execute its body after the back action, it seems to not affect the lock screen.

How it looks like:

https://user-images.githubusercontent.com/1225438/122985838-072dab80-d375-11eb-9316-47b598d1a4c0.mp4


roeierez (approved):
LGTM

PR #512, opened 2021-06-23
Uses system navigator pop instead of calling exit directly

ademar111190:
I was wondering the reason why the app closes abruptly instead of the smooth close that apps usually do. I find out the reason is the direct call to `exit` instead of calling `SystemNavigator.pop`.

This PR changes the close method, I could not record the animation it does now because the frame rate of the screen recorder is not enough to capture the animation, but as you run it you can notice the difference.

# The test running

https://user-images.githubusercontent.com/1225438/123309180-fa3ac480-d4fa-11eb-9966-075f0fc8a708.mp4

### DIFF
--- a/lib/widgets/close_popup.dart
+++ b/lib/widgets/close_popup.dart
@@ -1,6 +1,5 @@
-import 'dart:io';
-
 import 'package:breez/widgets/error_dialog.dart';
+import 'package:flutter/services.dart';
 import 'package:flutter/widgets.dart';
 
 WillPopCallback willPopCallback(
@@ -13,7 +12,7 @@ WillPopCallback willPopCallback(
     if (canCancel != null && canCancel()) return true;
     return promptAreYouSure(context, title, Text(message)).then((shouldExit) {
       if (shouldExit) {
-        exit(0);
+        SystemNavigator.pop(animated: true);
       }
       return false;
     });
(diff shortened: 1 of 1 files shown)


roeierez:
@ademar111190 thanks for this PR and for pushing this clean approach.
I think there were historical reasons that origin in flutter early behaviour for this work around and I am ready to re-test.
One thing I would ask you to test is exiting while a podcast is playing. This is because the audio player is an external service and I would like to see it behaves well before merging this PR.

The test is:
1. Run breez and listen to a podcast.
2. Exit using back button and confirm in the dialog.
3. Run breez again (see that it starts up) and by that you make sure that it really exited perviously.
4. Wait a few seconds and see that LND is up and running as expected (just run `getinfo` command from the dev console).

Once this test passes I am OK to merge this.

ademar111190:
Hi @roeierez thank you for the feedback

I did the test, it seems to be working, I did add the record to the PR description.

lib/widgets/close_popup.dart
```diff
@@ -13,7 +12,7 @@ WillPopCallback willPopCallback(
     if (canCancel != null && canCancel()) return true;
     return promptAreYouSure(context, title, Text(message)).then((shouldExit) {
       if (shouldExit) {
-        exit(0);
+        SystemNavigator.pop(animated: true);
```
roeierez, line 15: Shouldn't we just return true here for the default exit behaviour?

roeierez (changes requested):

lib/widgets/close_popup.dart
ademar111190: You nailed it 😄

roeierez (approved):

PR #516, opened 2021-07-10
Add Breez as app_name for Podcast Value payments

AbelLykens:
This is very untested ;). But it will identify Breez as payment sender


kingonly:
@AbelLykens do you want to take a shot at PR? Should be an easy one...

AbelLykens:
This is a PR right? Or am I misunderstanding you? ;)

kingonly:
Yep, sorry... Read it from slack, though it's an issue.
Thank you 🙏

roeierez:
@AbelLykens sorry as I merged a PR that was implemented the same as this one: https://github.com/breez/breezmobile/pull/520

I should have take yours since you was the first but I only noticed this after the merge.
Many thanks for your effort.

AbelLykens:
That's okay ;)

Issue #519, opened 2021-07-13
add "app_name" to tlv record 7629169


daveajones:
It would be very helpful to have the "app_name" field present in the 7629169 TLV record being sent with podcast lightning payments.

AbelLykens:
think this can be closed

kingonly:
Fixed in latest

PR #520, opened 2021-07-13
add app_name to tlv record - addresses #519

daveajones:
I'm not sure if static strings are allowed here.  Since it's unlikely for the app name to change, maybe this is ok?


roeierez (approved):
LGTM, Thank you!

Issue #511, opened 2021-06-23
Breez fails to pick up lnurl-withdrawl from clipboard or when pasted into 


Liongrass:
REPRO:
1) I copy a lnurl
2) I open Breez 0.11 (4261.1) on iOS 14.6
3) See the notice "Breez pasted from ..."

EXPECTED:
Breez picks up the lnurl in the clipboard and opens the usual dialogue

ACTUAL:
No reaction from Breez

REPRO:
4) Go to 'Send' -> 'Paste Invoice or Node ID'
5) Paste the lnurl from the clipboard into the field
6) Click 'Approve'

EXPECTED
Breez picks up the lnurl in the field and opens the usual dialogue

ACTUAL:
Error message: "Invalid invoice or node ID"

Thank you for all your work!

kingonly:
Fixed in latest, package is rolling out today.

PR #521, opened 2021-07-14
Load lnurl from clipboard automatically

ademar111190:
Fix the issue https://github.com/breez/breezmobile/issues/511

In the following screen record I did: invalid clipboards, repeated clipboards and valid clipboard, it seems to be working.

I did not record but I did test the camera flow and it looks to be working, it is because I move the private `_handleLNUrl` to `lib/utils/lnurl.dart` as a public function to be reused by the home.

How it looks like:

https://user-images.githubusercontent.com/1225438/125650306-5b78fa8e-04cc-4aee-9804-e972c7a33dea.mp4

### DIFF
--- a/lib/bloc/lnurl/lnurl_bloc.dart
+++ b/lib/bloc/lnurl/lnurl_bloc.dart
@@ -3,6 +3,7 @@ import 'dart:async';
 import 'package:breez/services/breezlib/breez_bridge.dart';
 import 'package:breez/services/breezlib/data/rpc.pbserver.dart';
 import 'package:breez/services/injector.dart';
+import 'package:breez/utils/lnurl.dart';
 import 'package:breez/utils/retry.dart';
 import 'package:rxdart/rxdart.dart';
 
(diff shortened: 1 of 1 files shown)


roeierez:
@ademar111190 thank you for this another contribution.
LNURLBlock supposed to handle lnurl from various streams here: https://github.com/breez/breezmobile/blob/master/lib/bloc/lnurl/lnurl_bloc.dart#L43

Currently it handles click on links as input source and intuitively I would think the correct way is to add the clipboard source there too. Did you try it?
The advantage is that the LNURLHandler is listening to this merged stream and handles the UI flow.

ademar111190:
Hi @roeierez I think your suggestions makes total sense.

I did move and it seems to be working, but I'm a little bit worried about this part here: https://github.com/breez/breezmobile/pull/521/files#diff-360d3b233edce14f48f506a9f7a077b3aa5620a9f3e32bb43debc77398d72dd8L46-R51 may you can take an special look at this part 🙏

roeierez:
> Hi @roeierez I think your suggestions makes total sense.
> 
> I did move and it seems to be working, but I'm a little bit worried about this part here: https://github.com/breez/breezmobile/pull/521/files#diff-360d3b233edce14f48f506a9f7a077b3aa5620a9f3e32bb43debc77398d72dd8L46-R51 may you can take an special look at this part 🙏

@ademar111190 why are you worried? because the change of behaviour of how we detect lnurl (isLNURL function) ?

ademar111190:
> > Hi @roeierez I think your suggestions makes total sense.
> > I did move and it seems to be working, but I'm a little bit worried about this part here: https://github.com/breez/breezmobile/pull/521/files#diff-360d3b233edce14f48f506a9f7a077b3aa5620a9f3e32bb43debc77398d72dd8L46-R51 may you can take an special look at this part 🙏
> 
> @ademar111190 why are you worried? because the change of behaviour of how we detect lnurl (isLNURL function) ?

Exactly, I did remove the old check `startsWith…` and now it uses `isLNURL`, I know it works on the uses cases I know, but I'm afraid of some flow I do not know.

roeierez:
> Exactly, I did remove the old check `startsWith…` and now it uses `isLNURL`, I know it works on the uses cases I know, but I'm afraid of some flow I do not know.

The new function seems to cover the previous cases, worst case it will show a match for one that wasn't matched before. I suggest to go with this as it is cleaner to have one point of testing for lnurl.
Now that we have a simpler code can we revert the change that moved `_handleLNUrl` method as it is not needed to be exported anymore? We will end up with a small and elegant diff.

ademar111190:
> > Exactly, I did remove the old check `startsWith…` and now it uses `isLNURL`, I know it works on the uses cases I know, but I'm afraid of some flow I do not know.
> 
> The new function seems to cover the previous cases, worst case it will show a match for one that wasn't matched before. I suggest to go with this as it is cleaner to have one point of testing for lnurl.
> Now that we have a simpler code can we revert the change that moved `_handleLNUrl` method as it is not needed to be exported anymore? We will end up with a small and elegant diff.

Nice! I have not realised that change was not necessary anymore, good point!
I just remove that changes and everything seems to be working 🎉

roeierez (approved):
Looks Great!

Issue #522, opened 2021-07-15
iOS build fails with "No such module bindings"


NorrinRadd:
GoMobile produces xcframeworks now, instead of fat frameworks.
https://github.com/golang/mobile/pull/65

Xcode is not finding  the necessary static library with the new xcframework format.

Procedure
===
Build with iOS simulator target

Observed Behavior
===

Showing All Issues
CompileSwift normal x86_64 ~/code/breezmobile/ios/Plugins/Breez/BreezLib/iCloudBackupProvider.swift (in target 'Runner' from project 'Runner')
    cd ~/code/breezmobile/ios

~/code/breezmobile/ios/Plugins/Breez/BreezLib/iCloudBackupProvider.swift:11:8: error: no such module 'Bindings'
import Bindings
       ^

Expected Behavior
===
Successful iOS  build

kingonly:
Fixed

PR #523, opened 2021-07-15
Fix for Issue #522: "No such module bindings".

NorrinRadd:
This fix is dependent on https://github.com/breez/breez/pull/136


roeierez (approved):
LGTM!

PR #535, opened 2021-07-21
Break home page build method into several methods

ademar111190:
I have been working on some issues of the project on the last weeks and I had some problems figuring out what is happening on home page's build method because it was big and much of the code was too close to the right border making it crushed on the last columns of the text editor.

I did invest this morning breaking the method into some methods so we get the code closer to the left side and smaller methods making it easy to understand what is going on.

I tried to not change any code style or logic despite some changes were inevitable.

Please take a look and share your thoughts,  I don't want to make any radical changes on the project, I just believe this little refactoring will make much easier futures improvements on home page 🙏

### DIFF
--- a/lib/home_page.dart
+++ b/lib/home_page.dart
@@ -192,19 +192,20 @@ class HomeState extends State<Home> with WidgetsBindingObserver {
       child: StreamBuilder<BreezUserModel>(
           stream: widget.userProfileBloc.userStream,
           builder: (context, userSnapshot) {
-            var user = userSnapshot.data;
+            final user = userSnapshot.data;
 
             return StreamBuilder<AccountModel>(
                 stream: widget.accountBloc.accountStream,
                 builder: (context, accSnapshot) {
-                  var account = accSnapshot.data;
+                  final account = accSnapshot.data;
                   if (account == null) {
                     return SizedBox();
                   }
+
                   return StreamBuilder<AccountSettings>(
                       stream: widget.accountBloc.accountSettingsStream,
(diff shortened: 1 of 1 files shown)


lib/home_page.dart
```diff
@@ -192,318 +192,46 @@ class HomeState extends State<Home> with WidgetsBindingObserver {
       child: StreamBuilder<BreezUserModel>(
           stream: widget.userProfileBloc.userStream,
           builder: (context, userSnapshot) {
-            var user = userSnapshot.data;
```
ademar111190, line 195: I did change those `vars` to `finals` before moving code to make sure it was not being edited at some point, and I keep it because I think it makes senses once that code comes from the streams.

lib/home_page.dart
```diff
@@ -514,6 +242,349 @@ class HomeState extends State<Home> with WidgetsBindingObserver {
...
 
+  Widget _build(
+    BuildContext context,
+    BreezUserModel user,
+    AccountModel account,
+    AccountSettings settings,
+    LSPStatus lspStatus,
+    List<AddFundVendorModel> vendor,
+  ) {
+    final addFundsVendors = _drawerConfigAddFundsVendors(
```
ademar111190, line 253: It seems this drawer piece is not being used anymore, I kept it but I guess it can be removed.

roeierez:
@ademar111190 I like this change just from reading you description. Thank you for caring about making the code more readable. I will review that tomorrow thoroughly.

roeierez (approved):

PR #542, opened 2021-07-27
Add support for Kollider vendor

erdemyerebasmaz:
- LNMarketsWebViewPage is now removed and vendors that require lnUrl authentication now use LNURLWebViewPage*
 - _Reordered vendor theme values in theme data to match marketplace config._

*:   _LNURLWebViewPage is a copy of LNMarketsWebViewPage that takes endpointURI, responseID parameters and considers 201 status response code._


roeierez (approved):
LGTM

PR #554, opened 2021-08-13
Prefill amount in LNURL Pay if it's unambiguous

### LINKED ISSUE #539: [Feature request] Prefill LNURL Pay amount
I think it would be useful to automatically set the amount in LNURL payments if the min amount equals max amount. You're not supposed to enter anything else there anyway!
Screenshot_20210722-112937
For example, on https://chat.blixtwallet.com/ you can post a message for 10 sats exactly and you have to enter 10 in the amount field. Entering 9 and less or 11 and more doesn't work obviously so there's no reason to bother the user.

rkfg:
Fixes #539


rkfg:
Travis seems to be broken, `NET::ERR_CERT_AUTHORITY_INVALID`...

kingonly:
Thank you @rkfg!
We'll get it merged ASAP.

yaslama (approved):

Issue #524, opened 2021-07-15
[Podcast] Allow for boost messages


AbelLykens:
The TLV record supports messages with boosts. 

My dream: a longpress on boost creates a popup where I can set a one time boost amount, and enter a message.

Hope you can consider it.

GO PODCASTING!

kingonly:
Why set the amount on long press? Just a message seems simpler and more streamlined.

AbelLykens:
True. Was thinking that for messaged boosts, I might do one-time larger payments. So that's easier than +++++++ing before doing the boost, then -------'ing again ;)

kingonly:
You don't have to use the +/-. You can change the amount by clicking on it.

AbelLykens:
Okay skip that requirement then. TIL. Cheers.

MaxHillebrand:
Consider it a bump in the feature request :) 
Notice custom amount is discussed in #466.

kingonly:
Added boostagram support in the latest update

MaxHillebrand:
Y'all are rockstars! 
Thanks @kingonly

kingonly:
Last update caused a regression in the "regular" boost. We'll fix soon.

kingonly:
I updated the APK: https://github.com/breez/breezmobile/wiki/Can-I-download-Breez-APK%3F

Btw, we're adding an amount field to the Boostagram dialog to allow a bit much flexibility.

MaxHillebrand:
tested latest update, it works well! 

Yes, changing amount while writing the text would be useful.

PR #558, opened 2021-08-27
Add boost message on long pressing boost message

erdemyerebasmaz:
This PR addresses #524

Long pressing boost button will now open an input dialog for users to attach message to their boosts.


roeierez (approved):
Looks good

PR #560, opened 2021-08-30
Lightning-Address (sending)

nochiel:
This PR depends on https://github.com/breez/breez/pull/140 .
- [x] LNUrlBloc should parse e-mail addresses in case they are Lightning-Addresses.
- [x] Test Lightning-Addresses:
    - [x] zebedee.gg. 
    - [x] coinos.io
    - [x] lntxbot.com

### DIFF
--- a/lib/bloc/lnurl/lnurl_bloc.dart
+++ b/lib/bloc/lnurl/lnurl_bloc.dart
@@ -48,7 +48,7 @@ class LNUrlBloc with AsyncActionsHandler {
       ])
           .where((l) => l != null)
           .map((l) => l.toLowerCase())
-          .where((l) => isLNURL(l))
+          .where((l) => isLNURL(l) || isLightningAddress(l))
           .asyncMap((l) {
         _lnUrlStreamController.add(fetchLNUrlState.started);
         return _breezLib.fetchLNUrl(l).catchError((error) {
--- a/lib/bloc/lnurl/lnurl_model.dart
+++ b/lib/bloc/lnurl/lnurl_model.dart
@@ -38,6 +38,7 @@ class PayFetchResponse {
   PayFetchResponse(this.response);
 
   String get host => response.host;
+  String get lightningAddress => response.lightningAddress;
   String get callback => response.callback;
   Int64 get minAmount => response.minAmount;
   Int64 get maxAmount => response.maxAmount;
--- a/lib/routes/lnurl_fetch_invoice_page.dart
+++ b/lib/routes/lnurl_fetch_invoice_page.dart
@@ -96,18 +96,26 @@ class LNURLFetchInvoicePageState extends State<LNURLFetchInvoicePage> {
     LNUrlBloc lnurlBloc = AppBlocsProvider.of<LNUrlBloc>(context);
 
     /*
-        4. `LN WALLET` displays a payment dialog where user can specify an exact sum to be sent which would be bounded by:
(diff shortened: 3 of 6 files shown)


lib/bloc/lnurl/lnurl_bloc.dart
```diff
@@ -48,7 +48,7 @@ class LNUrlBloc with AsyncActionsHandler {
       ])
           .where((l) => l != null)
           .map((l) => l.toLowerCase())
-          .where((l) => isLNURL(l))
+          .where((l) => isLNURL(l) || isLightningAddress(l))
```
roeierez, line 51: Did you try constructing the lnurl here and just pass it to the go library as the 'https' prefixed url?
If it is a valid lnurl then the go library will handle it well as is without any code changes.

roeierez (changes requested):
Thanks for the PR Nick!

lib/bloc/lnurl/lnurl_bloc.dart
kingonly: @roeierez since we don't want business logic in breezmobile, do you mean calling a separate fetchLNUrlFromAddress that will be in the Breez lib?

roeierez: AFAICT the we can build from the email the lnurl to fetch by convention (host + suffix). If we use this convention here we can use the regular lnurl endpoint in breez and not change breez library at all.
No sure it is required to create an endpoint only for this convention.

kingonly: This is business logic that will be expanded in the (near) future. I think this "construction" should be in Breez as well.

roeierez: Currently we anyway detect email by `isLightningAddress` so we know it is a lightning address. My take is that for this small convention logic right now a dedicated endpoint, protobuf etc is overkill.
Nevertheless I also don't think it is wrong.

kingonly: isLightningAddress is a simple regex for UI validation. It doesn't (and shouldn't) contain the URL construction logic. We shouldn't implement the Lightning address spec in Breez mobile. It's simple right now, but it's going to be expanded soon.

nochiel: Btw. if we construct the lnurl here, we lose the email address that was used and won't be able to display it to the user on the payment screen. Should I still push that change?

kingonly: @nochiel "payment screen", do you mean "payment details dialog"? What's the title of the payment for these payments (Zebedee etc)?

nochiel: > @nochiel "payment screen", do you mean "payment details dialog"? What's the title of the payment for these payments (Zebedee etc)?

Yes. That's it. 

Currently the title is the host of the lnurl e.g. zebedee.io/coinois.io, etc.

kingonly: I see. So, yes, I think it's important we show the address either in the title or in the payment details.

nochiel: > I see. So, yes, I think it's important we show the address either in the title or in the payment details.

I'm in favour of leaving the PR as is except if a bug or something I overlooked is found.

roeierez (approved):
Looks good

nochiel:
> Looks good

Thanks. I'll squash and push shortly.

PR #562, opened 2021-09-03
Add sleep functionality

ademar111190:
Depends on: https://github.com/breez/anytime_podcast_player/pull/27

Sleep functionality is an option to automatically pause the podcast after x minutes.

It is accessible through an icon on the left side of playing controls, a little moon, and allows the user to select a time to automatically sleep on a dialog.

|No timer yet|Choose timer when no timer is selected yet|Choose timer when a timer is already selected|
|---|---|---|
|Screenshot_20210916-084003_Breez_🐛|Screenshot_20210916-084030_Breez_🐛|Screenshot_20210916-084042_Breez_🐛|
|Screenshot_20210916-084111_Breez_🐛|Screenshot_20210916-084118_Breez_🐛|Screenshot_20210916-084133_Breez_🐛|


lib/routes/podcast/sleep_selector.dart
```diff
@@ -0,0 +1,65 @@
+import 'package:anytime/bloc/podcast/audio_bloc.dart';
+import 'package:anytime/l10n/L.dart';
+import 'package:anytime/state/sleep_state.dart';
+import 'package:anytime/ui/widgets/sleep_selector.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter/widgets.dart';
+import 'package:flutter_svg/flutter_svg.dart';
+import 'package:provider/provider.dart';
+
+class BreezSleepSelector extends SleepSelector {
```
roeierez, line 10: Why do we need to override this class?

roeierez (changes requested):

lib/routes/podcast/sleep_selector.dart
ademar111190: Good question, there are two reasons:

- The icon on AnytimePlayer is a standard icon of the material, a crescent moon, and it is filled or not depending on the activated status of the sleep timer in the other hand here on Breez it is an SVG icon without any changes on its state.
- The second is the `dialogItemTileRadioFillColor`

If we manage to move the dialogItemTileRadioFillColor to the theme data, I can think in a way to pass the icon to the original widget so we can remove this overridden widget.

roeierez: Yes, I think what you suggest is a good approach.

roeierez: BTW even if overriding the theme doesn't work it is more compact and simple  to pass both icon and color via constructor instead of inheritance.

ademar111190: That is true, I'll change that 👍

ademar111190: ✅   Now the theme is properly set and the `SleepSelector` receives some properties to be customizable 🥳

roeierez (approved):

PR #563, opened 2021-09-03
Bug fixes and small changes to lightning-address functionality.

nochiel:
This PR depends on https://github.com/breez/breez/pull/141.
- [x] Simplify 'Send' input labels and help-text because we can paste other   types of ids that are supported e.g. Lightning-Addresses.
- [x] Small improvement to widget layout.
- [x] Show Lightning-Address in payment details dialog.
- ~Silent fail if we've set an LNUrlResponseError.~
    - This will be improved because Breez should present the user with an error if 
        - they manually paste a non-lightning-address. -
        - they paste a lightning-address but the host server generates an error.
- [x] Allow user to manually paste a lightning-address.

---

- [x] Improve error handling for lightning-addresses.
    - [x] Do not automatically process lightning-addresses from the clipboard. Require the user to manually paste.
    - [x] Show user an error if they manually paste a lightning-address and the host server generates an error.
        - [x] Handle when the server generates an lnurl json error.
        - [x] Handle when the server fails in a non-standard way e.g.
            - [x] [coinos.io](coinos.io) responds with the following on 20210904:
            ```
            $ curl -LR https://coinos.io/.well-known/lnurlp/chiron
            LNURL payments temporarily disabled due to a server issue. Sep. 2, 2021. Join our Telegram channel https://t.me/coinoswallet for updates
            ```
            - [x] [lntxbot.com](lntxbot.com)  responds with nothing on 20210904.
    - [x] Show user an error if they manually paste an non-lightning-address.

### DIFF
--- a/lib/bloc/lnurl/lnurl_bloc.dart
+++ b/lib/bloc/lnurl/lnurl_bloc.dart
@@ -1,5 +1,6 @@
 import 'dart:async';
 
+import 'package:breez/logger.dart';
 import 'package:breez/services/breezlib/breez_bridge.dart';
 import 'package:breez/services/breezlib/data/rpc.pbserver.dart';
 import 'package:breez/services/injector.dart';
@@ -37,6 +38,7 @@ class LNUrlBloc with AsyncActionsHandler {
   }
 
   listenLNUrl() {
+    log.info('listenLNUrl');
     if (_lnUrlStreamController == null) {
       _lnUrlStreamController = StreamController.broadcast();
       final injector = ServiceInjector();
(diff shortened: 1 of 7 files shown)


lib/bloc/lnurl/lnurl_bloc.dart
```diff
@@ -37,14 +38,19 @@ class LNUrlBloc with AsyncActionsHandler {
...
       Rx.merge([
         injector.nfc.receivedLnLinks(),
         injector.lightningLinks.linksNotifications,
         _lnurlInputController.stream,
-        injector.device.distinctClipboardStream,
+
+        // Process lightning-addresses in the clipboard only when the user manually 
+        // pastes them in and they show up in _lnurlInputController.stream.
+        injector.device.distinctClipboardStream
+            .takeWhile((l) => !isLightningAddress(l)),
```
roeierez, line 53: `takeWhile` would close the stream on first not matched event.
Instead I think we should use:

`injector.device.distinctClipboardStream.where(isLightningAddress)`

roeierez (changes requested):

PR #566, opened 2021-09-07
Add support to web links opening on external browser

ademar111190:
When the user scan a url link, the app offers the possibility to open it with the device's browser.

https://user-images.githubusercontent.com/1225438/132504564-281c99b3-114a-430e-99cb-2d43c6eb94db.mp4


roeierez (approved):
we should use `forceSafariVC:false` for iOS to open in an external browser.
I will fix after merging.

PR #568, opened 2021-09-13
Add custom amount to boostagram

ademar111190:
The main change of the PR is to add on boostagram dialog a form to set a custom amount.

The amount field is the same being used on other amount dialogs, to do so I had to extract the form to its own widget `CustomAmountFormField`.

Those changes also swap the tap and long-press behaviors, so now the tap opens the boostagram dialog and long-press applies the boostagram with the current value.

https://user-images.githubusercontent.com/1225438/133261138-b7ce6dbd-8d3d-46c3-b6c9-218c7b7e5056.mp4

### DIFF
--- a/lib/routes/podcast/boost.dart
+++ b/lib/routes/podcast/boost.dart
@@ -34,9 +34,6 @@ class _BoostWidgetState extends State<BoostWidget> {
     final userBloc = AppBlocsProvider.of<UserProfileBloc>(context);
     final accountBloc = AppBlocsProvider.of<AccountBloc>(context);
 
-    final selectedIndex = widget.userModel.paymentOptions.boostAmountList
-        .indexOf(widget.userModel.paymentOptions.preferredBoostValue);
-
     return StreamBuilder<AccountModel>(
         stream: accountBloc.accountStream,
         builder: (context, acc) {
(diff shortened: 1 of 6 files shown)


lib/routes/podcast/boost.dart
```diff
@@ -51,40 +48,24 @@ class _BoostWidgetState extends State<BoostWidget> {
                 width: 88,
                 child: GestureDetector(
                   behavior: HitTestBehavior.translucent,
-                  onLongPress: () => showDialog(
+                  onTap: () => showDialog(
                     useRootNavigator: true,
```
roeierez, line 52: Since we have views that must hid the app content such as the pin code we should set the `useRootNavigator` to false.

lib/routes/podcast/custom_amount_form.dart
```diff
@@ -0,0 +1,70 @@
...
+          return null;
+        },
+        style: theme.dialogTheme.contentTextStyle.copyWith(height: 1.0),
+        decoration: decoration,
+      ),
+    );
+  }
+}
+
+String initialCustomAmount(
```
roeierez, line 54: For this component to be more self contained maybe it is better to initialize the textEditingController here instead of calling this method from outside?
We can pass the presetAmounts and customAmount to the widget.

lib/routes/podcast/boost_message_dialog.dart
```diff
@@ -46,50 +59,68 @@ class BoostMessageDialogState extends State<BoostMessageDialog> {
...
+            style: Theme.of(context)
+                .dialogTheme
+                .contentTextStyle
+                .copyWith(height: 1.0),
+            decoration: InputDecoration(
+              labelText: "Boostagram (optional)",
+            ),
+          ),
+        ),
+        CustomAmountForm(
```
roeierez, line 87: Extracting out the CustomAmount to its own widget is a good decision. I think though it is better to make it a `FormField` instead of a form. This way it can be inlined inside a top level form as we have here ( same pattern as with `AmountFormField`).

roeierez (changes requested):
Overall looks good. Dropped some comments.

lib/routes/podcast/boost.dart
ademar111190: Interesting, to be honest, I didn't know what this flag was intended for, thanks for your comment and further research now I got it 🙏

lib/routes/podcast/custom_amount_form.dart
ademar111190: @roeierez I did try it but it stops the field to work, I'm still trying to understand the reason under the hoods of this but my guess, for now, is that it messes with form<->controller relationship.

I thought of an alternative way and I would like to know what you think about it, the alternative is to have a custom `TextEditingController` let's say `CustomAmountTextEditingController` and pass it for `CustomAmountForm` instead of a normal `TextEditingController`. that way we set the initial text logic inside the `CustomAmountTextEditingController`, what do you think?

roeierez (approved):

PR #572, opened 2021-09-17
Fix pin widget on small screens

ademar111190:
The pin widget was not showing some numbers on small screens, this PR fixes it.

|Before|After|
|---|---|
|before|after|

Other scenarios:

|small but no so small|error messsage|entire/big screen|unlock screen|
|---|---|---|---|
|small but no so small|error messsage|entire screen|unlock screen|


roeierez (approved):
Very Nice!

PR #581, opened 2021-09-27
Add debug suffix to allow release and debug builds on the same device

ademar111190:
It is useful to have debug installations without messing with the release app.

I'm also adding a check [here](https://github.com/breez/breezmobile/compare/master...ademar111190:debug-mode?expand=1#diff-9526ccfd1d1813ed49c39f8c54dbeb512607376a007d824b905bc8b4e4d202d9R45) to allow debug apps even without release info, super useful to developers exploring the project.

I did add icons with a green tint instead of blue to have a visual difference and make it easier to identify the debug build and a "debug" suffix on the name.

icons


roeierez (approved):
LGTM

PR #586, opened 2021-09-30
Show that withdrawal (reverse swap) is pending in payment details dialog.

nochiel:
Implements [BU-349](https://breeztech.atlassian.net/browse/BU-349).


lib/widgets/payment_details_dialog.dart
```diff
@@ -119,7 +119,7 @@ Future<Null> showPaymentDetailsDialog(
                     child: Scrollbar(
                       child: SingleChildScrollView(
                         child: AutoSizeText(
-                          paymentInfo.description,
+                          '${paymentInfo.description} ${paymentInfo.pending ? "(Waiting to be broadcasted)." : ""}',
```
roeierez, line 122: Won't this message show up also for pending lightning payment that is not a part of swap?

roeierez (changes requested):

lib/widgets/payment_details_dialog.dart
nochiel: Fixing. Thank you. I've made a note to test every payment type whenever I make such a change.

PR #588, opened 2021-10-05
Add fiat total to pdf print if there is only a single currency

ademar111190:
How the pdf looks like:

|multi currency|single currency|
|---|---|
|[multi-currency.pdf](https://github.com/breez/breezmobile/files/7285229/multi-currency.pdf)|[single-currency.pdf](https://github.com/breez/breezmobile/files/7285232/single-currency.pdf)|


lib/utils/print_pdf.dart
```diff
@@ -143,19 +143,42 @@ class PrintService {
   }
 
   _addTotalLineToTable(Map<String, ByteData> fontMap) {
-    double totalAmount = printParameters.submittedSale.totalChargeSat /
-        printParameters.currentCurrency.satConversionRate;
+    final sale = printParameters.submittedSale;
```
roeierez, line 146: Small one:
The PrintService class supposed to do as less calculation and business logic as possible and concentrate at printing.
It seems that moving out the logic of calculating the totals outside this class would be compatible with this approach. This can be done by adding these methods on the Sale class and use them from here:
1.  A method to calculate totalAmount by by currency
2. A method to get the single currency if exists

roeierez (changes requested):

lib/utils/print_pdf.dart
ademar111190: makes all sense, I'll change it, thank you @roeierez

lib/bloc/pos_catalog/model.dart
```diff
@@ -273,6 +273,26 @@ class Sale implements DBItem {
     });
     return total;
   }
+
+  Map<String, double> get totalAmountInFiat {
+    Map<String, double> totals = {};
+    saleLines.forEach((saleLine) {
+      final previous = totals[saleLine.currency] ?? 0.0;
+      final itemAdditional = saleLine.pricePerItem * saleLine.quantity;
```
roeierez, line 281: Maybe use here a function on SaleLine for calculating `totalFiat` ?

lib/bloc/pos_catalog/model.dart
```diff
@@ -273,6 +273,26 @@ class Sale implements DBItem {
...
+      totals[saleLine.currency] = previous + itemAdditional;
+    });
+    return totals;
+  }
+
+  double get totalAmountInSats {
+    var total = 0.0;
+    saleLines.forEach((saleLine) {
+      double totalPriceInFiat = saleLine.pricePerItem * saleLine.quantity;
+      double totalPriceInSats = saleLine.satConversionRate * totalPriceInFiat;
```
roeierez, line 291: Maybe use here a function on SaleLine for calculating `totalSats` ?

lib/routes/charge/sale_view.dart
```diff
@@ -502,7 +502,7 @@ class CurrencyDisplay extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     double priceInFiat = saleLine.pricePerItem * saleLine.quantity;
-    double priceInSats = currency.satConversionRate * priceInFiat;
+    double priceInSats = saleLine.satConversionRate * priceInFiat;
```
roeierez, line 505: Can use here `totalSats` of SaleLine

lib/utils/print_pdf.dart
```diff
@@ -120,8 +120,7 @@ class PrintService {
           saleLine.currency, printParameters.account);
       double priceInFiat = saleLine.pricePerItem;
       double totalPriceInFiat = saleLine.pricePerItem * saleLine.quantity;
-      double totalPriceInSats =
-          saleCurrency.satConversionRate * totalPriceInFiat;
+      double totalPriceInSats = saleLine.satConversionRate * totalPriceInFiat;
```
roeierez, line 123: Can use here `totalSats` of SaleLine as well

roeierez (changes requested):
Looks good! some small last comments and ready to merge.

ademar111190:
> Looks good! some small last comments and ready to merge.

Hi @roeierez , I just update the PR with the changes, thank you for the review 🙏

PR #606, opened 2021-10-21
Use uppercase characters when generating QR codes for more compact QR codes.

### LINKED ISSUE #595: [BD] Use uppercase lettering for invoices and addresses
Breez currently uses lower cased bitcoin addresses and Lightning invoices. As [outlined here](https://bitcoinops.org/en/bech32-sending-support/#creating-more-efficient-qr-codes-with-bech32-addresses) using uppercase lettering results in more efficient QR code generation making them easier to scan. 

> In summary, when using bech32 addresses in QR codes, consider uppercasing them and any other adjacent characters that can be uppercased in order to produce smaller and less complex QR codes. (However, for all other purposes, bech32 addresses should use all lowercase characters.)

nochiel:
Fixes #595.

### DIFF
--- a/lib/bloc/account/add_funds_bloc.dart
+++ b/lib/bloc/account/add_funds_bloc.dart
@@ -14,7 +14,6 @@ import "package:ini/ini.dart";
 import 'package:rxdart/rxdart.dart';
 
 import '../blocs_provider.dart';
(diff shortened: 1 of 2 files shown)


kingonly:
Did @roeierez review it?

Issue #598, opened 2021-10-17
[BD] Review fiat currency symbol placement


kingonly:
"Placing currency € before the number is an English way of displaying it. All European users except UK (and Turkey) place it after the number (and one space) (347.57 €)
https://en.wikipedia.org/wiki/Language_and_the_euro#Summary
Since this currency will be used by continental Europeans the most, especially from eurozone, it would be a nice change.
Same goes for zł (polish złoty) - it is always placed after the number, and it looks really weird for national polish to see it before. I think same case might be for other national currencies, but this needs more research. "

kingonly:
@erdemyerebasmaz please take a look

kingonly:
We will close when publish

kingonly:
We'll close when we publish

kingonly:
Published

PR #614, opened 2021-10-29
Add locale support for currency format

ademar111190:
|Default `x,xx €`|English `€x,xx`|Dutch `€ x,xx`|Turkish `€x,xx`|
|---|---|---|---|
|pt|en|nl|tr|

|Greek `x,xx €`| Cyprus's Greek `€x,xx`|German `x,xx €`|Austria's German `€ x,xx`|
|---|---|---|---|
|el|el-cy|de|de-AT|

I'm using this table as reference: https://en.wikipedia.org/wiki/Language_and_the_euro#Written_conventions_for_the_euro_in_the_languages_of_EU_member_states

Fixes https://github.com/breez/breezmobile/issues/598

P.S. I did not add `Irish` and `Maltese` overrides because the support on my device is incomplete so I could not test it, in theory copying the English overrides with the proper tag, `ga` and `mt` should work.

### DIFF
--- a/lib/l10n/locales.dart
+++ b/lib/l10n/locales.dart
@@ -16,7 +16,6 @@ Iterable<LocalizationsDelegate<dynamic>> localizationsDelegates() {
 Iterable<Locale> supportedLocales() {
   return [
     const Locale('en', ''),
-    const Locale('de', ''),
     const Locale('pt', ''),
   ];
 }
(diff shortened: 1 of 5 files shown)


lib/l10n/locales.dart
```diff
@@ -16,7 +16,6 @@ Iterable<LocalizationsDelegate<dynamic>> localizationsDelegates() {
 Iterable<Locale> supportedLocales() {
   return [
     const Locale('en', ''),
-    const Locale('de', ''),
```
ademar111190, line 19: I found a bug here, different from the anytime player, the breez still doesn't support 'de' locale.

roeierez (approved):

PR #618, opened 2021-11-01
Fix Payment Vendor Breez identification for sale items

ademar111190:
Notice the first item, it is not a sale, fixed the identification on this PR

|Before|After|
|---|---|
|Before|After|

### DIFF
--- a/lib/bloc/account/account_model.dart
+++ b/lib/bloc/account/account_model.dart
@@ -737,11 +737,18 @@ class SinglePaymentInfo implements PaymentInfo {
       return PaymentVendor.FAST_BITCOINS;
     }
 
-    if (type == PaymentType.RECEIVED && memo.payeeName != null) {
+    if (_isSalePayment()) {
       return PaymentVendor.BREEZ;
     }
 
     return null;
+  }
+
+  bool _isSalePayment() {
+    final payeeName = _paymentResponse.invoiceMemo.payeeName;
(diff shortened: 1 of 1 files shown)


lib/bloc/account/account_model.dart
```diff
@@ -737,7 +737,10 @@ class SinglePaymentInfo implements PaymentInfo {
       return PaymentVendor.FAST_BITCOINS;
     }
 
-    if (type == PaymentType.RECEIVED && memo.payeeName != null) {
+    final payeeName = memo.payeeName;
+    if (type == PaymentType.RECEIVED &&
+        payeeName != null &&
+        payeeName.isNotEmpty) {
```
roeierez, line 743: This heuristic works indeed. We could also fetch all sale items when building the payments array and inject the sale into the payments but this might be sufficient for now.
Could you please wrap this in a getter (isSalePayment) so later when this heuristic changes we will easily replace it? Also for this code to be more easily understood?

roeierez (changes requested):

lib/bloc/account/account_model.dart
ademar111190: Of course, I'll do it.
I was avoiding reading the sale items to not hurt the initial build time of the home.

ademar111190: ✅

roeierez (approved):

PR #619, opened 2021-11-02
Handle lightning-address URIs.

nochiel:
Fixes: breez/breezmobile/issues/616.

- [x] Handle `lightning:name@domain.com` uris correctly.
    - [x] Autodetect lightning-address URIs in the clipboard.
    - [x] Handle `lightning:name@domain.com` QR codes correctly.
    - [x] Handle pasting lightning-address URIs using the bottom_actions_bar widget.
    - [x] Handle pasting lightning-address URI into payee information dialog.
    - [x] Handle intent when the user clicks on a lightning-address link and Breez opens.
- [x] Test to make sure that all the usual paths for processing invoices and lnurls have not broken.
    - [x] Test pasting plain lightning-addresses.
    - [x] Test autodetection of lnurls in clipboard.
- [ ] Code review.
- [ ] Send PR upstream to [lighting-address spec](github.com/andrerfneves/lightning-address) to mention `lighting:` prefix for lightning-address URIs.

### DIFF
--- a/lib/bloc/invoice/invoice_bloc.dart
+++ b/lib/bloc/invoice/invoice_bloc.dart
@@ -156,6 +156,9 @@ class InvoiceBloc with AsyncActionsHandler {
         .map((s) {
           String lower = s.toLowerCase();
           if (lower.startsWith("lightning:")) {
+            if (isLightningAddress(lower)) {
+              return null;
+            }
             return s.substring(10);
           }
 
--- a/lib/bloc/invoice/invoice_bloc.dart
+++ b/lib/bloc/invoice/invoice_bloc.dart
@@ -85,7 +85,7 @@ class InvoiceBloc with AsyncActionsHandler {
         }
 
         if (isLightningAddress(normalized)) {
-          return DecodedClipboardData(clipboardData, "lightning-address");
+          return DecodedClipboardData(normalized, "lightning-address");
         }
 
         if (normalized.startsWith("ln")) {
--- a/lib/bloc/invoice/invoice_bloc.dart
+++ b/lib/bloc/invoice/invoice_bloc.dart
@@ -154,11 +154,12 @@ class InvoiceBloc with AsyncActionsHandler {
           s.toLowerCase().startsWith("lightning:"))
     ])
         .map((s) {
+          if (isLightningAddress(s)) {
+            return null;
+          }
+
           String lower = s.toLowerCase();
           if (lower.startsWith("lightning:")) {
-            if (isLightningAddress(lower)) {
-              return null;
-            }
             return s.substring(10);
           }
 
--- a/lib/bloc/lnurl/lnurl_bloc.dart
+++ b/lib/bloc/lnurl/lnurl_bloc.dart
@@ -47,15 +47,19 @@ class LNUrlBloc with AsyncActionsHandler {
         injector.lightningLinks.linksNotifications,
         _lnurlInputController.stream,
 
-        // Process lightning-addresses in the clipboard only when the user manually 
+        // Process plain lightning-addresses in the clipboard only when the user manually
         // pastes them in and they show up in _lnurlInputController.stream.
-        injector.device.distinctClipboardStream
-            .where((l) => !isLightningAddress(l)),
+        // Also process lightning-address URIs.
+        injector.device.distinctClipboardStream.where((l) =>
+            !isLightningAddress(l) || l.toLowerCase().startsWith('lightning:')),
       ])
           .where((l) => l != null)
           .map((l) => l.toLowerCase())
           .where((l) => isLNURL(l) || isLightningAddress(l))
           .asyncMap((l) {
+        if (l.toLowerCase().startsWith('lightning:')) {         // lightning-address URIs could come from arbitrary streams.
+          l = l.substring('lightning:'.length);
+        }
         _lnUrlStreamController.add(fetchLNUrlState.started);
         return _breezLib.fetchLNUrl(l).catchError((error) {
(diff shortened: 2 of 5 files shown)


lib/bloc/lnurl/lnurl_bloc.dart
```diff
@@ -47,15 +47,19 @@ class LNUrlBloc with AsyncActionsHandler {
...
         _lnurlInputController.stream,
 
-        // Process lightning-addresses in the clipboard only when the user manually 
+        // Process plain lightning-addresses in the clipboard only when the user manually
         // pastes them in and they show up in _lnurlInputController.stream.
-        injector.device.distinctClipboardStream
-            .where((l) => !isLightningAddress(l)),
+        // Also process lightning-address URIs.
+        injector.device.distinctClipboardStream.where((l) =>
+            !isLightningAddress(l) || l.toLowerCase().startsWith('lightning:')),
```
roeierez, line 54: should it be` !isLightningAddress(l) &&l.toLowerCase().startsWith('lightning:'))` ?
Or I am missing the condition here?
Previousely we filtered lightning addresses (emails) now it seems that we won't filter them if they prefixed with "lightning:"

lib/utils/lnurl.dart
```diff
@@ -20,11 +20,19 @@ bool isLNURL(String url) {
   return false;
 }
 
-bool isLightningAddress(String email) {
-  // This is just a normal e-mail address.
+// FIXME(nochiel) Maybe this should parse 'lightning:' uris and return the address instead of a boolean?
```
roeierez, line 23: I think it is a good idea to change the method to return the parsed address and null if there isn't a match.
It will save a lot of code above that always checks `startsWith('lightning:')`

roeierez (changes requested):

lib/bloc/lnurl/lnurl_bloc.dart
nochiel: I've clarified the test in the latest commit.

       
        injector.device.distinctClipboardStream.where((l) {
          var result = true;
          if (isLightningAddress(l)) {
            result = false;
            result = l.toLowerCase().startsWith('lightning:');
          }
          return result;
        })

lib/bloc/lnurl/lnurl_bloc.dart
```diff
@@ -46,18 +46,23 @@ class LNUrlBloc with AsyncActionsHandler {
...
         _lnurlInputController.stream,
-
-        // Process lightning-addresses in the clipboard only when the user manually 
-        // pastes them in and they show up in _lnurlInputController.stream.
-        injector.device.distinctClipboardStream
-            .where((l) => !isLightningAddress(l)),
+        injector.device.distinctClipboardStream.where((l) {
+          var result = true;
+          if (isLightningAddress(l)) {
+            result = false;
```
roeierez, line 52: This line is not needed.

lib/bloc/invoice/invoice_bloc.dart
```diff
@@ -156,6 +156,9 @@ class InvoiceBloc with AsyncActionsHandler {
         .map((s) {
           String lower = s.toLowerCase();
           if (lower.startsWith("lightning:")) {
```
roeierez, line 158: Checking for lower case repeats itself in several places. The reason is I guess to to test for prefix 'lightning:'.
Can we use parseLlightningAddress instead? Maybe with a boolean parameter 'prefixed' that enforce prefix or not?
It is better to encapsulate this condition.

roeierez (changes requested):

lib/bloc/lnurl/lnurl_bloc.dart
nochiel: The redundancy is just to document the intention that we should not process email addresses in the clipboard. In case more logic was ever added to or changed in that block, the initial intention would be clear.

lib/bloc/invoice/invoice_bloc.dart
nochiel: Will do.

lib/bloc/lnurl/lnurl_bloc.dart
roeierez: Regarding documentation Isn't it better to add a comment instead? I find it confusing to see the variable assigned with a value that is overridden a line after.

nochiel: > Regarding documentation Isn't it better to add a comment instead?

Okay. Thanks. I'll do that.

lib/bloc/invoice/invoice_bloc.dart
nochiel: > Checking for lower case repeats itself in several places. The reason is I guess to to test for prefix 'lightning:'. [...]
> It is better to encapsulate this condition.

To clarify: Do you want me to change all the places that parse/test for a `lightning:` URIs to use a single utility function, even for this case when we are testing invoices, not lightning-addresses?

nochiel: That is, if you'd like me to refactor all the places that check the `lightning:` prefix, even when they're not checking for lightning-addresses, is it okay if I do that in a different PR?

roeierez: I think now that you added the isLightningURI you already encapsulated the logic so no need to do more than that now.

nochiel: Okay. Thanks.

roeierez (approved):

PR #622, opened 2021-11-02
Add Spanish initial support

ademar111190:
Picking texts already added to l10n from the Going Global PR https://github.com/breez/breezmobile/pull/464 Spanish translation

# How it looks like

|Menu|Menu POS|Home (timeago message)|
|---|---|---|
|Screenshot_20211102-111158_Breez_Debug|Screenshot_20211102-111246_Breez_Debug|Screenshot_20211102-112131_Breez_Debug|


lib/utils/date.dart
```diff
@@ -23,7 +24,7 @@ class BreezDateUtils {
 
   static String formatTimelineRelative(DateTime d) {
     if (DateTime.now().subtract(Duration(days: 4)).isBefore(d)) {
-      return timeago.format(d, locale: Platform.localeName);
+      return timeago.format(d, locale: getSystemLocale().languageCode);
```
ademar111190, line 27: Using just the language code to avoid default "English" translations if the locale doesn't match language and code.

e.g:
es_MX should use `es` instead of `en`.
pt_AO should use `pt` instead of `en`.

ios/Runner/Info.plist
```diff
@@ -5,8 +5,8 @@
 	<key>CFBundleLocalizations</key>
 	<array>
 		<string>en</string>
+		<string>es</string>
 		<string>pt</string>
-		<string>de</string>
```
ademar111190, line 9: Forget to remove de translation from ios, related: https://github.com/breez/breezmobile/pull/614

Issue #624, opened 2021-11-10
TLV record field types incorrect


daveajones:
Some of the types on the JSON properties are incorrect (feedID and value_msat_total should be integers, not strings for example).  It might be worth taking a look through the TLV 7629169 spec just to make sure everything matches up:

https://github.com/satoshisstream/satoshis.stream/blob/main/TLV_registry.md#field-7629169

I'll look at doing a pull request also to help out.

kingonly:
Published

PR #625, opened 2021-11-10
Fix type errors in tlv json - #624

daveajones:
This will need review.  I'm not a Dart guy.

Two of the property types didn't match the spec defined at:

https://github.com/satoshisstream/satoshis.stream/blob/main/TLV_registry.md#field-7629169

### DIFF
--- a/lib/bloc/podcast_payments/podcast_payments_bloc.dart
+++ b/lib/bloc/podcast_payments/podcast_payments_bloc.dart
@@ -335,7 +335,7 @@ class PodcastPaymentsBloc with AsyncActionsHandler {
     tlv["episode"] = episode.title;
     tlv["action"] = boost ? "boost" : "stream";
     tlv["time"] = _formatDuration(position.position);
-    tlv["feedID"] = int.parse(_getPodcastIndexID(episode));
+    tlv["feedID"] = _getPodcastIndexID(episode);
     tlv["app_name"] = "Breez";
     tlv["value_msat_total"] = msatTotal;
     if (boost && boostMessage != null) tlv["message"] = boostMessage;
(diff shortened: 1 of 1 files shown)


lib/bloc/podcast_payments/podcast_payments_bloc.dart
```diff
@@ -335,9 +335,9 @@ class PodcastPaymentsBloc with AsyncActionsHandler {
     tlv["episode"] = episode.title;
     tlv["action"] = boost ? "boost" : "stream";
     tlv["time"] = _formatDuration(position.position);
-    tlv["feedID"] = _getPodcastIndexID(episode);
+    tlv["feedID"] = int.parse(_getPodcastIndexID(episode));
```
roeierez, line 338: There is a conversion inside the method `_getPodcastIndexID`. Maybe just omit it instead of the roundtrip?

roeierez (changes requested):
Thanks a lot! need a small fix and ready to go.

lib/bloc/podcast_payments/podcast_payments_bloc.dart
kingonly: @ademar111190 please fix according to @roeierez suggestion, but make sure it does create a regression (if there are other calls that expect a string)

ademar111190: Changes applied @kingonly @daveajones @roeierez 🎉

roeierez (changes requested):
I should have comment on this before... apologies.

lib/bloc/podcast_payments/podcast_payments_bloc.dart
```diff
@@ -359,12 +359,12 @@ class PodcastPaymentsBloc with AsyncActionsHandler {
     return "";
   }
 
-  String _getPodcastIndexID(Episode episode) {
+  int _getPodcastIndexID(Episode episode) {
     final metadata = episode?.metadata;
     if (metadata != null && metadata["feed"] != null) {
-      return metadata["feed"]["id"]?.toString();
+      return metadata["feed"]["id"] ?? 0;
```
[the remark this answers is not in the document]
roeierez: > So if the id is null the `id is int` evaluates as false?

Yes

roeierez (approved):

PR #648, opened 2021-12-07
Extract view switch component and make it dynamic

ademar111190:
The idea is to re-use this component in a new screen being developed, to do it I had not only to extract but I had to make the spacing dynamic to work properly given the length of texts. I had to make the icons optional too.

How it looks like in different languages:

|KeyPad|Items|
|---|---|
|after_en_1|after_en_2|
|after_pt_1|after_pt_2|
|after_es_1|after_es_2|
|after_fi_1|after_fi_2|

Thank you @tlindi for the Finnish translations.

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -121,5 +121,7 @@
       }
     }
   },
+  "pos_invoice_tab_keypad": "Keypad",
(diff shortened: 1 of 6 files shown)


roeierez (approved):
Looks Good.

PR #652, opened 2021-12-16
Add translations for default profile (animals and colors)

ademar111190:
Thanks to @tlindi for the Finnish translations

I did my best to keep the compatibility with previous versions of the app, but there is an issue I can't solve:

A user with a version of the app previous to this feature receives a transaction from a user using the app that already has this feature where the user uses a language different from English, in this case, the user with the older version of the app will not see the other user avatar, it will be a white circle only as you can see below:

|Sender (out of date app)|Receiver (update app, not english)|
|---|---|
|Receiver|Sender|


lib/bloc/user_profile/profile_animal.dart
```diff
@@ -0,0 +1,238 @@
...
+  OX,
+  PANDA,
+  PIG,
+  RABBIT,
+  SEAGULL,
+  SHEEP,
+  SNAKE,
+}
+
+ProfileAnimal profileAnimalFromName(String name, AppLocalizations texts) {
```
roeierez, line 38: Isn't is better to build the global map (with all locales) once as a static map and when an animal is searched just use it?
The advantage is that it will faster for searching and smaller code.

roeierez (changes requested):

lib/bloc/user_profile/profile_animal.dart
ademar111190: True, I did apply it :)

roeierez (approved):

PR #663, opened 2021-12-27
Add translation to fiat currency names

ademar111190:
This PR is a proposal on how to display currencies on fiat currencies screen showing its native name plus a localized name if necessary.

Below are some examples, for instance, the US Dollar is showing as `US Dollar` in English and `US Dollar (Dólar Americano)` in Portuguese, the rationale behind this is to show the currency native name so it is easier to its user find it and inside the parenthesis, the name localized by the language of the device.

|English|Portuguese|Spanish|
|---|---|---|
|en|pt|es|

### DIFF
--- a/lib/routes/fiat_currencies/fiat_currency_settings.dart
+++ b/lib/routes/fiat_currencies/fiat_currency_settings.dart
@@ -4,6 +4,7 @@ import 'package:breez/bloc/account/account_bloc.dart';
 import 'package:breez/bloc/account/account_model.dart';
 import 'package:breez/bloc/user_profile/user_actions.dart';
 import 'package:breez/bloc/user_profile/user_profile_bloc.dart';
+import 'package:breez/services/currency_data.dart';
 import 'package:breez/theme_data.dart' as theme;
 import 'package:breez/widgets/back_button.dart' as backBtn;
 import 'package:breez/widgets/flushbar.dart';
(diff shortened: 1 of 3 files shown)


kingonly:
@ademar111190 hmmm... This seems a bit too much tbh. Why not keep it as is and is Spanish, Portuguese, etc use the localized name (only for the languages we support)? With the acronym and the symbol adding the native name seems a bit much.

ademar111190:
@kingonly Do you mean to show `US Dollar` in English and `Dólar Americano` in Spanish? Or keep it as it is today showing the English name?

kingonly:
@ademar111190 yes,  US Dollar in English and Dólar Americano in Spanish (use the localized name in the description). I just don't think we need to show the native name. Makes sense?

ademar111190:
Makes sense @kingonly I think I overengineering a little bit this part haha.
I'll apply the changes

tlindi:
I wouldn't mind to see also currencies localized too. eg Japan Yen - is bit hard to read for finnish person as it has local Japanese "letter". And we have somewhat many tourists from Japan.

kingonly:
@tlindi Currencies will be localized, but there's no need to localize them *and* add the original language. That too much for the UI.

ademar111190:
Changes applied, thank you for the feedbacks <3

PR #671, opened 2021-12-31
Better identification of payment items

ademar111190:
This PR is a purpose on how to fix the misidentified sale items.

e.g., on the following videos, you can see a Sale (first item +206 sats) and a Connect to Pay transaction (third item +100 sats).

Before these changes, because of the heuristics, both are shown in the list as a Sale Item, after the changes only the real Sale item is shown.

A bad side effect of these changes is the load that sometimes some transactions will show, on the videos I leave, and I back to the Balance page to show some items being loaded.

Please let me know your thoughts about this approach and the code, Happy new year 🥳 

# Before

https://user-images.githubusercontent.com/1225438/147827241-e8e49a9e-04b5-40a6-b59d-3e20d258fb5e.mp4

# After

https://user-images.githubusercontent.com/1225438/147827263-23758db1-1675-4b58-b31a-c478b88e6b0f.mp4

### DIFF
--- a/lib/bloc/account/account_bloc.dart
+++ b/lib/bloc/account/account_bloc.dart
@@ -663,25 +663,29 @@ class AccountBloc {
     DateTime _firstDate;
     print("refreshing payments...");
 
+    final hashedSales = await _posRepository.fetchSalesPaymentHashes();
     return _breezLib.getPayments().then((payments) {
-      List<PaymentInfo> _paymentsList = payments.paymentsList.map((payment) {
-        var singlePaymentInfo =
-            SinglePaymentInfo(payment, _accountController.value);
-
-        return singlePaymentInfo;
-      }).toList();
+      List<PaymentInfo> _paymentsList = payments.paymentsList
+          .map((payment) => SinglePaymentInfo(
+                payment,
+                _accountController.value,
+                hashedSales.contains(payment.paymentHash),
+              ))
+          .toList();
 
       if (_paymentsList.length > 0) {
         _firstDate = DateTime.fromMillisecondsSinceEpoch(
-            _paymentsList.last.creationTimestamp.toInt() * 1000);
+          _paymentsList.last.creationTimestamp.toInt() * 1000,
+        );
       }
       print("refresh payments finished " +
           payments.paymentsList.length.toString());
       return PaymentsModel(
-          _paymentsList,
-          _filterPayments(_paymentsList),
-          _paymentFilterController.value,
-          _firstDate ?? DateTime(DateTime.now().year));
+        _paymentsList,
+        _filterPayments(_paymentsList),
+        _paymentFilterController.value,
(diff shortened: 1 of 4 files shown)


kingonly:
This seems like a bug we must fix, but how do I reproduce it? It's not something I've experienced.
I'll let @roeierez comment on the code and on the item loading fix.

roeierez (commented):
@ademar111190 thanks for catching this.
This solution introduces a significant amount of code so we need to make sure it worth it and is the right solution.
If I understand correctly the issue itself is the wrong heuristics we use to identify sales items which fails to identify the connect to pay items right?

It seems to me that introducing an abstraction layer on top of PaymentInfo for fetching data is  overkill for solving this issue.
Did you consider setting a boolean flag on PaymenetInfo (salesItem) for that purpose?
This flag can be set at the time we fetch payments from breez library (AccoutBloc) and if you are worried about performance we can construct one query that takes all current hashes (returned from breez items) and let the PosRepository return only those that are sales items.
Let me know what you think.

ademar111190:
@kingonly Make a "Connect to Pay", that should be enough to reproduce the bug.

@roeierez I do agree, the main problem is today there is no way to diff a Sale from other Breez transfers such as Connect to Pay without fetching the sales, adding this boolean flag for sure solves it, maybe an int type is better so we can have different transactions types (Sale, Connect to Pay, etc…)

roeierez:
@ademar111190 I think we don't need to store this flag at this point which enables us to be pretty flexible in the type decision (boolean vs int). One option is to enhance the paymentInfo with the SalesItem itself so we can access it from the UI without additional fetching.
It seems to me that `_filterPayments` in the AccountBloc is a good place to enhance the list of PaymentInfo with this data.

ademar111190:
@roeierez Applied the changes, it seems to be working and there is much less code 🥳

lib/bloc/account/account_bloc.dart
```diff
@@ -663,28 +663,36 @@ class AccountBloc {
...
+        _firstDate ?? DateTime(DateTime.now().year),
+      );
     });
   }
 
+  Future<PaymentInfo> _fetchPaymentInfo(Payment payment) async {
+    return SinglePaymentInfo(
+      payment,
+      _accountController.value,
+      await _posRepository.fetchSaleByPaymentHash(payment.paymentHash),
```
roeierez, line 692: This will hit the sql db on every payment. Instead we can query once for all sales items, create a hash table in memory and use it to fetch a sale by hash.

roeierez (changes requested):

lib/bloc/account/account_bloc.dart
ademar111190: Changes applied ;)

roeierez (approved):
Looks great!

PR #683, opened 2022-01-06
Add lightnite logo

ademar111190:
There is no logo afaik of [lightnite](https://lightnite.io/) on vendors, so adding this one fix the following bug

|Before|After|
|---|---|
|Screenshot_20220106-172945_Breez_Debug|Screenshot_20220106-173044_Breez_Debug|


kingonly:
Litenite is no longer in our apps.

ademar111190:
@kingonly I'm receiving litenite on the apps list as you can check on the screenshot, I think it is necessary to remove it from there, but I don't know where this data is coming from.

kingonly:
@ademar111190 your marketplace configuration isn't up to date. Get a new one from @yaslama or just ignore it...

PR #693, opened 2022-01-12
Reduce welcome animation size

ademar111190:
I did translate the walkthrough process and I learned how the initial animation is made.

It happens to be a sequence of images (png) so I did pass those images through the [ImageOptim](https://imageoptim.com/mac).

ImageOptim removes metadata and compresses images without losing quality by testing some algorithms and picking the best one.

This process saved 213 KB out of 1.3 MB (21.5%), I think it is a free improvement so I'm opening this PR 😄 

Screenshot 2022-01-12 at 12 19 42


roeierez (approved):
Very nice!

Issue #688, opened 2022-01-11
Refund to bitcoin address with QR scan causes error


tlindi:
Platform: Android
Version: latest
Way to reproduce:
On breez X - Open Receive Bitcoin with address
On breez Y - Open Refund and scan btc address QR from breez X
On breez Y - look to beging of address, it begings with "BITCOIN:"
On breez Y - click Redeem and you will get error about incorrect address

PR #695, opened 2022-01-12
Fix bitcoin uppercase coming from camera scanner

ademar111190:
Fixes https://github.com/breez/breezmobile/issues/688

For some reason the camera scanner is sending the address as uppercase, I did add a `.toLowerCase()` to fix it,

Fixed a translation text too, the action should be in present mode "TRANSMITIR" instead of past mode "TRANSMITIDO/TRANSMITIDA".

### DIFF
--- a/lib/l10n/app_es.arb
+++ b/lib/l10n/app_es.arb
@@ -952,7 +952,7 @@
   "get_refund_action_continue": "SEGUIR",
   "get_refund_transaction": "Transacción de Reembolso",
   "get_refund_failed": "falló",
-  "send_on_chain_broadcast": "TRANSMITIDO",
+  "send_on_chain_broadcast": "TRANSMITIR",
   "send_on_chain_btc_address": "Dirección BTC",
   "send_on_chain_scan_barcode": "Escanear código de barras",
   "send_on_chain_invalid_btc_address": "Por favor ingrese una dirección BTC válida",
--- a/lib/l10n/app_pt.arb
+++ b/lib/l10n/app_pt.arb
@@ -952,7 +952,7 @@
   "get_refund_action_continue": "CONTINUAR",
   "get_refund_transaction": "Transação de Reembolso",
   "get_refund_failed": "falhou",
-  "send_on_chain_broadcast": "TRANSMITIDA",
+  "send_on_chain_broadcast": "TRANSMITIR",
   "send_on_chain_btc_address": "Endereço BTC",
   "send_on_chain_scan_barcode": "Ler código de barras",
   "send_on_chain_invalid_btc_address": "Por favor insira um endereço BTC válido",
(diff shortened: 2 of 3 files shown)


kingonly:
Something doesn't make sense. Why does the scanner return upper case? It doesn't do it if I scan an url etc... And even if it's upper case, why is that a problem? Shouldn't we support upper case?

kingonly:
We want to generate upper case QRs for Lightning invoices, not sure why it affects scanning.
I'll let @roeierez chime in on the best course of action.

roeierez:
@ademar111190 I think the 

> I found the reason why it is uppercase, it is here https://github.com/breez/breezmobile/blob/master/lib/widgets/compact_qr_image.dart#L64

This is not the scanner it is the QR we show for lightning and bitcoin addresses right? It doesn't explain why the scanner return uppercase strings.

roeierez:
> For some reason the camera scanner is sending the address as uppercase, I did add a `.toLowerCase()` to fix it,

But this fix is not in the scanner, it is is in the QR image we show.

kingonly:
I don't think the scanner always returns upper case. It doesn't if I scan a URL for example.

kingonly:
Maybe the fix is that we need to support "BITCOIN:xxx" strings there.

ademar111190:
> @ademar111190 I think the
> 
> > I found the reason why it is uppercase, it is here https://github.com/breez/breezmobile/blob/master/lib/widgets/compact_qr_image.dart#L64
> 
> This is not the scanner it is the QR we show for lightning and bitcoin addresses right? It doesn't explain why the scanner return uppercase strings.

It is because I got this uppercase following the https://github.com/breez/breezmobile/issues/688

> Maybe the fix is that we need to support "BITCOIN:xxx" strings there.

I think it is better to convert to lowercase because it supports `BITCOIN:` and `bitcoin:` and all possible combinations such as `BiTCoiN:`…

kingonly:
We can convert to lower case, but we need to be careful and review all the cases to make sure it makes sense. Also, we should ensure that the generated QRs remain the same (upper case).

roeierez:
@ademar111190 don't you think the issue is here?: https://github.com/breez/breezmobile/blob/master/lib/services/breezlib/breez_bridge.dart#L667
Schema (bitcoin: prefix) validation should be case sensitive here according to bip21 (https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki)

ademar111190:
@roeierez excellent doc that solves all questions, the data is case sensitive and the schema is case insensitive, I'll change the code where you point out because it solves completely 💪

ademar111190:
I did apply the changes, here it is working, thank you 🙏  @roeierez

roeierez (approved):

tlindi:
Nice, thanks to all team :)

PR #696, opened 2022-01-13
Add finnish translations to "duration" and app_fi.arp + typo fix

tlindi:
Review @ademar111190 and comment, please?

### DIFF
--- a/lib/l10n/app_fi.arb
+++ b/lib/l10n/app_fi.arb
@@ -10,24 +10,24 @@
   "home_drawer_item_title_apps": "Sovellukset",
   "home_drawer_item_title_pos_settings": "Myyntipisteasetukset",
   "home_drawer_item_title_developers": "Sovelluskehittäjille",
-  "home_drawer_item_title_get_refund": "Get Refund",
-  "home_podcast_title": "Anytime Podcast Player",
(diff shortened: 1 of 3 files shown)


lib/l10n/app_fi.arb
```diff
@@ -10,24 +10,24 @@
   "home_drawer_item_title_apps": "Sovellukset",
   "home_drawer_item_title_pos_settings": "Myyntipisteasetukset",
   "home_drawer_item_title_developers": "Sovelluskehittäjille",
-  "home_drawer_item_title_get_refund": "Get Refund",
-  "home_podcast_title": "Anytime Podcast Player",
-  "home_podcast_no_subscriptions": "Use the Discover view to find and subscribe to your first podcast",
+  "home_drawer_item_title_get_refund": "Varojen lunastus",
+  "home_podcast_title": "Anytime Podcast -sovellus",
```
ademar111190, line 14: I'm not sure if the `Player` on `Anytime Podcast Player` should be translated. I did not translate it to pt/es as I thought it was a brand name. What do you think @kingonly should we keep the `Player` not translated?

kingonly: Keep Player as is

PR #707, opened 2022-01-21
Translate payment failed report dialog

ademar111190:


lib/l10n/app_pt.arb
```diff
@@ -1067,6 +1067,11 @@
   "payment_request_dialog_requesting": "Você esta solicitando um pagamento de:",
   "payment_request_dialog_action_cancel": "CANCELAR",
   "payment_request_dialog_action_approve": "APROVAR",
+  "payment_failed_report_dialog_title": "O Pagamento Falhou",
```
bitcoinuser, line 1070: Eu colocaria as iniciais de Pagamento  e falhou em minusculo

ademar111190: Alterei https://github.com/breez/breezmobile/pull/710

PR #737, opened 2022-01-26
Create a generic BuildContext extension and update deprecated values & methods

erdemyerebasmaz:
Migrating to null safety on an app this scale is a big task, but an app on this scale will likely benefit the most as null safety approach helps with prevention of errors, and results in smaller, faster binaries.

As a preperation for null-safety AppLocalizations is moved to LocalizedBuildContext.

AppLocalizations.of(context) returns a nullable object, yet on runtime it can never return a null value. We will be adding ! operator to denote that it will never return a null value after upgrading Dart sdk to >2.12.0. Instead of adding this operator to every instance we call AppLocalizations.of(context), we will call it via LocalizedBuildContext.l10n.

- Removed every `final texts = AppLocalizations.of(context);`
- Imported LocalizedBuildContext on classes with BuildContext that uses AppLocalizations
- AppLocalizations is now accessed via context.l10n

Extra:
- Removed unnecessary imports,
- Theme, Navigator, MediaQuery are now accessed directly instead of via unnecessary variables that don't add much to readability.


ademar111190:
That is amazing, I did not know something was possible on the dart.

Two questions:
- Is not better to have an extension for the theme too? Something like `context.theme` and maybe the other functions? Following the same approach with AppLocalizations?
- About the variable removal, is not better to call it just once and pay the call once keeping the result in a variable and reusing n times instead of calling it n times? I know those functions are O(1) but there are several checks inside those functions that are being paid n times without necessity as the result is the same.

erdemyerebasmaz:
> That is amazing, I did not know something was possible on the dart.
> 
> Two questions:
> 
> * Is not better to have an extension for the theme too? Something like `context.theme` and maybe the other functions? Following the same approach with AppLocalizations?
> * About the variable removal, is not better to call it just once and pay the call once keeping the result in a variable and reusing n times instead of calling it n times? I know those functions are O(1) but there are several checks inside those functions that are being paid n times without necessity as the result is the same.

1. `Theme`, `Navigator`, `MediaQuery` are all non-nullable, accessing them via `.of(context)` is safe. `AppLocalizations `is nullable by default and we will be casting nullable `AppLocalizations? `away to non-nullable `AppLocalizations`.
```
  static AppLocalizations? of(BuildContext context) {
    return Localizations.of<AppLocalizations>(context, AppLocalizations);
  }
```
2. Yes, I understand your concern and this change is temporary. I'm looking for a better approach that reduces overhead costs & boilerplate code. We're still declaring them under build() methods, I think moving them to didChangeDependencies() where possible is a better idea.

roeierez:
@erdemyerebasmaz very nice work here preparing the null safety changes and introducing clean pattern for LocalizedBuildContext.
Regarding theme  I think the build method is the right place to fetch it from the context and I agree with @ademar111190 that the pattern you introduced (extension)  may fit perfectly for theme as well.
I understand we don't need it for null safety but IMO it just looks better regardless.
BTW the same approach is used here: https://pub.dev/packages/build_context

ademar111190:
This build_context library is amazing, and the code is "just" several "alias" so it does not introduce any weird black-box magic. e.g. `Size get mediaQuerySize => MediaQuery.of(this).size;` to be used as `context.mediaQuerySize`.

IMHO If we are ok with the overhead I think this library (or our own extensions) in combination with the AppLocalizations's extension would make an excellent pattern :)

roeierez:
> IMHO If we are ok with the overhead I think this library (or our own extensions) in combination with the AppLocalizations's extension would make an excellent pattern :)

I am OK with that.

erdemyerebasmaz:
I'd like to avoid introducing dependencies where we can and [build_context](https://github.com/pedromassango/build_context/blob/master/lib/src/build_context_impl.dart) is no different than our approach.
I will create a generic BuildContext extension under utils folder that will have AppLocalizations's extension and a filtered version of build_context with properties used on our app.

erdemyerebasmaz:
The changes are pretty straightforward but this is a hard PR to review if you go file-by-file. I suggest we create an online build for testing.

audio_service: Update deprecated notificationClickEventStream with notificationClicked
webview_flutter: Update deprecated evaluateJavascript with runJavascript
flutter_web_browser: Update deprecated addDefaultShareMenuItem with shareState
image_picker: Update deprecated getImage with pickImage and change return type from PickedFile to XFile

Corrected a typo: Confetty -> Confetti
FetchInvoiceDialogForm: Use maxLengthEnforcement instead of maxLengthEnforced
Update FlatButton to TextButton

Theme Changes:  
- Remove references to buttonColor, accentTextTheme(All references are removed but this change is partial until accent color values are moved to ThemeData)
AppBarTheme
- Update deprecated brightness with systemOverlayStyle
- Update deprecated textTheme with titleTextStyle & toolbarTextStyle

Ran Reformat Code & Optimize Imports
[PREFER x.isNotEmpty to !x.isEmpty for Iterable and Map instances.](https://dart-lang.github.io/linter/lints/prefer_is_not_empty.html)
Ignore missing_enum_constant_in_switch for backupKeyType switch
[Avoid using braces in interpolation when not needed.](https://dart-lang.github.io/linter/lints/unnecessary_brace_in_string_interps.html)
Ignore missing_required_param for circular buttons
Ignore deprecated_member_use for generated files under services/breez_server/generated

Future changes:
 - Organize theme values to introduce colors we lost access via theme due to removal of buttonColor, accentColor
 - Rename import 'package:breez/theme_data.dart' as theme; for naming consistency
   - Currently, we declare context.theme as ThemeData themeData wherever theme_data.dart is imported as theme.  
 - Use Form extension

roeierez:
@erdemyerebasmaz All looks like great changes but impossible to review as it includes formatting changes.
Anyway let's stop here with the changes to this PR, merge it and on the future suggestions you wrote above let's do it with smaller PRs. 
Great job!

roeierez (approved):

erdemyerebasmaz:
@roeierez Hold on with the merge. I will run an extensive test today.

erdemyerebasmaz:
I crosschecked against live build and fixed regressions but we still need to test C2P flow.

PR #746, opened 2022-01-29
Fix warning box text cut

ademar111190:
Thank you @bitcoinuser for the error report 🙏 

How it looks like:

|Before|After|
|---|---|
|before|after|


tlindi:
Thanx same occured on iPhone SE.

roeierez (approved):

Issue #107, opened 2019-06-09
Breez node disconnected for a long time and users do not know what is happening.


bitcoinuser:
Hi, 

Sometimes Breez node is disconnected for long time, so users open Breez wallet and don't have idea why they can't make or receive payments. 
Suggestion: When node is disconnected, show some kind of message on Breez app that service is offline and will be back soon.

kingonly:
Great suggestion 👍

tlindi:
Would be really nice feature :)

Consider also to notify of custom channels and custom Bitcoin node setting.

Maybe show "disconnected" error on Exclamation mark on top right, similar to Reclaim notification?

PR #751, opened 2022-02-01
Add channel connection check

ademar111190:
Fixes https://github.com/breez/breezmobile/issues/107

How it looks like

https://user-images.githubusercontent.com/1225438/152033822-bf9c7436-bb38-451c-9c47-6218df19b729.mp4

### DIFF
--- a/lib/handlers/check_channel_connection_handler.dart
+++ b/lib/handlers/check_channel_connection_handler.dart
@@ -1,0 +1,84 @@
+import 'dart:async';
+
+import 'package:breez/bloc/account/account_bloc.dart';
+import 'package:breez/logger.dart';
+import 'package:breez/services/device.dart';
+import 'package:breez/services/injector.dart';
+import 'package:breez/theme_data.dart' as theme;
+import 'package:breez/widgets/flushbar.dart';
+import 'package:flushbar/flushbar.dart';
+import 'package:flutter/widgets.dart';
+import 'package:flutter_gen/gen_l10n/app_localizations.dart';
+
+class CheckChannelConnection {
+  static final _instance = CheckChannelConnection._internal();
+
+  factory CheckChannelConnection() => _instance;
+
+  CheckChannelConnection._internal();
+
+  Timer _notReadyTimer;
+  Flushbar _flushbar;
+  StreamSubscription<bool> _subscription;
+
+  void startListen(
+    BuildContext context,
+    AccountBloc accountBloc,
+  ) {
+    ServiceInjector().device.eventStream.distinct().listen((event) {
+      switch (event) {
+        case NotificationType.RESUME:
+          _subscription = accountBloc.accountStream
+              .map((acc) => acc.readyForPayments)
+              .distinct()
+              .listen((ready) {
+            if (ready) {
+              _readyForPayments();
+              _notReadyTimer?.cancel();
+            } else {
+              _notReadyTimer = Timer(Duration(seconds: 30), () {
+                _notReadyForPayments(context);
+              });
+            }
+          });
+          break;
+        case NotificationType.PAUSE:
+          _flushbar?.dismiss();
+          _flushbar = null;
+          _subscription?.cancel();
+          _notReadyTimer?.cancel();
+          break;
+      }
+    });
+  }
+
+  void _readyForPayments() {
+    log.info("Account is ready for payments");
+    _flushbar?.dismiss();
+    _flushbar = null;
+  }
+
+  void _notReadyForPayments(BuildContext context) {
+    log.info("Account is not ready for payments, Breez is offline");
+    if (_flushbar != null) {
+      return;
+    }
+
+    final texts = AppLocalizations.of(context);
+    _flushbar = showFlushbar(
+      context,
+      buttonText: texts.handler_channel_connection_close,
+      onDismiss: () {
+        _flushbar = null;
+        return true;
+      },
+      position: FlushbarPosition.TOP,
+      duration: Duration.zero,
+      messageWidget: Text(
+        texts.handler_channel_connection_message,
+        style: theme.snackBarStyle,
+        textAlign: TextAlign.start,
+      ),
+    );
+  }
+}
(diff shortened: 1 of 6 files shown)


tlindi:
Much better than failing payments or impossibility to create invoices :)

Will notification stay visible until connection issue is cleared? Will \!/ be shown also?

kingonly:
Yes, but you can manually close it as well. This is instead of a warning.

roeierez:
@ademar111190 one concern is how it will behave when the app gets out of paused state (going from background to foreground).
In many cases channels will transition to in-active and it will take some time (same as restart) for them to be active again.
I think product wise we should use the delay in this case as well otherwise the user will be prompted with this warning on every time it goes to foreground.

Maybe a better way is to start the delay from the time we get the "channel inactive" event and use it as a "grace" period before we prompt?

roeierez:
Thinking again I can see weird cases happening also in my suggestion above...
A simple approach I was thinking of is to start the delay every time the app is going to foreground (assuming it will take time for the channels to reactivate).

bitcoinuser:
> Fixes #107
> 
> How it looks like
> 
>  breez_is_offline.mp4

I think together with Breez is offlline, should show a link to an external breez page with the status of the service and to devs communicate with more messages, like: We are aware Breez is offline and we are working to restore the service.

tlindi:
Translate app_fi.arb
- "handler_channel_connection_message": "Breez is offline",
- "handler_channel_connection_close": "CLOSE",

+ "handler_channel_connection_message": "Breez on offline-tilassa",
+ "handler_channel_connection_close": "SULJE",

ademar111190:
@roeierez I got it, I'm not sure which approach is better, the problem is complex but I'll start by restarting the timer given the foreground/background events and we can evolve from it

ademar111190:
> @roeierez I got it, I'm not sure which approach is better, the problem is complex but I'll start by restarting the timer given the foreground/background events and we can evolve from it

@roeierez I did apply a second time too and a mechanic of resetting the timer each time the account is updated, this improves a lot the behavior here on my tests, it is not bulletproof, pretty hard heuristic.

lib/handlers/check_channel_connection_handler.dart
```diff
@@ -0,0 +1,93 @@
...
+  final _subscriptions = CompositeSubscription();
+  final _notReadySubscriptions = CompositeSubscription();
+
+  Flushbar _flushbar;
+
+  factory CheckChannelConnection() => _instance;
+
+  CheckChannelConnection._internal();
+
+  void startListen(
```
roeierez, line 23: How about something like the following code (didn't test it) to simplify the implementation?

```
  Timer notReadyTimer;
  
  accountBloc.accountStream.
  map((acc) => acc.readyForPayments).
  distinct().
  listen((ready) {
    if (ready) {
      // remove flushbar
      notReadyTimer?.cancel();
    } else {
      notReadyTimer = Timer(Duration(seconds: 30), (){
        // show not ready flushbar
      });
    }
  });
```

roeierez (changes requested):

lib/handlers/check_channel_connection_handler.dart
ademar111190: 🤔  how can I cancel these tasks?

roeierez: What do you mean by tasks?

ademar111190: The "remove/show flushbar", on my tests this cancel and start again improved a lot the flushbar behaviour

lib/handlers/check_channel_connection_handler.dart
```diff
@@ -0,0 +1,80 @@
...
+              notReadyTimer?.cancel();
+            } else {
+              notReadyTimer = Timer(Duration(seconds: 30), () {
+                _notReadyForPayments(context);
+              });
+            }
+          });
+          break;
+        case NotificationType.PAUSE:
+          notReadyTimer?.cancel();
```
roeierez, line 45: To me it looks almost complete. Only missing cancelling the subscription on pause

ademar111190: ✅  @roeierez

roeierez (changes requested):
@ademar111190 just two small comments and ready to go in.
Looks good!

lib/handlers/check_channel_connection_handler.dart
```diff
@@ -0,0 +1,82 @@
...
+              _notReadyTimer?.cancel();
+            } else {
+              _notReadyTimer = Timer(Duration(seconds: 30), () {
+                _notReadyForPayments(context);
+              });
+            }
+          });
+          break;
+        case NotificationType.PAUSE:
+          _subscription?.cancel();
```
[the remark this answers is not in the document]
ademar111190: Good question, I did not add before because I did assume no, but I'm not sure the correct answer

lib/handlers/check_channel_connection_handler.dart
```diff
@@ -0,0 +1,82 @@
...
+  Flushbar _flushbar;
+  StreamSubscription<bool> _subscription;
+
+  void startListen(
+    BuildContext context,
+    AccountBloc accountBloc,
+  ) {
+    ServiceInjector().device.eventStream.distinct().listen((event) {
+      switch (event) {
+        case NotificationType.RESUME:
```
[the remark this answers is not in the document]
ademar111190: I did run to background and backing, and it behaves correctly, but to be honest this kind of situation need more test from different devices IMHO because it can be biased to my device specific configuration.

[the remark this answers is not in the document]
tlindi: If I can Help anyhow just let me know how.

lib/handlers/check_channel_connection_handler.dart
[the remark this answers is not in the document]
roeierez: I think we need to remove, otherwise if from some reason the connection is restored in the background, upon resuming the flushbar will stay visible even though the channel is live.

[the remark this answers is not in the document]
ademar111190: ✅

roeierez (approved):

PR #776, opened 2022-02-11
Translate pos payment dialog

ademar111190:
How it looks like

|En|Es|Pt|
|---|---|---|
|en|es|pt|


tlindi:
Ademar said around 2021 Nov, it is huge and long task to translate. I assumed it'll take some 1/2 years or so.

I'm so happy we are here - Local language is nice when you are asking people to give they funds to your business.

Looking forward to see release too. Thanks @ademar111190 &  @kingonly

kingonly:
Can't thank you enough @tlindi 👏👏👏

PR #785, opened 2022-02-28
Add sender name to boostagram and increase message size to 128 characters

ademar111190:
Reference: https://github.com/lightning/blips/blob/master/blip-0010.md

### DIFF
--- a/lib/routes/podcast/payment_adjustment.dart
+++ b/lib/routes/podcast/payment_adjustment.dart
@@ -255,8 +255,11 @@ class PaymentAdjustmentState extends State<PaymentAdjustment> {
                           key: boostWidgetKey,
                           userModel: userModel,
                           onBoost: (int boostAmount, {String boostMessage}) {
-                            paymentsBloc.actionsSink.add(PayBoost(boostAmount,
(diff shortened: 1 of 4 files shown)


lib/routes/podcast/payment_adjustment.dart
```diff
@@ -255,8 +257,11 @@ class PaymentAdjustmentState extends State<PaymentAdjustment> {
                           key: boostWidgetKey,
                           userModel: userModel,
                           onBoost: (int boostAmount, {String boostMessage}) {
-                            paymentsBloc.actionsSink.add(PayBoost(boostAmount,
-                                boostMessage: boostMessage));
+                            paymentsBloc.actionsSink.add(PayBoost(
+                              boostAmount,
+                              boostMessage: boostMessage,
+                              senderName: _senderName,
```
roeierez, line 263: Since we already listen to the user stream can we take the send name from `userModel.name`  and remove the `_senderName` property?

roeierez (changes requested):
small comment, otherwise Looks great!

roeierez (approved):

PR #790, opened 2022-03-04
Update app_fi.arb Add Translate FI Payment Options

tlindi:

PR #792, opened 2022-03-07
Translate payment received alert FI

tlindi:


tlindi:
Merged onto https://github.com/breez/breezmobile/pull/790

PR #793, opened 2022-03-07
Add Ukrainian Translation (Initial / Partial)

tlindi:


tlindi:
Initial translation by tlindi with Google Translate.

Pending additional work from native Ukrainians.

ademar111190:
@tlindi We cannot accept Google Translate texts.
All translations need to be made by real persons.

PR #798, opened 2022-03-15
Improve offline handler

ademar111190:


lib/handlers/check_channel_connection_handler.dart
```diff
@@ -41,13 +41,16 @@ class CheckChannelConnection {
 
   void startSubscription(AccountBloc accountBloc, BuildContext context) {
     _subscription = accountBloc.accountStream
-        .map((acc) => acc.readyForPayments)
+        .map((acc) => _StartSubscriptionHelper(
+              acc.readyForPayments,
+              acc.connected,
+            ))
         .distinct()
```
roeierez, line 48: Since we use a class as the subscription event we need to implement the equality here.

roeierez (changes requested):

lib/handlers/check_channel_connection_handler.dart
ademar111190: I did apply the enum idea, I think it is better.

roeierez: It seems that the ConnectionStatus.OTHER is never used? so maybe we can use just a boolean in the map function.

ademar111190: @roeierez It is "used" mark two of the possible combination, check this table:

|readyForPayments|connected|status|action|

roeierez: @ademar111190  but when `readyForPayment` = true we don't care about the `connected`
and when `connected`=false we don't care about `readyForPayment`
So we have only two possibilities.
In other words we should show the offline flushbar only if connected=true and readyForPayment=false.
On all other possibilities we should remove the flushbar.
Do you see it like this as well?

ademar111190: 🤔  I think I misunderstood something in the process, but now it is clear.

the combinations we have are the following right?

|connected|readyForPayments|action|
|---|---|---|
|true|false|show snackbar|
|true|true|dismiss snackbar|
|false|true|dismiss snackbar|
|false|false|dismiss snackbar|

ademar111190: @roeierez I did update the rule, now it makes more sense

roeierez (approved):

PR #799, opened 2022-03-16
Add POS sales report

ademar111190:
How it looks like:

https://user-images.githubusercontent.com/1225438/158620112-0fc8de0f-7b77-4860-93be-6b54829ec35c.mp4


tlindi:
+1 Quite nice, simple and clever :-)

lib/bloc/pos_catalog/bloc.dart
```diff
@@ -276,8 +293,130 @@ class PosCatalogBloc with AsyncActionsHandler {
     _loadItems();
   }
 
+  void _loadSelectedReportTimeRange() async {
```
roeierez, line 296: All serialization/deserialization to preferences is implemented using a lot of code right now.
Don't you think it would be better to encapsulate it inside PosReportTimeRange?
The standard way we use to serialize models to preferences is toJson/fromJson methods and I think it can be implemented here in _loadSelectedReportTimeRange/_updatePosReportTimeRange methods

lib/bloc/pos_catalog/bloc.dart
```diff
@@ -276,8 +293,130 @@ class PosCatalogBloc with AsyncActionsHandler {
...
+  void _loadSelectedReportResult() async {
+    _posReportResult.add(PosReportResult.empty());
+  }
+
+  Future _fetchPosReport(
+    FetchPosReport action,
+  ) async {
+    _posReportResult.add(PosReportResult.load());
+    final timeRange = action.range;
+    final payments = await _breezLib.getPayments();
```
roeierez, line 370: Why don't we query the db for sales items directly? Is there a reason we go through the payments list?

roeierez (changes requested):

lib/bloc/pos_catalog/bloc.dart
ademar111190: The problem is the time range, I did not find a way to fetch the sales by date.
If you see a way to fetch it using this time range filter it would be much better than reading the payment first.

roeierez: Indeed you are right. The sale_payments table should have a payment_time column...
Maybe it makes sense to add a migration step to add this time to all existing payments? And then change the implementation to use pure SQL query?

ademar111190: @kingonly The CSV uses the payment list too.

@roeierez I'm checking this migration thing

roeierez: @ademar111190  great!
In our sqflite plugin we can pass onUpgrade parameter to the openDatabase function for migration to run.

lib/bloc/pos_catalog/bloc.dart
ademar111190: Changed 🎖️

lib/bloc/pos_catalog/bloc.dart
ademar111190: @roeierez I did add a migration process to SQL, please take a double care on this part because it is hard to test and I'm afraid of something wrong that I did not catch 🙏

lib/bloc/pos_catalog/sqlite/repository.dart
```diff
@@ -198,4 +198,36 @@ class SqliteRepository implements Repository {
...
+      whereArgs: [
+        start.millisecondsSinceEpoch,
+        end.millisecondsSinceEpoch,
+      ],
+    );
+    final result = Set<Sale>();
+    for (final sale in sales){
+      final baseSale = Sale.fromMap(sale);
+
+      final saleLines = await _fetchDBItems(
```
roeierez, line 217: So here we are going over all sales and fetching their sale lines one by one which is not an effective use of sqlite.
Instead let's see if we can let the sql lite engine do the work by querying sale lines table directly for what we need as the final result, aggregating the total sats and fiat values.
For example something like this:

`Select sum( (quantity * price_per_item)) as fiat_value, sum( (quantity * price_per_item * sat_conversion_rate) ) as sat_value, count(sale_id) as total_sales from sale_lines left join sales on sale_lines.sale_id = sales.id where (sales.date < ? and sales.date > ?)`

We can create a model for this report item and get it directly in the UI.

roeierez (changes requested):
Looks a lot better, migration looks great.
See my comment regarding report query.

lib/bloc/pos_catalog/sqlite/repository.dart
ademar111190: I understood now, I had to make some changes:

```sql
SELECT
  sum((quantity * price_per_item)) as fiat_value,
  sum((quantity * price_per_item * sat_conversion_rate)) as sat_value,
  count(sale_id) as total_sales
FROM sale_line
LEFT JOIN sale ON sale_line.sale_id = sale.id
WHERE (sale.date >= ? and sale.date <= ?);
```

But there is a problem, It is not checking the fiat currency type. i.e. it sums all together sales made in USD, MXN, EUR, etc…

The Sale model has the correct logic:

```dart
  Map<String, double> get totalAmountInFiat {
    Map<String, double> totals = {};
    saleLines.forEach((saleLine) {
      final previous = totals[saleLine.currency] ?? 0.0;
      totals[saleLine.currency] = previous + saleLine.totalFiat;
    });
    return totals;
  }

  double get totalAmountInSats {
    var total = 0.0;
    saleLines.forEach((saleLine) {
      total += saleLine.totalSats;
    });
    return total;
  }
```
I'm not sure if it is possible to replicate this logic in pure SQL but even if it is possible I don't know if it is a good idea.

roeierez: I think adding group by currency to the SQL query will do the job.
IMO imementing this in SQL is the right thing. It uses SQLite engine, code more compact and far better performance.

ademar111190: OK, I'll study this SQL group thing, I never did that.

kingonly: If you group by fiat currency, remember we need to display the sats amount of all the groups (including an empty group) and the fiat amount only if the group by returns one group. I haven't done sql in a while though...

ademar111190: I did add the GROUP BY but I had to add a complementary code too, I'm not literate on SQL, if you have some idea on how to improve it please share it.

https://github.com/breez/breezmobile/pull/799/files#diff-9ef5e59fe1bc906213338b91d9ddd81cd09fe7ef8d5233f6004cc19173f08a4dR203

lib/bloc/pos_catalog/sqlite/repository.dart
```diff
@@ -198,4 +198,48 @@ class SqliteRepository implements Repository {
       ),
     );
   }
+
+  @override
+  Future<PosReportResult> fetchSalesReport(DateTime start, DateTime end) async {
+    final db = await getDB();
+    final report = await db.rawQuery(
+      """
+      SELECT
```
roeierez, line 207: @ademar111190 I think that the requirement is:
1. for sats - to show always the sum of all sales
2. for fiat - to show sum of fiat only if ALL sale lines in the period has the same currency.

So I think this sql should do it:

```
SELECT
        sum((quantity * price_per_item)) as fiat_value,
        sum((quantity * price_per_item * sat_conversion_rate)) as sat_value,
        count(distinct(sale_id)) as total_sales,
        currency
      FROM sale_line
      LEFT JOIN sale ON sale_line.sale_id = sale.id
      WHERE (sale.date >= ? and sale.date <= ?)
      GROUP BY currency
      limit 2;
```
      
Note that it is based on your query with the following changes:
1. without grouping by sale_id (why did you need that?) 
2. added count for sales
3. added limit to optimize a bit.

If the result set contains only one line then we should show the fiat value, if we get more than one line than we know there are multiple currencies in use and we show only sats in the report.
Does it make sense?

roeierez (changes requested):

lib/bloc/pos_catalog/sqlite/repository.dart
ademar111190: It seems to be working 😄

roeierez (approved):
Looks good.

PR #811, opened 2022-03-25
Update app_fi.arb Translate QR Dialog FI

tlindi:


lib/l10n/app_fi.arb
```diff
@@ -1503,33 +1503,33 @@
...
-  "qr_code_dialog_warning_message_with_lsp": "A setup fee of {setupFee} ({fiatFee}) is applied to this invoice. Keep Breez open until the payment is completed.",
+  "qr_code_dialog_synchronizing": "Sovellus synkronoi tietoja",
+  "qr_code_dialog_invoice": "Lasku",
+  "qr_code_dialog_action_close": "SULJE",
+  "qr_code_dialog_share": "Jaa laskun tiedot",
+  "qr_code_dialog_copy": "Kopioi laskun tiedot",
+  "qr_code_dialog_copied": "Laskun tiedot kopioitiin leikepöydälle.",
+  "qr_code_dialog_warning_message_error": "Laskun luonti epäonnistui",
+  "qr_code_dialog_warning_message": "Älä sulje sovellusta ennenkuin tapahtuma on valmis.",
+  "qr_code_dialog_warning_message_with_lsp": "Tämän laskun suorituksesta veloitetaan saapuvan kanavan perustamiskulu {setupFee} eli ({fiat}). Älä sulje sovellusta ennenkuin tapahtuma valmistuu.",
```
ademar111190, line 1514: You replaced the `fiatFee` key by `fiat`m it is necessary to use the `fiatFee` key.

ademar111190 (changes requested):

lib/l10n/app_fi.arb
tlindi: Fixed

PR #846, opened 2022-03-30
Add fiat value to payment item

ademar111190:
How it looks like:

https://user-images.githubusercontent.com/1225438/160919998-c127dac2-cc63-4fb9-8725-b084606d16da.mp4

### DIFF
--- a/lib/bloc/pos_catalog/sqlite/repository.dart
+++ b/lib/bloc/pos_catalog/sqlite/repository.dart
@@ -211,14 +211,27 @@ class SqliteRepository implements Repository {
   }
 
   @override
-  Future<Set<String>> fetchSalesPaymentHashes() async {
-    return Set.from(
-      await _fetchDBItems(
-        await getDB(),
-        "sale_payments",
-        (e) => e["payment_hash"],
-      ),
+  Future<Map<String, SaleSummary>> fetchSaleSummaryByPaymentHashes() async {
+    Map<String, SaleSummary> hashedSales = {};
+    final db = await getDB();
+    final lines = await db.rawQuery(
+      """
+      SELECT
+        sale_payments.payment_hash as payment_hash,
+        sum(quantity * price_per_item) as fiat_value,
+        sum(quantity * price_per_item * sat_conversion_rate) as sat_value,
+        group_concat(distinct(currency)) as currencies
+      FROM sale_line
+      LEFT JOIN sale ON sale_line.sale_id = sale.id
+      LEFT JOIN sale_payments ON sale_line.sale_id = sale_payments.id
+      WHERE sale_payments.paid_date NOT NULL
+      GROUP BY sale.id, payment_hash;
+      """,
     );
+    lines.map((e) => SaleSummary.fromMap(e)).forEach((e) {
+      hashedSales[e.paymentHash] = e;
+    });
+    return hashedSales;
   }
 
   @override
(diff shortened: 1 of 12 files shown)


lib/routes/transactions/pos_payment_item.dart
```diff
@@ -42,7 +47,14 @@ class PosPaymentItem extends StatelessWidget {
               mainAxisAlignment: MainAxisAlignment.start,
               mainAxisSize: MainAxisSize.min,
               children: [
-                _valueText(context),
+                Column(
```
roeierez, line 50: It is better to wrap this component in a stream builder of account and take the account from there rather than push the account model to all payment items as the currency changes independently from the payment items.

lib/bloc/pos_catalog/bloc.dart
```diff
@@ -264,6 +265,15 @@ class PosCatalogBloc with AsyncActionsHandler {
     }
   }
 
+  Future _fetchSales(FetchSales action) async {
+    final result = Map<String, Sale>();
+    final futures = action.paymentHashes.map((hash) => _repository
+        .fetchSaleByPaymentHash(hash)
```
roeierez, line 271: Here we query for each item while it is better to fetch all sales we need at once.
Even better maybe populate the sale object on the payment item at the account bloc here: https://github.com/breez/breezmobile/blob/master/lib/bloc/account/account_bloc.dart#L671
Right now we only fetch the hashes and we can easily change it to fetch other sale properties (without the sale lines of course).

roeierez (changes requested):

lib/routes/transactions/pos_payment_item.dart
roeierez: @ademar111190 I agree with you that if that blinks it is not better.
I am OK to leave it as is.

lib/bloc/pos_catalog/bloc.dart
ademar111190: I did apply the changes, looks better now IMHO.
One thing is it uses the `fetchSaleByID` and it loads the sale lines and it is necessary to load because it is used through the `totalAmountInFiat`.

roeierez (changes requested):

lib/bloc/pos_catalog/bloc.dart
ademar111190: Updated to use the new model `SaleSummary`

lib/bloc/pos_catalog/sqlite/repository.dart
```diff
@@ -211,14 +211,22 @@ class SqliteRepository implements Repository {
...
-      ),
+  Future<Map<String, Sale>> fetchSalesByPaymentHashes() async {
+    Map<String, Future<Sale>> futures = {};
+    await _fetchDBItems(
+      await getDB(),
+      "sale_payments",
+      (e) {
+        final hash = e["payment_hash"];
+        final saleID = e["sale_id"];
+        futures[hash] = fetchSaleByID(saleID);
```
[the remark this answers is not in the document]
ademar111190: Updated to use the new model `SaleSummary`

lib/bloc/pos_catalog/sqlite/repository.dart
```diff
@@ -211,14 +211,27 @@ class SqliteRepository implements Repository {
...
 
   @override
-  Future<Set<String>> fetchSalesPaymentHashes() async {
-    return Set.from(
-      await _fetchDBItems(
-        await getDB(),
-        "sale_payments",
-        (e) => e["payment_hash"],
-      ),
+  Future<Map<String, SaleSummary>> fetchSaleSummaryByPaymentHashes() async {
```
roeierez, line 214: Nice!

roeierez (approved):
Looks good!

PR #850, opened 2022-04-04
Translate custom pos item name

ademar111190:
As we talked on Telegram @kingonly @tlindi

### DIFF
--- a/lib/bloc/pos_catalog/model.dart
+++ b/lib/bloc/pos_catalog/model.dart
@@ -1,5 +1,7 @@
 import 'dart:convert';
 
+import 'package:breez/utils/locale.dart';
+
 abstract class DBItem {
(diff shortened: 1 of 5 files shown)


tlindi:
Does this enable also to select Custom icon?

kingonly:
"It" isn't an icon, it's dynamically derived from the first two letters of the item.

tlindi:
:-( Would it be possible to enable the customization of it too?

eg When PoS uses keypad one could define icon as a cash register or keypad, or what ever is their most commonly sold product.

kingonly:
Not planning to implement something like that. Your feedback is noted and if we'll get similar feedback in the future we'll reconsider, but the cost/value of implementing something like that seems to be high atm.

tlindi:
I see. Works for me.

PR #886, opened 2022-04-17
Custom and Flexible Playback Speed (Podcast), and other changes

Sahilsingh0808:
1. Custom Playback speed for the podcast by adding a custom input for speed.
2. Also made the speed slider more flexible by adding more divisions in it (break of 0.25 changed to 0.05)
3. Flutter packages updated in pubsec.yaml
4. Worked on size issue for different devices by using MediaQuery.of(context).size by working with dynamic sizes according to device's size (currently worked on the podcast).


kingonly:
Thanks for your contribution!
It would be better to break it down to multiple PRs, one per feature. Also, each commit should describe the specific change in the code. It would make it much easier to review and merge.

Sahilsingh0808:
I tried breaking the pr into multiple ones, but it was not allowed as it
said that i was a first time contributor.

Thanks and Regards
Sahil Singh

kingonly:
So I would suggest you close this one and submit a PR for one feature with a "clean" commit and we'll take it from there.

kingonly:
Also, if there's a UI change l, attaching before/after videos or pictures really helps.

Sahilsingh0808:
Ok, thanks for the suggestion. I will do that.

PR #887, opened 2022-04-19
Playback Speed made more flexible and option for Custom speed input

Sahilsingh0808:
- **Custom Playback Speed**: Custom Playback speed for the podcast is included by adding a custom input for speed. A new row is added after the playback speed slider and before the Increase volume row. There is a switch in this row. A dialog box will pop up on activating this switch and ask for filling the custom speed. On filling the custom speed (and it being in the correct range, there is a check for that), the playback speed will be updated in the app, and the user can now enjoy whatever pace they wish to listen to the podcast.

- **Flexible Playback Speed Slider**: The playback speed slider is also made more flexible. Earlier the break was 0.25; now, it is changed to 0.05. Due to this, the user now has more flexibility in playback speed to choose from.

Screenshot_20220419_184809

Screenshot_20220419_184920


kingonly:
Thank you @Sahilsingh0808, but changing the speed step and adding a custom speed don't make much sense to me based on users feedback. I know there some advance podcasting apps that have it, but I think it's a bit of an overkill. The only change I would do is to increase the rage up to x3.

Sahilsingh0808:
Thanks for the advice, @kingonly. I will try incorporating the range increase.

PR #888, opened 2022-04-19
Playback Speed Range increase to 3*

Sahilsingh0808:
**Playback Speed Range increased**: The range of playback speed is set from 2.00 to 3.00 to give the user more freedom to choose their podcast streaming speed. The speed steps' remains 0.25 and the number of speed blocks are increased from 6 to 10. 
Screenshot_20220419_194841
Screenshot_20220419_194939

**Flutter Packages updated**: The packages from pub.dev have been updated to their latest version. These include a few packages which turned a few versions outdated in the mean time.


kingonly:
@Sahilsingh0808 that looks good, but why all the commits?

Sahilsingh0808:
Sorry for that, fixing that soon.:)

Issue #780, opened 2022-02-12
[Feature Request] Improve backup by including PoS Settings and data onto backup


tlindi:
Please include Avatar, PoS Settings, Currencies and Items onto backup.

Also Receipt Printing is not possible after restore for PoS transactions prior restore.

kingonly:
@erdemyerebasmaz please take a look. Podcasts in library is another item. Keep in mind that efficiency is key and we already backup breez.db so one approach would be to store everything there. @roeierez should advise.

PR #890, opened 2022-04-20
Include PoS Items, Podcast Library and User Preferences data onto backup

erdemyerebasmaz:
This PR addresses #780.

POS items, Podcasts library and BackupUserPreferences* are now included in backup and saved after each respective change.

---
 #### Breezmobile Changes:  
- [Backwards compatibility] App does not attempt to restore app data if it does not exist in backup.
- POS Items are backed up after CRUD operations, Podcast library is backed up after Subscribe & Unsubscribe events.
- Published user changes are reduced to BackupUserPreferences and compared for changes against stored user preferences before a backup is triggered.
- There are several user preferences that send multiple backup requests by nature but the ones where it can be reduced to **1 change=1 backup request** are updated. 
   - [Business Address](https://github.com/breez/breezmobile/commit/32150ec29d7f8456996e2574879a691b679dc3fd)
   - [Payment Cancellation Timeout](https://github.com/breez/breezmobile/commit/3fcae22cf307d44ac017161e55f2f57340831f6b)
 - Preferences that need rework for **1 change=1 backup request** are as follows:
     - hideBalance: We hide balance and then revert to first preferred currency  at the same time
     - preferredCurrencies: We check for fiat currencies in the list that are above minimum amount to be displayed on Balance screen, revert to first item on list if none are above minimum amount and we also revert posCurrencyShortName to BTC if it's no longer in the list. These are chained logic and can amount to ~4 backup requests with single change.
  ---
 #### Anytime Changes:  
- Reload and Backup Database Stream/Sinks are added to Anytime for 
   - [listening backup requests](https://github.com/breez/anytime_podcast_player/commit/d45d98b27b4b9df38584b1f1484e595af29bc35c) and 
   - [sending reload database requests](https://github.com/breez/anytime_podcast_player/commit/822a7b542dc0c4344569c242486d98bcc43852af)**
---
*: [User preferences that will be used in backup](https://raw.githubusercontent.com/breez/breezmobile/92f3e2fb6c0dc79d73183b6a68c56e7dd002edb4/lib/bloc/user_profile/backup_user_preferences.dart)
**: We are force reloading databases after restore because UI is built with the DB files on startup, pre-restoration. This causes UI not to show DB elements until a CRUD operation is executed, or the app is restarted.


lib/bloc/backup/backup_bloc.dart
```diff
@@ -337,13 +475,48 @@ class BackupBloc {
...
+  }
+
+  Future<void> _restorePosDB() async {
+    final backupPosDbPath =
+        _backupAppDataDirPath + Platform.pathSeparator + 'product-catalog.db';
+    final posDbPath = await databaseFactory.getDatabasesPath() +
+        Platform.pathSeparator +
+        'product-catalog.db';
+    if (await File(backupPosDbPath).exists()) {
+      await File(backupPosDbPath).copy(posDbPath).catchError((err) {
```
roeierez, line 500: Maybe it is better to delete after restore? So in case there is an error and the user decides to restore from a different backup he won't end with the old app data?

lib/bloc/user_profile/user_profile_bloc.dart
```diff
@@ -361,13 +362,31 @@ class UserProfileBloc {
...
       }
+      userToRegister = await _restoreUserPreferences(userToRegister);
     } catch (e) {
       _registrationController.addError(e);
     }
     userToRegister = userToRegister.copyWith(registrationRequested: true);
     await _saveChanges(preferences, userToRegister);
   }
 
+  Future<BreezUserModel> _restoreUserPreferences(
```
roeierez, line 373: Not sure it is a good idea to leak backup implementation details to here. I think a better approach would be to read the user preferences in the backup bloc.

roeierez (changes requested):

lib/bloc/backup/backup_bloc.dart
```diff
@@ -251,9 +341,57 @@ class BackupBloc {
       await _breezLib.signOut();
     }
     await _breezLib.signIn(_backupServiceNeedLogin);
+    await _saveAppData();
```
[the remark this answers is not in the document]
erdemyerebasmaz: No, because _breezLib.requestBackup(); triggers a full backup afaik:
image

Issue #504, opened 2021-06-19
Add flutter lockfile pubspec.lock to repo


emanuelb:
https://dart.dev/tools/pub/glossary#lockfile
https://dart.dev/guides/libraries/private-files#pubspeclock

This will help to reproducible-builds as the same versions can be reused, this file need to be in repos that are app packages (like this repo)

this file exists currently in .gitignore which is incorrect:
https://github.com/breez/breezmobile/blob/b00bba9e9d682fb3ead8148c83f48e70af8f14e7/.gitignore#L21

PR #894, opened 2022-04-26
Make up to date libraries version explicit

ademar111190:
In order to advance on the reproducible build issue referred to https://github.com/breez/breezmobile/issues/247  and https://github.com/breez/breezmobile/issues/504 I'm replacing the ranges syntax `^x.y.z` by specific static version `"x.y.z"`.

These changes don't change any dependency, it just makes the version explicit and avoids future library updates to be applied automatically.

P.S. The collection is not up-to-date but we can't update it because dart SDK version, so I'm fixing the version on this PR too


roeierez:
@ademar111190 I think when you write ^1.3.1 it allows the lock file to apply minor version upgrades so to be on the safe side instead of transforming it to "1.3.1" we should check the latest minor version such as: "1.9.9"

ademar111190:
@roeierez I didn't 🤔  I have focused on the replicable build, not on this aspect.
So it is a problem, do you prefer to revert to the range syntax and add to the git the pubspec.lock file?

roeierez:
> @roeierez I didn't 🤔 I have focused on the replicable build, not on this aspect. So it is a problem, do you prefer to revert to the range syntax and add to the git the pubspec.lock file?

I think you can just change this PR to have the pubspec.lock valus which reflects the current status.

ademar111190:
@roeierez I did back the syntax to range `^x.y.z` and add the lock file.
I think now we have the best of both words.

roeierez (approved):

Issue #897, opened 2022-04-27
Please improve the restore wizard so that you can go back and correct your input words.


fresheneesz:
The restore wizard really needs to be improved. I've restored a good number of times now having gone through a couple phones since I started using breez. The restore wizard is pretty frustrating to use in large part because the input is paged and going back a page deletes everything you've input. This is incredibly frustrating after spending 5 minutes inputting 24 words if it says "can't restore" and you need to go back and input everything all over again. Please fix this. Just put the whole thing on one page please. 

Because I can't go back to previous pages and verify what words I had, I don't know if I'm inputting the words wrong or if there's a bug in the app. I've gotten it wrong enough times while being sure I input it right that my gut feeling is there's a bug in the restore code. Perhaps swype input ends up including invisible spaces that cause it to not work correctly? I don't know. But the first step would be to make it so users can go back and review all 24 words they've put in.

Also, the completion options don't seem to work well with swype and sometimes even with just backspace. The suggestions will stop updating sometimes and also sometimes covers the input you're typing into. I'm on Android 12 btw.

kingonly:
@erdemyerebasmaz please take a look

erdemyerebasmaz:
Hello,
I have reduced the restore wizard to single page and the form is not cleared after a verification fails.
I also fixed a bug where suggestions would cease to work until focus is reset on the input field if user types a word that's not included in the [BIP 39 Wordlist](https://github.com/bitcoin/bips/blob/master/bip-0039/english.txt). Swype seemed to work as expected after this fix according to my tests.
During this implementation I encountered another bug where if the user exits enter backup phrase page and go back to initial walkthrough page, they won't be able to go through another restore unless they restart the app.

fresheneesz:
Thank you!

PR #898, opened 2022-04-27
Do no reset form after changing pages on restore backup phrase

erdemyerebasmaz:
This PR addresses #897  

Restore backup phrase form was being reset upon changing pages. This caused a frustrating experience if the backup failed and users had to correct a word on previous page because they would need to enter all inputs from the start, rather than checking and correcting their existing input. 
Typeahead plugin is also updated to null safety, which caused errors on empty results. Suggestions now return an empty list rather than null to correct this behavior.

- Form is no longer reset upon changing pages.
- Fixed suggestion box behaviour on empty results.


fresheneesz:
🙏

roeierez (approved):

Issue #903, opened 2022-05-04
Remove avatar image from the description


iBobik:
When I pay invoices created by Breez by Phoenix it will look like this:
![IMG_F57E3F357AFC-1](https://user-images.githubusercontent.com/614232/166805668-45676c7b-0df7-4eae-a141-199a977ff22e.jpeg)

When I open it I will see the name, but on the second line and with profile metadata, because it is not standardized:
```
|
Akvárna | breez://profile_image?animal=Pig&color=Blue
```

It is also unusable for anything serious, because we can not add our own picture or logo there and there is no standard about it, so I does not think other wallets will implement it.

How to remove it from my wallet?
How about to remove this feature from Breez?

kingonly:
You can add your own image there. It's used for Breez to Breez payments and cannot be removed. Other wallets can easily support this format. We'll discuss with Phoenix.

iBobik:
Aha, the Gallery button works now, previously it was unresponsive.

But I would prefer to remove it until most other wallets will understand or it will at least be backward compatible.


Honza Pobořil

kingonly:
Maybe putting a default note is the middle ground?
Try to add a note to a sale and see how it looks.

iBobik:
My primary use-case is POS mode, when receiving from friends I does not
care much about the branding.


Honza Pobořil


4. 5. 2022 22:06:49, Roy Sheinfeld ***@***.***> napsal:

kingonly:
You didn't understand my point. Try to add a note to the sale and see how it looks in Phoenix. If it looks good, we can add a default note setting in POS (for example: "McDonald's sale"). With a note, there no need to remove the additional data.

iBobik:
Aha, good idea. Works well.





Honza Pobořil

iBobik:
![image](https://user-images.githubusercontent.com/614232/166835547-f03d6dac-e414-4311-9673-8939e260224a.png)

![image](https://user-images.githubusercontent.com/614232/166835527-5ee9286b-5ee6-43e1-95eb-3c045e484b9a.png)

kingonly:
Cool. We'll add it.

iBobik:
Anyway, I think you should change the format for the avatar, so it is not
broken for most of users when they interact with other wallets.

Specifically, name text should be on the first line, because this will most
realiably work everywhere. Second line could be URL, but only https or data
schemes, because breez://profile_image is not useable for another wallets,
but https://breez.technology/profile_image does.


[...]

PR #907, opened 2022-05-06
Add default note to POS

ademar111190:
Fixes https://github.com/breez/breezmobile/issues/903

|Default|Custom|Cart|
|---|---|---|
|Screenshot_20220506-115458_Breez_Debug|Screenshot_20220506-115443_Breez_Debug|Screenshot_20220506-115429_Breez_Debug|

### DIFF
--- a/lib/bloc/pos_catalog/bloc.dart
+++ b/lib/bloc/pos_catalog/bloc.dart
@@ -7,6 +7,7 @@ import 'package:breez/bloc/pos_catalog/actions.dart';
 import 'package:breez/bloc/pos_catalog/pos_csv_utils.dart';
 import 'package:breez/bloc/pos_catalog/repository.dart';
 import 'package:breez/bloc/pos_catalog/sqlite/repository.dart';
+import 'package:breez/bloc/user_profile/breez_user_model.dart';
 import 'package:breez/bloc/user_profile/currency.dart';
 import 'package:breez/services/breezlib/data/rpc.pb.dart';
 import 'package:breez/services/injector.dart';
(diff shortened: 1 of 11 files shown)


lib/bloc/pos_catalog/bloc.dart
```diff
@@ -176,6 +180,16 @@ class PosCatalogBloc with AsyncActionsHandler {
     });
   }
 
+  void _trackDefaultSaleNote(Stream<BreezUserModel> userStream) {
+    userStream.listen((user) {
+      final defaultSaleNote = user.defaultPosNote;
+      final currentSale = _currentSaleController.value;
+      _currentSaleController.add(currentSale.copyWith(
```
roeierez, line 187: just want to make sure that we don't override existing sale models. Where is the point that the _currentSaleController cleans up existing sales?

ademar111190: It seems to be here: https://github.com/breez/breezmobile/blob/master/lib/bloc/pos_catalog/bloc.dart#L150

roeierez: @ademar111190 what happens if the sale is not paid (e.g cancelled)? does the controller still hold the old value?

ademar111190: If I cancel the value is kept, but if I clear the sale the value is set to empty 🤔 
I guess it should be "cleared" to the default right?

ademar111190: I did apply this change, now the sale has the default note after a clear.

roeierez (approved):

PR #909, opened 2022-05-19
readme: reorganize setting up part in README.md

sunboyy:
After trying to build and run Breez application for a while, I would like to reorganize and update README file so that it provide more detailed (a little bit) and updated information for the new contributors.


roeierez (approved):
Thanks.

Issue #916, opened 2022-06-01
Eurosign printed incorrectly on receipt


tlindi:
Using Nokia 5.1 Plus and (non-paid version) Basic Receipt Printer Driver (paid version name is Receipt Printer Driver) prints euro sign incorrectly on "Item, Price, Amount, Total -line.

"Total sat (nn €)" line prints sign correctly. Please see image.
![IMG_20220601_212208](https://user-images.githubusercontent.com/7249767/171476007-a30567d3-8c9c-421d-bb8a-c9d97cad63bf.jpg)

kingonly:
I just tested with the free driver and I get a wrong sign in the total amount also. This is definitely a bug in the driver as the regular printer works as expected. Not sure what we can do here besides using 'EUR' instead of '€' but that wouldn't look as good.
Can you maybe contact the driver creators?

tlindi:
I'm not sure if EUR would work - U know better ;-)

If multible sales Items can only be only added with one currency then ASCII fiat Name (EUR, USD, etc) would work.

Unfor driver developer is not too find of to reply my contact. Or he asks fees for changes to his product.

kingonly:
@ademar111190 let's change the currency representation in the receipt to 3 letters instead of a symbol. Sales view should be the same as today with one exception: Let's use丰 for sats (currently we use Ş) which is becoming more popular.

kingonly:
Everywhere (all the places in the POS functionality)

kingonly:
Fixed in latest

PR #920, opened 2022-06-07
Change satoshi symbol and print currencies tickers instead of symbol

ademar111190:
Apply the changes proposals from https://github.com/breez/breezmobile/issues/916

- Changes the satoshi symbol from `Ş` to `丰`
- Uses currency ticker (3 letters) instead of the currency symbol.

# How it looks like

|Before|After|
|---|---|
|before-1|after-1|
|before-2|after-2|
|before-3|after-3|
|before-4|after-4|

### DIFF
--- a/lib/bloc/user_profile/currency.dart
+++ b/lib/bloc/user_profile/currency.dart
@@ -55,7 +55,7 @@ class Currency extends Object {
       case "BTC":
         return "₿";
       case "SAT":
-        return "Ş";
+        return "丰";
       default:
         return "₿";
     }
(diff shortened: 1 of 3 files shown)


kingonly:
@ademar111190 thank you 🙏
We need to be consistent in the invoice. In the POS context, always use SAT, not 'sats'.

ademar111190:
That makes sense @kingonly I did update the PDF I'm looking for other screens that may have a sats instead of SAT

kingonly:
Thanks @ademar111190 🙏 I think we're good if you've fixed the PDF.

ademar111190:
@kingonly I found a button using SATS, I did change to SAT. I guess it is correct everywhere now

kingonly:
@ademar111190 what was the button? I want to see how it behaves today.

ademar111190:
@kingonly  This blue button at the bottom:
<details>
  <summary>Screenshot</summary>

Screenshot_20220608-182540_Breez_Debug

</details>

kingonly:
Great catch! Thank you 🙏

roeierez (approved):

PR #922, opened 2022-06-10
Added Podcast History functionality 

### LINKED ISSUE #686: Show stats of my usage of podcasting
I'm a heavy sat-streamer to podcasts in Breez. It would be cool if the app had some option to see stats on my usage of podcasts. How much I send per month, per year. What is the top podcast I listened to, etc.

utkarshshendge:
fixes #686

- The podcast history tab in the drawer is visible in the "podcast" mode.
- As the user listens to the podcast the data is stored locally using the sqflite package.
- The data can be filtered on the range of Daily, This Week, This Month, This Year and All Time.
- The list of podcast history can be sorted.
- Added support for both themes.
- Added share functionality using [screenshot](https://pub.dev/packages/screenshot) package.
- Added language translations.

 <p float="middle">
  
   
</p>

 <p float="middle">
  

</p>

cc- @roeierez @kingonly


kingonly:
@utkarshshendge looks great 👏
Can we use a history icon in the side menu?

kingonly:
@utkarshshendge please rename the screen to 'Top Podcasts' and I'll get our designer to suggest an icon and provide UI feedback.

lib/bloc/podcast_history/podcast_history_bloc.dart
```diff
@@ -0,0 +1,202 @@
...
+  ) async {
+    final prefs = await ServiceInjector().sharedPreferences;
+    prefs.setString(_podcastHistoryTimeRangeKey, action.range.toJson());
+    await fetchPodcastHistory(
+        startDate: action.range.startDate, endDate: action.range.endDate);
+
+    _podcastHistoryRange.add(action.range);
+  }
+
+  Future<List<PodcastHistoryModel>> fetchPodcastHistory(
```
roeierez, line 112: I think it is better to use the sqlite engine to return the result set grouped by podcast id with the totals related to the podcast. the number of payments is far greater than the number of podcasts listened to so it would minimize a lot the rows returned.
We can also do the same for sorting but it is a nice to have, not critical.

lib/bloc/podcast_payments/podcast_payments_bloc.dart
```diff
@@ -110,6 +112,17 @@ class PodcastPaymentsBloc with AsyncActionsHandler {
               seconds: _listeningTime[currentPlayedEpisode.contentUrl].floor())
           .inMinutes;
 
+      if (_listeningTime[currentPlayedEpisode.contentUrl].floor() %
+              updatePodcastHistoryFrequencyInSeconds ==
+          0) {
+        _addToPodcastHistory(
```
roeierez, line 118: shouldn't we add to the podcast history after we pay?

roeierez (changes requested):
Thanks a lot for the PR. This is a very good work!

lib/bloc/podcast_history/podcast_history_bloc.dart
```diff
@@ -0,0 +1,202 @@
...
+      _podcastHistoryRange.stream;
+
+  final BehaviorSubject<bool> _showShareButton = BehaviorSubject();
+
+  Stream<bool> get showShareButton => _showShareButton.stream;
+
+  final BehaviorSubject<PodcastHistorySummationValues>
+      _podcastHistorySumValues = BehaviorSubject();
+
+  Stream<PodcastHistorySummationValues> get podcastHistorySumValues =>
```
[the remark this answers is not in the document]
utkarshshendge: @roeierez yes makes sense. On it!

lib/bloc/podcast_history/podcast_history_bloc.dart
utkarshshendge: @roeierez Sure. For sorting, we are not reading from local DB we are simply modifying the already fetched list.

lib/bloc/podcast_payments/podcast_payments_bloc.dart
utkarshshendge: @roeierez It's also recorded. Please see https://github.com/breez/breezmobile/blob/ab6ba1213bd60bfa12a86ddc544eacd7a2c10bba/lib/bloc/podcast_payments/podcast_payments_bloc.dart#L266

lib/bloc/podcast_history/model.dart
```diff
@@ -0,0 +1,225 @@
...
+      {this.totalSatsStreamedSum,
+      this.totalBoostagramSentSum,
+      this.totalDurationInMinsSum});
+}
+
+const _podcastHistorySortRecentlyAdded = 'SORT_RECENTLY_HEARD';
+const _podcastHistorySortDurationDescending = 'SORT_DURATION_DESCENDING';
+const _podcastHistorySortSatsDescending = 'SORT_SATS_DESCENDING';
+
+abstract class PodcastHistorySortOptions {
```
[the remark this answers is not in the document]
utkarshshendge: Done

lib/bloc/podcast_history/podcast_history_bloc.dart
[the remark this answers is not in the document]
utkarshshendge: Done

lib/bloc/podcast_history/podcast_history_bloc.dart
utkarshshendge: Done

lib/bloc/podcast_history/model.dart
```diff
@@ -0,0 +1,239 @@
...
+  List<PodcastHistoryModel> podcastHistoryList;
+
+  PodcastHistoryRecord(
+      {this.totalSatsStreamedSum,
+      this.totalBoostagramSentSum,
+      this.totalDurationInMinsSum,
+      this.podcastHistoryList});
+}
+
+abstract class PodcastHistorySortOptions {
```
roeierez, line 174: It seems that we don't need this class as we can use directly the enum

lib/bloc/podcast_history/podcast_history_bloc.dart
```diff
@@ -0,0 +1,184 @@
...
+
+  Stream<bool> get showShareButton => _showShareButton.stream;
+
+  final BehaviorSubject<PodcastHistoryRecord>
+      _podcastHistoryRecordBehaviourSubject = BehaviorSubject();
+
+  Stream<PodcastHistoryRecord> get podcastHistoryRecord =>
+      _podcastHistoryRecordBehaviourSubject.stream;
+
+  final BehaviorSubject<List<PodcastHistoryModel>>
```
roeierez, line 26: Can we add the podcastHistoryList to the podcastHistoryRecord? The code shows that they are calculating and changing at the same time (when the report is generated).

lib/bloc/podcast_history/sqflite/podcast_history_database.dart
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  Future<Database> _initDB(String filePath) async {
+    final dbPath = await getDatabasePath();
+    final path = dbPath + filePath;
+    return await openDatabase(path, version: 1, onCreate: _createDB);
+  }
+
+  Future _createDB(Database db, int version) async {
+    final fieldIdType = 'INTEGER PRIMARY KEY AUTOINCREMENT';
```
roeierez, line 29: Not sure these constants help make the code more readable. It seems easier to follow a complete sql statement string.

roeierez (changes requested):
Almost there! just a few comments.

lib/bloc/podcast_history/model.dart
utkarshshendge: @roeierez  Yes. Also, I think it's better if we do not store the selected sort options in `shared_preferences` this looks confusing and other apps don't do it.

roeierez: @utkarshshendge  so why did you implement it?

lib/bloc/podcast_history/sqflite/podcast_history_database.dart
utkarshshendge: @roeierez Something like this?

```
 Future _createDB(Database db, int version) async {
    await db.execute('''
CREATE TABLE $podcastHistoryTable(
${PodcastHistoryFields.fieldId} INTEGER PRIMARY KEY AUTOINCREMENT,
${PodcastHistoryFields.podcastId} TEXT NOT NULL,
${PodcastHistoryFields.timeStamp} TEXT NOT NULL,
${PodcastHistoryFields.satsSpent} INTEGER NOT NULL,
${PodcastHistoryFields.boostagramsSent} INTEGER NOT NULL,
${PodcastHistoryFields.podcastName} TEXT NOT NULL,
${PodcastHistoryFields.podcastImageUrl} TEXT NOT NULL,
${PodcastHistoryFields.durationInMins} DECIMAL NOT NULL
 )
 ''');
  }
```

lib/bloc/podcast_history/podcast_history_bloc.dart
utkarshshendge: Done

lib/bloc/podcast_history/model.dart
utkarshshendge: Done

roeierez (approved):
LGTM!

PR #930, opened 2022-06-22
Add support to pay POS with NFC

ademar111190:
An important piece of information for the reviewers, I don't have a device with NFC so I did test this functionality using mock, you can see in the video a button "not present on this PR" with an NFC icon that I tap simulating an NFC event.

A second thing important to know is I could not test real cases where the withdraw LnUrl allows ranges, all tests I did successfully including the recorded ones below are withdrawn LnUrl with a fixed amount.

So before approval, I'm kindly asking you to run this branch and test with a real NFC device and a real range full withdraw LnUrl.

How it looks like:

# Success (manual value)

https://user-images.githubusercontent.com/1225438/175128199-055a5692-df86-44b3-a7cb-b2eaacc60835.mp4

# Success (item value)

https://user-images.githubusercontent.com/1225438/175128197-f1dbb286-8158-40ac-b5b1-4616c219a37b.mp4

# Error - when the value is out of range of the LnUrl withdraw

https://user-images.githubusercontent.com/1225438/175128181-7b53c4f9-1cc3-4e74-8ede-8a0c26a7d35a.mp4


tlindi:
Stunning Works :)
I could sponsor (with reasonable Price) test phone  with NFC to Breez. Please DM here or TG how to proceed, please?

roeierez:
@ademar111190 this use case is complicated as we have LNURLHandler that handles all lnurl responses and now we need to handle withdrawal differently... It took  me a while to build my opinion here.
Right now the way how the code intercepts the responses seems a bit complicated since a UI widget (pos) interacts with another UI handler (LNURL) that was made singleton only for the ability to add interception... I was trying to think how to solve the problem with our simple rules where a UI component only access bloc for business logic maintaining the unidirectional flow and came to this suggestion:
1. Pos will register the invoice by invoking an action on LNURLBloc (and not LNURLHandler)
2. LNURLBloc when receiving lnurl withdraw response, checks if it has an invoice from pos.
3. If it has an invoice that satisfies the min/max amounts it responds immediately with withdraw and not adding event to the _lnUrlStreamController
4. If it hasn't an invoice flow keeps as it is today by adding the event to the _lnUrlStreamController

This has the following benefits:
1. LNURLHandler is kept as is without being made accessible globally
2. POS only interacts with the bloc as it should
3. POS is not aware of LNURLWithdraw or how the invoice is paid, just register the invoice and keeps waiting for payment as today.

LMK what you think and if I am missing something.

ademar111190:
@roeierez `LNURLHandler` is used directly by UIs `Home` and `QrActionButton` but it is not a UI itself, I believe the problem is this concept of Handlers that is a mix of UI and business rule, converting it to Bloc sounds a good thing, but it is a refactor in my opinion.

As it is a refactor and as it involves testing the Home, QrAction, and other LNUrls types: `ChannelFetchResponse`, `AuthFetchResponse`, and `PayFetchResponse` "if not more areas that I've no knowledge of right now" I prefer to have a separate PR to do these changes… What do you think?

lib/bloc/lnurl/lnurl_bloc.dart
```diff
@@ -69,8 +71,12 @@ class LNUrlBloc with AsyncActionsHandler {
         _lnUrlStreamController.add(fetchLNUrlState.completed);
         if (response.runtimeType == LNUrlResponse) {
           if (response.hasWithdraw()) {
-            _lnUrlStreamController
-                .add(WithdrawFetchResponse(response.withdraw));
+            final withdrawResponse = WithdrawFetchResponse(response.withdraw);
+            if (withdrawFetchResponseInterceptor != null) {
+              withdrawFetchResponseInterceptor.intercept(withdrawResponse);
```
roeierez, line 76: A lot better but the issue here is that still the logic is delegated to the UI instead of being executed here.
If the UI would have registered only the invoice (withdrawalInvoice) here (instead of the interceptor) the bloc could handle this automatically and complete the lnurl flow by invoking _withdraw function directly.
Wouldn't it be cleaner in your opinion?

roeierez (changes requested):

lib/bloc/lnurl/lnurl_bloc.dart
ademar111190: That makes sense, let me see if that works 👀

ademar111190: I did move the logic to the bloc, I think it is ready for a new review @roeierez

lib/routes/charge/pos_payment_dialog.dart
```diff
@@ -82,12 +97,25 @@ class _PosPaymentDialogState extends State<PosPaymentDialog> {
         }
       });
     });
+
+    _nfcInvoiceSubscription =
+        widget._lnUrlBloc.nfcWithdrawStream.listen(_listenNfcWithdraw);
+
+    widget._lnUrlBloc.actionsSink.add(RegisterNfcSaleRequest(
```
roeierez, line 104: isn't it better to send the paymentRequest along with the sale request? It ensures we use the same invoice in the QR and the NFC and eliminate the need to create another invoice in the bloc.

Also I think we shouldn't count on the NfcWithdrawInvoiceStatus.completed status to signal payment succeed but we should wait for the invoice to be paid as we do with the QR which is another reason why to use the same bolt11 payment request.

roeierez (changes requested):
@ademar111190 looks good now. See my comment regarding passing the bolt11 to the sale nfc request.

roeierez (approved):

PR #935, opened 2022-06-30
Add high quality image for top podcasts

utkarshshendge:
- [x] Add a high-quality png image for My Top Podcasts.

 <p float="middle">
  
</p>


ademar111190:
Thank you for the improvement @utkarshshendge 

May I ask you to reduce the image size to 512x512? The other icons from the drawer menu have this size and they look great, so an image of 1280x1280 with 36.6kb seems unnecessary big.

If possible uses a lossless image compressor to reduce the png size and remove unnecessary metadata too, some options depending on your operational system:

- Mac (https://imageoptim.com/mac)
- Linux (https://trimage.org/)
- Windows (https://css-ig.net/pinga)

Just make sure to use lossless not lossy as we don't want to lose quality we just want to reduce the size to make the app smaller as possible 🙏

utkarshshendge:
@ademar111190 I mistakenly added a different size image. Creating another PR

ademar111190:
> @ademar111190 I mistakenly added a different size image. Creating another PR

It is better to use the same PR, you can rebase or squash your commits, as you already closed this one go ahead and open another, next time just uses the same PR 👍

PR #940, opened 2022-07-04
Update dependencies fixing date NPE on some podcast episodes

ademar111190:
Some podcast episodes use a non-standard date format.

e.g. `Podcasting 4 Value` episode 1 uses the expected format `<pubDate>Mon, 17 Jan 2022 04:02:29 -0700</pubDate>` but episode 16 doesn't `<pubDate>20 June 2022 15:12:29 -0700</pubDate>` as you can see it is missing the day of the week in the beginning.

so what I did:

- I did add this pattern to the parser: https://github.com/breez/podcast_search/commit/13bc13bda6432d53905b7b72217b810c5f2db92e so now that pattern without the day of the week is supported
- I did add a check and a warning for whenever a similar problem happens in the future so the app will ignore the date and log a warning so we can implement that specific date pattern later. https://github.com/breez/anytime_podcast_player/commit/4373ff9a9bc6d006729f80f34a1dd9cd72b0722b

# Result

|Before|After|
|---|---|
|Screenshot_2022-07-03-22-24-12-22_73a31df444cba46df764485ac1a8c3ef|Screenshot_20220704-122900_Breez Debug|

### DIFF
--- a/lib/routes/podcast/podcast_index_api.dart
+++ b/lib/routes/podcast/podcast_index_api.dart
@@ -46,14 +46,16 @@ class PodcastIndexAPI extends PodcastApi {
   }
 
   static Future<SearchResult> _search(String term) {
-    return Search(userAgent: Environment.userAgent())
-        .search(
-          term,
-          queryParams: {"val": "lightning"},
+    return Search(
+          userAgent: Environment.userAgent(),
           searchProvider: PodcastIndexProvider(
             key: 'XXWQEGULBJABVHZUM8NF',
             secret: 'KZ2uy4upvq4t3e\$m\$3r2TeFS2fEpFTAaF92xcNdX',
           ),
+        )
+        .search(
+          term,
+          queryParams: {"val": "lightning"},
         )
         .timeout(Duration(seconds: 10));
   }
@@ -61,11 +63,11 @@ class PodcastIndexAPI extends PodcastApi {
   static Future<SearchResult> _charts(int size) {
     return Search(
       userAgent: Environment.userAgent(),
-    ).charts(
       searchProvider: PodcastIndexProvider(
         key: 'NBVJ9GPWMLPXJMFFD3KV',
         secret: 'be7kR2CWeCFk4nkKCKV4XAX35y2eXrDeCPvs#v4F',
       ),
+    ).charts(
       queryParams: {
         'val': 'lightning',
         'aponly': 'true',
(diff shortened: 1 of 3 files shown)


roeierez:
> * I did add this pattern to the parser: [breez/podcast_search@13bc13b](https://github.com/breez/podcast_search/commit/13bc13bda6432d53905b7b72217b810c5f2db92e) so now that pattern without the day of the week is supported

isn't it better to use DateFormat to parse instead of regex?

ademar111190:
> > * I did add this pattern to the parser: [breez/podcast_search@13bc13b](https://github.com/breez/podcast_search/commit/13bc13bda6432d53905b7b72217b810c5f2db92e) so now that pattern without the day of the week is supported
> 
> isn't it better to use DateFormat to parse instead of regex?

I'm not sure if it is better. There are other 3 patterns already using regex, which is something to consider, uses DateFormat for all these 4 patterns or just for this new pattern? TBH I don't know if it's worth it.

kingonly:
> > > * I did add this pattern to the parser: [breez/podcast_search@13bc13b](https://github.com/breez/podcast_search/commit/13bc13bda6432d53905b7b72217b810c5f2db92e) so now that pattern without the day of the week is supported
> > 
> > 
> > isn't it better to use DateFormat to parse instead of regex?
> 
> I'm not sure if it is better. There are other 3 patterns already using regex, which is something to consider, uses DateFormat for all these 4 patterns or just for this new pattern? TBH I don't know if it's worth it.

I agree, we don't really want regressions and it's not really important to refactor

roeierez:
I actually was thinking of one line of code instead of 15: 
`DateFormat("d MMM yyyy hh:mm:ss ").parse(date)`

But trying to implement it and discovered DateFormat surprisingly doesn't support timezone offset so I guess this was the reason from the first place to go with regex.

roeierez (approved):

ademar111190:
> I actually was thinking of one line of code instead of 15: `DateFormat("d MMM yyyy hh:mm:ss ").parse(date)`
> 
> But trying to implement it and discovered DateFormat surprisingly doesn't support timezone offset so I guess this was the reason from the first place to go with regex.

I didn't like that too, 15+ lines, but it is how it was made for the other 3 cases, I just add one more.

Dart std or pub.dev not having a library that does the conversions was a surprise to me.

PR #950, opened 2022-07-21
Update QR code scanner plugin

erdemyerebasmaz:
qr_code_scanner isn't working as expected after Flutter 3.0.x update, it requires going into Gallery and returning back to QR screen to start scanning. This is likely a render order/state management issue but I think this is a great opportunity to upgrade [qr_code_scanner](https://pub.dev/packages/qr_code_scanner) that's in maintenance only mode to [mobile_scanner](https://pub.dev/packages/mobile_scanner) by the same author.
- qr_code_scanner's underlying frameworks are no longer maintained,
  - Replaced qr_code_scanner with mobile_scanner, that uses the latest version of MLKit for detecting barcodes and QR codes
  - Created a ScanOverlay widget to match overlay behavior of qr_code_scanner
  - Replaced helper methods with Widgets

### DIFF
--- a/lib/routes/qr_scan.dart
+++ b/lib/routes/qr_scan.dart
@@ -1,13 +1,13 @@
-import 'dart:async';
 import 'dart:io';
 
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
-import 'package:flutter_gen/gen_l10n/app_localizations.dart';
 import 'package:flutter_svg/svg.dart';
 import 'package:image_picker/image_picker.dart';
-import 'package:qr_code_scanner/qr_code_scanner.dart';
-import 'package:qr_code_tools/qr_code_tools.dart';
+import 'package:mobile_scanner/mobile_scanner.dart';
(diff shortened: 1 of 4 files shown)


lib/routes/qr_scan.dart
```diff
@@ -54,85 +58,136 @@ class QRScanState extends State<QRScan> {
...
-          )
+                ? const QRScanCancelButton()
+                : const SizedBox(),
+          ),
+          const ScanOverlay(),
         ],
       ),
     );
   }
+  /*
```
ademar111190, line 76: I believe this commented code can be deleted right?

ademar111190 (approved):
LGTM

lib/routes/qr_scan.dart
erdemyerebasmaz: Yes, they were added for demonstration purposes. We were discussing if these features would be beneficial. I will remove them and merge the changes. Thanks.

PR #954, opened 2022-08-02
Move Vendors out of marketplace.conf

erdemyerebasmaz:
Our build systems are currently reading all config files from configuration repository. This meant making changes to marketplace.conf requires access to configuration repository and if you don't have access, it requires internal communication. Although not severe and resolved quickly, this also caused app code-config mismatches from time to time.

It's decided that there are no security concerns of making contents of marketplace.conf public. It's no longer tagged as a config file and it's contents are moved to JSON file called vendors.json.

- Removed conf/marketplace.conf and moved it's contents to vendors.json
- Removed logo field from VendorModel
- Renamed LNMarkets logo to match `${vendorID}_logo_lg.png` format
- VendorList is loaded from vendors.json and added to vendor stream instead of config file


roeierez (approved):

erdemyerebasmaz:
Our current structure
```
{
    "vendorID2": {
      "url" : ""
    },
    "vendorID1": {
      "url" : ""
    },
    "vendorID3": {
      "url" : ""
    },
}
```
Should I add order field to ease ordering and removal from vendor list without removing it's data?
```
{
"order": ["vendorID2", "vendorID1", "vendorID3"],
"vendors": [
   {
    "vendorID1": {
      "url" : ""
    },
    "vendorID2": {
      "url" : ""
    },
    "vendorID3": {
      "url" : ""
    },
  }
}
```

PR #967, opened 2022-08-13
Feat: Added feature to create podcast clips

utkarshshendge:
Added functionality to create and share podcast clips.

- [x] Episode is paused when the clip UI appears. 
- [x]  User can add custom clip duration ranging from 10 secs to 120 secs.
- [x] [ffmpeg_kit_flutter](https://pub.dev/packages/ffmpeg_kit_flutter) is used for processing clips.
- [x] Added waveform for the clips.
- [x] Created clips can be shared.
- [x] Added translations for relevant strings.

 <p float="middle">
  
   
</p>

https://user-images.githubusercontent.com/57677520/184511088-1e1bf93e-f764-4a57-af2f-9f12eda3f3a0.mp4


lib/bloc/podcast_clip/podcast_clip_bloc.dart
```diff
@@ -0,0 +1,238 @@
...
+        episodeLength: position.length,
+        episodeDetails: position.episode);
+
+    _clipDetailsBehaviourSubject.add(podcastClipDetails);
+  }
+
+  setPodcastClipDuration({int durationInSeconds}) async {
+    var clipDetails = await getCurrentPodcastClipDetails();
+
+    if ((clipDetails.endTimeStamp + Duration(seconds: durationInSeconds)) >
```
roeierez, line 53: What does this code check?

roeierez: Also are we missing a check here for _maxClipDuration?

lib/bloc/podcast_clip/podcast_clip_bloc.dart
```diff
@@ -0,0 +1,238 @@
...
+    Duration clipEndTime = clipDetails.endTimeStamp +
+        Duration(seconds: durationInSeconds) -
+        Duration(seconds: clipDetails.clipDuration);
+
+    _clipDetailsBehaviourSubject.add(clipDetails.copy(
+        clipDuration: durationInSeconds, endTimeStamp: clipEndTime));
+  }
+
+  incrementDuration() async {
+    var clipDetails = await getCurrentPodcastClipDetails();
```
roeierez, line 67: I think this function should calculate the new duration and call `setPodcastClipDuration` to actually validate and set the new duration. This will reuse the validation code.

lib/bloc/podcast_clip/podcast_clip_bloc.dart
```diff
@@ -0,0 +1,238 @@
...
+      incrementedClipEndTime = Duration(
+          seconds: clipDetails.endTimeStamp.inSeconds + _initialSeconds);
+    }
+
+    _clipDetailsBehaviourSubject.add(clipDetails.copy(
+        clipDuration: incrementedClipDuration,
+        endTimeStamp: incrementedClipEndTime));
+  }
+
+  decrementCount() async {
```
roeierez, line 97: decrementDuration maybe?
Same as with incrementDuration if possible to reuse the `setPodcastClipDuration` it would leave a single validation point.

lib/bloc/podcast_clip/podcast_clip_bloc.dart
```diff
@@ -0,0 +1,238 @@
...
+        '-i "$episodeUrl" -ss ${clipDetails.startTimeStamp.inSeconds} -to ${clipDetails.endTimeStamp.inSeconds}  -acodec copy $audioClipPath';
+
+    FFmpegSession ffpegSessionDetails =
+        await FFmpegKit.execute(clippedAudioCommand);
+
+    final returnCode = await ffpegSessionDetails.getReturnCode();
+    if (ReturnCode.isSuccess(returnCode)) {
+      return audioClipPath;
+    } else {
+      return null;
```
roeierez, line 143: why not just throw error? returning null will swallow the real error.

lib/bloc/podcast_clip/podcast_clip_bloc.dart
```diff
@@ -0,0 +1,238 @@
...
+    FFmpegSession ffpegSessionDetails =
+        await FFmpegKit.execute(clippedVideoCommand);
+
+    String logs = await ffpegSessionDetails.getLogsAsString();
+    log('cliplogs $logs');
+    final returnCode = await ffpegSessionDetails.getReturnCode();
+    if (ReturnCode.isSuccess(returnCode)) {
+      return videoClipPath;
+    } else {
+      return null;
```
roeierez, line 163: throw error?

lib/bloc/podcast_clip/podcast_clip_bloc.dart
```diff
@@ -0,0 +1,238 @@
...
+    }
+  }
+
+  Future<bool> isEpisodeClipable() async {
```
roeierez, line 167: Why do we need this method? It seems like the code ensures that the clipDetails always maintains clip-able episode.

lib/bloc/podcast_clip/podcast_clip_bloc.dart
```diff
@@ -0,0 +1,238 @@
...
+const int _maxClipDuration = 120;
+
+class PodcastClipBloc with AsyncActionsHandler {
+  final BehaviorSubject<PodcastClipDetailsModel> _clipDetailsBehaviourSubject =
+      BehaviorSubject();
+
+  Stream<PodcastClipDetailsModel> get clipDetails =>
+      _clipDetailsBehaviourSubject.stream;
+
+  final BehaviorSubject<PodcastClipState> _clipStateBehaviourSubject =
```
roeierez, line 31: Can we hold the status on the PodcastClipState model? It seems a fit and we can reduce the number of streams to one.

lib/bloc/podcast_clip/podcast_clip_bloc.dart
```diff
@@ -0,0 +1,238 @@
...
+
+    return videoClipPath;
+  }
+
+  setPodcastState(PodcastClipState clipState) {
+    _clipStateBehaviourSubject.add(clipState);
+  }
+
+  Future<String> initClipDirectory({bool isVideoClip}) async {
+    Directory appDocDirectory = await getApplicationDocumentsDirectory();
```
roeierez, line 206: Maybe better to use `getTemporaryDirectory` ?

lib/routes/podcast/podcast_clip.dart
```diff
@@ -0,0 +1,559 @@
...
+              if (position != null) {
+                podcastClipBloc.setPodcastClipDetails(position: position);
+              }
+              bool canBeclipped = await podcastClipBloc.isEpisodeClipable();
+              if (!canBeclipped) {
+                return;
+              }
+              _pause(audioBloc);
+
+              showModalBottomSheet<void>(
```
roeierez, line 52: It is hard to review and maintain such a large widgets tree. Better to break this into separate widgets.

roeierez (changes requested):
@utkarshshendge I have done an initial pass. The overall structure looks good. I left some comments.
Ping me once you finish with the feedback and I will do another pass.

lib/bloc/podcast_clip/podcast_clip_bloc.dart
```diff
@@ -0,0 +1,238 @@
...
+    finalPath = x.path;
+    return finalPath;
+  }
+
+  Future<PodcastClipDetailsModel> getCurrentPodcastClipDetails() async {
+    try {
+      return await clipDetails.first.timeout(Duration(seconds: 1));
+    } catch (e) {
+      return null;
+    }
```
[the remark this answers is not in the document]
utkarshshendge: Done

lib/bloc/podcast_clip/podcast_clip_bloc.dart
utkarshshendge: It checks if the  `endTimeStamp` of clip duration exceeds the episode length. 
When a user enters custom duration in the text field `_maxClipDuration` is validated.

lib/bloc/podcast_clip/podcast_clip_bloc.dart
utkarshshendge: Done.

lib/bloc/podcast_clip/podcast_clip_bloc.dart
utkarshshendge: Done. Thanks for highlighting it.

lib/bloc/podcast_clip/podcast_clip_bloc.dart
utkarshshendge: Added throw.

lib/bloc/podcast_clip/podcast_clip_bloc.dart
utkarshshendge: Done.

lib/bloc/podcast_clip/podcast_clip_bloc.dart
utkarshshendge: This method checks if there is time to create a clip of minimum 10 secs. If not, the bottom sheet won't be shown to the user.

lib/bloc/podcast_clip/podcast_clip_bloc.dart
utkarshshendge: Done. Thanks

lib/bloc/podcast_clip/podcast_clip_bloc.dart
utkarshshendge: Done

lib/routes/podcast/podcast_clip.dart
utkarshshendge: Refractored the widget code.

utkarshshendge:
> @utkarshshendge I have done an initial pass. The overall structure looks good. I left some comments. Ping me once you finish with the feedback and I will do another pass.

Thanks a lot @roeierez.
I have done the changes.

lib/bloc/podcast_clip/podcast_clip_bloc.dart
```diff
@@ -0,0 +1,206 @@
...
+    _clipDetailsBehaviourSubject.add(clipDetails.copy(
+        clipDuration: durationInSeconds, endTimeStamp: clipEndTime));
+  }
+
+  incrementDuration() async {
+    var clipDetails = getCurrentPodcastClipDetails();
+    int currentClipDuration = clipDetails.clipDuration;
+
+    //If duration is less than 10 seconds return
+    if (_maxClipDuration - currentClipDuration <= 0) {
```
roeierez, line 58: Seems like ensuring the clip duration is within bounds belongs in the `setPodcastClipDuration` function.
This way we will have the proper validation for custom duration set by the user.

lib/bloc/podcast_clip/podcast_clip_bloc.dart
```diff
@@ -0,0 +1,206 @@
...
+    }
+    setPodcastClipDuration(durationInSeconds: incrementedClipDuration);
+  }
+
+  decrementDuration() async {
+    var clipDetails = getCurrentPodcastClipDetails();
+    int currentClipDuration = clipDetails.clipDuration;
+
+    //return if duration needs to be greator than 10 seconds
+    if (currentClipDuration <= _initialSeconds) {
```
roeierez, line 77: Seems like ensuring the clip duration is within bounds belongs in the setPodcastClipDuration function.

lib/bloc/podcast_clip/podcast_clip_bloc.dart
```diff
@@ -0,0 +1,206 @@
...
+    }
+    String audioClipPath = await _getAudioClipPath(clipDetails: clipDetails);
+
+    if (audioClipPath == null) {
+      return null;
+    }
+
+    String videoClipPath = await getVideoClipPath(
+        audioClipPath: audioClipPath, episodeImagepath: podcastImageWidgetPath);
+    _clipDetailsBehaviourSubject
```
roeierez, line 169: changing status to idle should be in try/finally block as it should always happen. Right now the code returns null at some points and the status stays `FETCHING_AUDIO_CLIP`

lib/routes/podcast/podcast_clip.dart
```diff
@@ -0,0 +1,492 @@
...
+                                                                  clipDetailSnapshot
+                                                                      .data
+                                                                      .episodeDetails),
+                                                          context: context,
+                                                          delay: const Duration(
+                                                              milliseconds:
+                                                                  10));
+
+                                              String path =
+                                                  await podcastClipBloc
```
roeierez, line 180: try/catch and show error maybe?

lib/bloc/podcast_clip/podcast_clip_bloc.dart
roeierez: Also this function throws exception when it fails to clip so the caller should catch this exception and show error to the user (I believe flushbar will fit here).

lib/bloc/podcast_clip/podcast_clip_bloc.dart
```diff
@@ -0,0 +1,206 @@
...
+    var clipDetails = getCurrentPodcastClipDetails();
+    _clipDetailsBehaviourSubject.add(clipDetails.copy(
+        podcastClipState: PodcastClipState.FETCHING_AUDIO_CLIP));
+    if (clipDetails == null) {
+      return null;
+    }
+    String audioClipPath = await _getAudioClipPath(clipDetails: clipDetails);
+
+    if (audioClipPath == null) {
+      return null;
```
roeierez, line 164: We don't need this test as the function throws if it can't get the audio clip path. We can count on that the audioClipPath is not null.

lib/bloc/podcast_clip/podcast_clip_bloc.dart
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    String videoClipPath = await getVideoClipPath(
+        audioClipPath: audioClipPath, episodeImagepath: podcastImageWidgetPath);
+    _clipDetailsBehaviourSubject
+        .add(clipDetails.copy(podcastClipState: PodcastClipState.IDLE));
+
+    if (videoClipPath == null) {
```
roeierez, line 172: Same here, we don't need this condition. In general it is better to count on exception thrown instead of returning null as exceptions contains more information.

roeierez (changes requested):
@utkarshshendge this is a good work! thanks for addressing the feedback quickly. I went over the PR again and added my final feedback.
Thank you.

lib/bloc/podcast_clip/podcast_clip_bloc.dart
utkarshshendge: Yes done.

lib/bloc/podcast_clip/podcast_clip_bloc.dart
utkarshshendge: Done

lib/routes/podcast/podcast_clip.dart
utkarshshendge: Added 👍

lib/bloc/podcast_clip/podcast_clip_bloc.dart
utkarshshendge: > @utkarshshendge this is a good work! thanks for addressing the feedback quickly. I went over the PR again and added my final feedback. Thank you.

@roeierez I have done the changes. Thanks

roeierez (approved):
Good work

PR #975, opened 2022-09-03
Update app_fi.arp Translate Finnish Fixes and updates

tlindi:
Between 03.08 and 03.09.2022. Changes Lähetä and Vastaanota submenus to more streetwise financial terms.


tlindi:
After long time I messed (yet again) commits. This void.

Issue #978, opened 2022-09-19
Bad Message UX : Funding Breez Wallet from an empty NFC-Card


arbadacarbaYK:
When using a LNURLw on the POS to receive funds while having too less sats on the card / card is empty the message is too nerdy and not helpful. Should say smth like "Not enough funds available"
<img width="415" alt="Bildschirmfoto 2022-09-19 um 23 55 28" src="https://user-images.githubusercontent.com/63317640/191126596-aad29c60-ba24-4ac4-af8f-a4264fbc6688.png">

kingonly:
The "Link not working" message is coming from the lnurl service (in this case probably lnbits).
@ademar111190 let's fix the formatting though and just display the error coming from the service w/o the error domain/nslocalizeddescription etc.

arbadacarbaYK:
exactly 🥳 Thanks!

kingonly:
Another use case to improve: https://t.me/breez_lightning/25393

PR #979, opened 2022-09-21
Improve error handling

ademar111190:
I need help testing iOS to check if this case https://github.com/breez/breezmobile/issues/978 is correctly covered.

The other two know cases (https://t.me/breez_lightning/25393 and one I have experienced myself) are correctly inserted, and I did add unit tests to help add future additions.

This structure allows us to add more exception message extraction improvements and be used by other parts of the app.


lib/handlers/lnurl_handler.dart
```diff
@@ -26,7 +27,7 @@ class LNURLHandler {
         return executeLNURLResponse(context, lnurlBloc, response);
       }
     }).onError((err) async {
-      final errorMessage = _getErrorMessage(texts, err.toString());
+      final errorMessage = _getErrorMessage(texts, err.toString()) ?? extractExceptionMessage(err);
```
erdemyerebasmaz, line 30: I suggest we pass error object to `_getErrorMessage(...)` and return `extractExceptionMessage` as it's default case instead of using nullish coalescing operator(**??**) here for sound null safety purposes.

lib/utils/exceptions.dart
```diff
@@ -0,0 +1,45 @@
...
+    _ExtractMessageFromStringRegexHelper(r'LNURL: .* method:', r'LNURL: ', r' method:'),
+  ]) {
+    final message = helper.extract(exception);
+    if (message != null) {
+      return message;
+    }
+  }
+
+  return null;
+}
```
erdemyerebasmaz, line 25: I believe we can use a single regex to extract the message instead of matching through a loop of _ExtractMessageFromStringRegexHelper's and calling extract function to remove lhs & rhs with replace operations.

Something like:
`((?<={NSLocalizedDescription=)(.*)(?=}))|((?<=LNURL:.)(.*)(?=.method:))|((?<=\"detail\":\")(.*)(?=.*\"}))`

It uses capture groups to combine the regexp's that we're checking against and extracts the message with the use of lookaround assertions.

We can define group strings on their own and join them with "|" for readability.

ademar111190: Grouping like that, how can we remove the leading and trailing parts of the string?

Right now I'm doing that:
`match.group(0).replaceAll(RegExp(lhsReplace), '').replaceAll(RegExp(rhsReplace), '');`

lib/handlers/lnurl_handler.dart
ademar111190: Nice suggestion, applied.

lib/utils/exceptions.dart
erdemyerebasmaz: The regex I mentioned make use of lookaround assertions(look behind, look ahead) to retrieve the part that's between leading and trailing characters. It returns the match only if it's preceeded and followed by those assertions.
Here's a straightforward source on lookaround assertions: https://javascript.info/regexp-lookahead-lookbehind

ademar111190: Interesting, I believe I managed to apply it; not sure if there is a better way.

erdemyerebasmaz: I believe we can merge extractExceptionMessage and _extractMessageFromString into a single function. There's also no need to go through all matches, using grouped.stringMatch(exception) directly to get first match(or null) is sufficient.

ademar111190: Ahh, much better.

About the methods, the idea is to have different extract strategies given the error type; I did apply your suggestion for now as we have only String; next time I add a new style, I can break it again; there is no need to overengineering for now.

lib/utils/exceptions.dart
```diff
@@ -0,0 +1,13 @@
...
+  final nsLocalizedRegex = r'((?<={NSLocalizedDescription=)(.*)(?=}))';
+  final lnUrlRegex = r'((?<=LNURL:.)(.*)(?=.method:))';
+  final grouped = RegExp('$nsLocalizedRegex|$lnUrlRegex|$detailRegex');
+
+  final match = grouped.stringMatch(exception.toString());
+  if (match != null) {
+    return match;
+  }
+
+  return exception.toString();
```
erdemyerebasmaz, line 12: We can make use of null-aware operator here:
return grouped.stringMatch(exception.toString()) ?? exception.toString();

ademar111190: 🤯 applied

PR #1008, opened 2022-11-26
Add Episode Item ID to Podcast Payment TLV

MerryOscar:
This change adds the PodcastIndex `itemID` to the podcast payment TLV records.

The `itemID` is very useful for podcasters and services on the receiving side to properly associate boosts with episodes, as the episode titles can sometimes change which is unreliable.

### DIFF
--- a/lib/bloc/podcast_payments/podcast_payments_bloc.dart
+++ b/lib/bloc/podcast_payments/podcast_payments_bloc.dart
@@ -406,6 +406,7 @@ class PodcastPaymentsBloc with AsyncActionsHandler {
     tlv["action"] = boost ? "boost" : "stream";
     tlv["time"] = _formatDuration(position.position);
     tlv["feedID"] = _getPodcastIndexID(episode);
+    tlv["itemID"] = episode.id;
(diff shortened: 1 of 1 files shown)


kingonly:
@ademar111190 please review and merge

adamc199:
Since this change was implemented, https://saturn.fly.dev/dashboard which is a stats package for getalby, is apparently receiving different TLV episode ID's and is displaying  each payment from Breez as different episodes. 

All other v4v apps show up under one episodeID including fountain

I cannot look at the TLV's to see the episode ID's now added to the TLV record. I'm presuming is is not consistent.

kingonly:
@ademar111190 can you please make sure this isn't an issue on the Breez end?

kingonly:
I've looked at the code again, it looks good on our end. @MerryOscar how does it look on yours?

ademar111190:
The issue is on our side; taking a close look, we are sending the `id` (local DB id) instead of the `guid` received from the podcast source itself.
This PR should fix it https://github.com/breez/breezmobile/pull/1032

adamc199:
Thank you for double checking!

PR #1012, opened 2022-12-05
Extract error message on send BTC flow

ademar111190:
Fix the reported bug [add-error-extraction-on-send-btc-flow
//github.com/breez/breezmobile/pull/new/add-error-extraction-on-send-btc-flow](https://t.me/breez_lightning/26414)

I couldn't test on an actual iOS device; the error message already comes without the messy part on Android.

How it looks like, (Not exactly the same error as reported as it is on Android):

Screenshot_20221205_192740_Breez Debug

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -897,6 +897,15 @@
   "reverse_swap_confirmation_error_funds_fee": "You don't have enough funds for this payment network fee.",
   "reverse_swap_notification_title": "Action Required",
   "reverse_swap_notification_body": "Please open Breez to complete your requested transaction.",
+  "reverse_swap_upstream_generic_error_message": "{errorMessage}. Please try again later.",
+  "@reverse_swap_upstream_generic_error_message": {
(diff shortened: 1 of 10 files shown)


kingonly:
@ademar111190 what is the error that we'll display? I want to make sure it's user friendly. The user needs to understand they should retry to send.

ademar111190:
@kingonly Right now, it is not user-friendly; it just extracts the error message and shows it instead of the previous "error.toString()" approach, but the message extracted is still not helpful.

To make something user-friendly, I need to dig into the breez go lib to change how this specific error is propagated upstream.

Or we can add a user-friendly message and a retry button to **all** error cases; not sure which is better.

kingonly:
Let's keep it simple and postfix "Please try again later." at the end (no retry button). Make sure there's a dot at the end of the previous sentence.

ademar111190:
I did apply the changes and added a screenshot so we can see what it looks like. Keep in mind that the error is not the same on iOS as I did the test on Android.

PR #1016, opened 2022-12-07
display txid first and not claimTxID

ubbabeck:
This pr references a fix to displaying the lockup txID in #995 

Do we even need to display claim txID here ?


kingonly:
The fix is fine. Please merge and I'll test.

PR #1020, opened 2022-12-12
Add German translation (app_de.arb)

caloon:


ademar111190:
Thank you very much @caloon

PR #1022, opened 2022-12-13
Update app_de.arb

arbadacarbaYK:
There is no Bitcoins but bitcoin (Bitcoin with big B is the network not the currency) and some minor other things. Very good translation though!


JssDWt (approved):
Thanks Yvette! Amazing

PR #1024, opened 2022-12-13
Add German translation 🇩🇪🇦🇹🇨🇭🇱🇺🇱🇮🇧🇪

ademar111190:
Thank you @caloon and @arbadacarbaYK for the translations, feel free to make any changes to improve the translations 🙏

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,4 @@ ios/Flutter/flutter_export_environment.sh
 manual-build/
 ios/.symlinks/
 coverage/
+untranslated_messages.txt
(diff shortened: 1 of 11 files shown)


caloon:
Looking quite good. 

Two questions:
- The translation of "Connect to Pay" might be lacking misleading. Could you explain this feature so I can double-check if the translation fits?
- the app uses US date formats by default which might be confusing to international users because e.g. in German-speaking countries, you would use 24H time format and DD.MM.YYYY date format.

arbadacarbaYK:
Found a small error in one screen (seine Channel instead of ihre Channel). 

## Looking at the screen .. would be nice if the currency would be derived by the country vice versa (if only one uses it).. 😅 

Keep up the good work mates and congratz for the funding 🥳

ademar111190:
@arbadacarbaYK The euro currency is formatted with the locale specificities.

PR #1032, opened 2022-12-28
Uses guid instead of id to TLV

ademar111190:


roeierez (approved):

Alwin-Stockinger:
This change seems to have accidentally put a bug into the tlv record. Episode guid should be sent on the episode_guid field and itemID on the itemID field. But currently the episode_guid is sent on the itemID field.

PR #1035, opened 2023-01-05
Uses mobile scan instead of scan removing scan lib

ademar111190:
Scan lib was adding hms which contains a tracker, so I'm replacing it with mobile_scan https://github.com/flutter-package/flutter_scan/blob/main/android/build.gradle#L44 

As I was testing the scan flow, I also updated the lib.

As a bonus, as we are already using the mobile scanner, we are reducing the number of dependencies and the size of our app 🎉 

Scan made by the Classy Shark:

|Before|After|
|---|---|
|before|after|

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -31,7 +31,7 @@ android {
     defaultConfig {
         applicationId "com.breez.client"
         minSdkVersion 24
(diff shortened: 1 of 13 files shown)


roeierez (approved):
Great job!

PR #1046, opened 2023-01-19
Add divider back to network page

ademar111190:
A regression on the network page was made by removing the divider after the tor switch

Screenshot_20230120_085020


ademar111190:
> @ademar111190 please attach a screenshot because the font was smaller than it should've been.

Screenshot added.

I would ask you to test on your device after this change was merged because the font size on this page is the same as on the other pages.

roeierez (approved):

PR #1050, opened 2023-01-27
Pull clipboard fixes from c-breez

ademar111190:
Once the owner of the clipboard watcher accepts this PR https://github.com/leanflutter/clipboard_watcher/pull/8 I'll change the reference to the mainstream.


erdemyerebasmaz (approved):

lib/services/device.dart
```diff
@@ -9,34 +9,37 @@ import 'package:shared_preferences/shared_preferences.dart';
...
 
   Device() {
+    log.finest("Initing Device");
+
     var sharedPreferences = SharedPreferences.getInstance();
     sharedPreferences.then((preferences) {
-      _lastClipping =
-          preferences.getString(LAST_CLIPPING_PREFERENCES_KEY) ?? "";
+      _lastFromAppClip = preferences.getString(LAST_FROM_APP_CLIPPING_PREFERENCES_KEY);
+      _clipboardController.add(preferences.getString(LAST_CLIPPING_PREFERENCES_KEY) ?? "");
```
roeierez, line 34: @ademar111190  Why do we add to the clipboard stream the last clipped value from the persistent storage?

ademar111190: That is a great question, I haven't noticed before, but we always skip 1 because of that; I'll remove this logic because for me it looks like a residual thing from the previous clipboard logic.

ademar111190: I just removed it, but I'll add it back, I did find a flow where it is necessary; If you close the app and open it again with the same clipboard content you are going to trigger a flow that you already ran o the previous session.

ademar111190: I did add back, now that specific flow works as expected too

erdemyerebasmaz: Could you add comments for that case

> I just removed it, but I'll add it back, I did find a flow where it is necessary; If you close the app and open it again with the same clipboard content you are going to trigger a flow that you already ran o the previous session.

Could you add this case as a comment above skip logic?

ademar111190: Comment added

Issue #1057, opened 2023-02-05
Amount should be empty instead max on lnurl-withdraw 


kingonly:

kingonly:
@ubbabeck please take a look, DM me if needed. Make sure not to create regressions in the POS flow.

kingonly:
We should still set the amount when the min & max value are equal.

erdemyerebasmaz:
@ubbabeck See applyWithdrawFetchResponse and compare it with applyPayFetchResponse on how to leave it empty.
Afterwards, I suggest you replace the comparison logic on both instances by adding isFixedAmount variable to both WithdrawFetchResponse & PayFetchResponse for checking if min & max amounts of response are equal.

ubbabeck:
This issue has been resolved and merged to master. #1060

PR #1060, opened 2023-02-06
remove autofil lnurl-withdraw

ubbabeck:
# About pr

This pr refers to #1057 where the amount of lnurlwithdraw where filled by default. 
## What is done 
- remove the autofilling
- add  _isFixedAmount_  to class  _PayFetchResponse_ and _WithdrawFetchResponse_
- display invoice limit if min and max withdraw are not the equal

### Testing
Test where done with https://lnurl.fiatjaf.com/

### DIFF
--- a/lib/routes/create_invoice/create_invoice_page.dart
+++ b/lib/routes/create_invoice/create_invoice_page.dart
@@ -240,6 +240,21 @@ class CreateInvoicePageState extends State<CreateInvoicePage> {
                             validatorFn: validatePayment,
                             style: theme.FieldTextStyle.textStyle,
                           ),
+                          if (!_withdrawFetchResponse.isFixedAmount) ...[
+                            Padding(
+                              padding: const EdgeInsets.only(top: 8),
+                              child: Text(
+                                texts.lnurl_fetch_invoice_limit(
+                                  acc.currency
+                                      .format(_withdrawFetchResponse.minAmount),
+                                  acc.currency
+                                      .format(_withdrawFetchResponse.maxAmount),
+                                ),
+                                textAlign: TextAlign.left,
+                                style: theme.FieldTextStyle.labelStyle,
+                              ),
+                            )
+                          ],
                           _buildReceivableBTC(context, acc, lspStatus),
                           StreamBuilder<AccountModel>(
                             stream: accountBloc.accountStream,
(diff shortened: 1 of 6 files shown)


tlindi:
I like this 🙏

lib/routes/create_invoice/create_invoice_page.dart
```diff
@@ -467,11 +482,9 @@ class CreateInvoicePageState extends State<CreateInvoicePage> {
   ) {
     _withdrawFetchResponse = response;
     _descriptionController.text = response.defaultDescription;
-    _amountController.text = account.currency.format(
-      response.maxAmount,
-      includeDisplayName: false,
-      userInput: true,
-    );
+    if (response.isFixedAmount) {
+      _amountController.text = "${response.minAmount}";
```
erdemyerebasmaz, line 486: Why are we not using the same format here?

lib/routes/lnurl_fetch_invoice_page.dart
```diff
@@ -218,17 +218,19 @@ class LNURLFetchInvoicePageState extends State<LNURLFetchInvoicePage> {
...
+                          child: Text(
+                            texts.lnurl_fetch_invoice_limit(
+                              acc.currency.format(response.minAmount),
+                              acc.currency.format(response.maxAmount),
+                            ),
+                            textAlign: TextAlign.left,
+                            style: theme.FieldTextStyle.labelStyle,
+                          ),
+                        )
+                      : Padding(padding: const EdgeInsets.only(top: 8)),
```
erdemyerebasmaz, line 233: Is the padding necessary when the amount is fixed? If it's not necessary, I suggest to add this widget to the widget tree as such:
```suggestion
                    if (!response.isFixedAmount) ...[
                      Padding(
                           padding: const EdgeInsets.only(top: 8),
                           child: Text(
                             texts.lnurl_fetch_invoice_limit(
                               acc.currency.format(response.minAmount),
                               acc.currency.format(response.maxAmount),
                             ),
                             textAlign: TextAlign.left,
                             style: theme.FieldTextStyle.labelStyle,
                           ),
                         )
                    ]
```

lib/routes/create_invoice/create_invoice_page.dart
```diff
@@ -209,6 +209,8 @@ class CreateInvoicePageState extends State<CreateInvoicePage> {
                 return acc.validateIncomingPayment(amount);
               }
 
+              final response = widget.lnurlWithdraw;
```
erdemyerebasmaz, line 212: It looks like you can use _withdrawFetchResponse variable here directly as this value is assigned to it initially.

erdemyerebasmaz (approved):

PR #1065, opened 2023-02-10
Fetch c-breez log system

ademar111190:
I'm in a task where I need to read the breez mobile logs, and I realize how much easier it is to read the C-Breez logs, mainly because of the tags that allow us to focus, but also because of the LogCat integration.

Here on this PR, I'm fetching the c-breez structure and migrating the logs we have to use its structure.

A good bonus is that we can centralize it more easily when we start working in a shared shell between both apps.

[breez.logs.zip](https://github.com/breez/breezmobile/files/10767366/breez.logs.zip)
Screen Shot 2023-02-17 at 10 43 43

### DIFF
--- a/lib/logger.dart
+++ b/lib/logger.dart
@@ -1,59 +1,96 @@
 library breez.logger;
 
-import 'dart:async';
 import 'dart:io';
 
-import 'package:breez/services/breezlib/breez_bridge.dart';
-import 'package:breez/services/injector.dart';
-import 'package:breez_translations/breez_translations_locales.dart';
-import 'package:flutter/cupertino.dart';
-import 'package:logging/logging.dart';
+import 'package:archive/archive_io.dart';
+import 'package:fimber_io/fimber_io.dart';
+import 'package:flutter_fimber/flutter_fimber.dart';
+import 'package:flutter/foundation.dart';
+import 'package:path_provider/path_provider.dart';
 import 'package:share_plus/share_plus.dart';
 
-final Logger log = Logger('Breez');
+final _log = FimberLog("logger");
 
-Future<File> get _logFile async {
-  var logPath = await ServiceInjector().breezBridge.getLogPath();
-  return File(logPath);
-}
-
-void shareLog() {
-  _logFile.then((file) {
-    Share.shareFiles(
-      [file.path],
-      text: getSystemAppLocalizations().share_log_text,
-    );
-  });
-}
-
-Future shareFile(String filePath) {
-  return Share.shareFiles(
-    [filePath],
-    text: getSystemAppLocalizations().share_file_title,
-  );
+void shareLog() async {
+  final appDir = await getApplicationDocumentsDirectory();
+  final encoder = ZipFileEncoder();
+  final zipFilePath = "${appDir.path}/breez.logs.zip";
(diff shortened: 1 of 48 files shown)


roeierez:
@ademar111190 can you put here a sample log file so we can see how it looks like?

lib/logger.dart
```diff
@@ -1,59 +1,96 @@
...
+  }
 }
 
-class BreezLogger {
-  BreezLogger() {
-    BreezBridge breezBridge = ServiceInjector().breezBridge;
-    Logger.root.level = Level.ALL;
-    Logger.root.onRecord.listen((LogRecord rec) {
-      breezBridge.log(rec.message, rec.level.name);
-      print(rec.message);
```
erdemyerebasmaz, line 42: Since native loggers are no longer used I think they should be cleaned up.

ubbabeck (commented):
Looks good to me, will test as soon as possible.

ademar111190:
> @ademar111190 can you put here a sample log file so we can see how it looks like?

@roeierez I did add a screenshot of how it looks on LogCat and I did attach the zip file that share logs generate.

lib/logger.dart
ademar111190: I'm trying to understand which native loggers you are talking about @erdemyerebasmaz 

I'm not changing the java/swift or the golang logs here, only the flutter logs.

erdemyerebasmaz: Okay, I'm still trying to understand how our logger used to work and I may be wrong, so let me try to break things into parts.

First, the cleanup I was talking about:
```
 void log(String msg, String level) {
    _invokeMethodImmediate("log", {"msg": msg, "lvl": level});
  }
```
from BreezBridge is removed and I believe they were handled by:
`public void onMethodCall(MethodCall call, MethodChannel.Result result) {...}` of **Breez.java** 
&
 `func handle(_ call: FlutterMethodCall, result: @escaping FlutterResult) {...}` of **Breez.swift**.

Since `_breezLib.log()` is removed, their handlers should also be removed. Aren't they?
***
My other question is about logs from native classes such as **ChainSync.java**, **BackgroundTask.swift**.

I believe their logs were added to `{workingDir.path}/logs/bitcoin/mainnet/lnd.log` which was the file we used to share. These are no longer part of our logs, right?
Which I believe the same goes for Go library logs. They can be seen on the LogCat screenshot you've sent but won't be logged through Fimber and the shared log file will be missing native & library logs from how I see it.

Looks like it's been 4 years+ now but maybe @roeierez can clarify things as he was the one implemented logging.

ademar111190: The first point you are right, the log method can be safely deleted as we era not using it anymore.

Regarding the second question, the logs are split, there is the `lnd.log` which contains the go and native logs and the `breez.log` which contains the flutter logs, they show up on logcat together but right now they are being saved in different files.

erdemyerebasmaz: Let's log to a single file if we can. If all else fails we can zip and share both files.

ademar111190: I can just copy the c-breez strategy and broadcast the logs of the native side to flutter, but I need to change how the library works today.

PR #1067, opened 2023-02-15
Fix autosize text and extract number panel

ademar111190:
How it looks like

|Before|After|
|---|---|
|error|fixed|


ubbabeck (approved):
Looks good.  Minor remark I think you have a typo in the filename shouldn't it be `number_panel.dart` ?

Wait for roei or erdems approval before mergning 😄

erdemyerebasmaz (approved):

lib/routes/podcast/boost.dart
```diff
@@ -159,62 +178,6 @@ class BoostWidget extends StatelessWidget {
...
-              ),
-              AutoSizeText(
-                texts.podcast_boost_sats,
-                textAlign: TextAlign.center,
-                style: TextStyle(
-                  fontSize: 10,
-                  letterSpacing: 1,
-                ),
-                minFontSize: MinFontSize(context).minFontSize,
-                stepGranularity: 0.1,
```
erdemyerebasmaz, line 209: maxLines was missing here, was that the issue?

PR #1069, opened 2023-02-16
Revert from /dav to /webdav and error better

ubbabeck:
# About PR
Earlier I changed the attempting url from `/remote.php/ in this f128ee6f5c59f775a18484488e7841868c17749a commit.

This creates a successful sign-in, but the app is not able to locate or upload backup files successfully. Though this is not good it does at least not cause any issues in regards to us having to maintain the `remote.php/dav` url.

The app will first try the url the user inserted, but it this does not work it will attempt to try connect to the shorturl + remote.php/webdav.

## Depends

This pr depends on https://github.com/breez/breez/pull/182 

## Todo

- [x] Pass torconfig to the breez libary so we can restore from an instance behind tor.

## Test

- [x] with invalid url that is not a nextcloud or dav instance
- [x] restore from backup using short url
- [x] upload using short url 

### Test if regression
- [x] Test with google 
- [x] iCloud

@erdemyerebasmaz  if you have an old instance can you help with testing with an old instance of nextcloud ?
- [x] restore from old instance 
- [x] upload to old instance

### DIFF
--- a/lib/bloc/backup/backup_bloc.dart
+++ b/lib/bloc/backup/backup_bloc.dart
@@ -534,7 +534,7 @@ class BackupBloc {
       if (error is PlatformException) {
         var e = error;
         // Handle PlatformException(ResultError, "AuthError", Failed to invoke testBackupAuth, null)
-        switch (e.message) {
+        switch (e.code) {
           case _signInFailedCode:
             throw SignInFailedException(provider);
             break;
--- a/lib/bloc/backup/backup_bloc.dart
+++ b/lib/bloc/backup/backup_bloc.dart
@@ -23,8 +23,11 @@ import '../async_action.dart';
 import 'backup_actions.dart';
 
 class BackupBloc {
-  static const String _signInFailedCode = "AuthError";
-  static const String _notFoundCode = "NotFoundError";
+  static const String _signInFailedCode = "401";
+  static const String _methodNotFound = "405";
+  static const String _notFoundCode = "404";
+  static const String _noAccess = "403";
+  static const String _resultError = "ResultError";
   static const String USER_DETAILS_PREFERENCES_KEY = "BreezUserModel.userID";
 
   static const _kDefaultOverrideFee = false;
@@ -170,6 +173,7 @@ class BackupBloc {
       var map = settings.remoteServerAuthData.toJson();
       authData = json.encode(map);
     }
+
     await _breezLib.setBackupProvider(settings.backupProvider.name, authData);
   }
 
@@ -495,20 +499,25 @@ class BackupBloc {
           snapshots
               .sort((s1, s2) => s2.modifiedTime.compareTo(s1.modifiedTime));
           _multipleRestoreController.add(snapshots);
+          if (backups == null) {
+            throw NoBackupFoundException(
+                _backupSettingsController.value.backupProvider);
+          }
         }).catchError((error) {
           if (error.runtimeType == PlatformException) {
             PlatformException e = (error as PlatformException);
             if (e.code == _signInFailedCode || e.message == _signInFailedCode) {
               error = SignInFailedException(
                   _backupSettingsController.value.backupProvider);
+            } else if (e.message == _noAccess) {
+              error = NoBackupFoundException(
+                  _backupSettingsController.value.backupProvider);
             } else {
               error = (error as PlatformException).message;
             }
           }
-
           _restoreFinishedController.addError(error);
         });
-
         return;
       }
 
@@ -528,25 +537,33 @@ class BackupBloc {
   Future testAuth(BackupProvider provider, RemoteServerAuthData authData) {
     return _breezLib
         .testBackupAuth(provider.name, json.encode(authData.toJson()))
-        .catchError((error) {
-      log.info('backupBloc.testAuth caught error: $error');
+        .catchError(
+      (error) {
+        log.info('backupBloc.testAuth caught error: $error');
 
-      if (error is PlatformException) {
-        var e = error;
-        // Handle PlatformException(ResultError, "AuthError", Failed to invoke testBackupAuth, null)
-        switch (e.code) {
-          case _signInFailedCode:
-            throw SignInFailedException(provider);
-            break;
-
-          case _notFoundCode:
-            throw RemoteServerNotFoundException();
-            break;
+        if (error is PlatformException) {
+          var e = error;
+          // Handle PlatformException(ResultError, "AuthError", Failed to invoke testBackupAuth, null)
+          switch (e.message) {
+            case _signInFailedCode:
+              throw SignInFailedException(provider);
+              break;
+            case _methodNotFound:
+              throw MethodNotFoundException(provider);
+              break;
+            case _noAccess:
+              throw NoBackupFoundException(provider);
+              break;
+            case _notFoundCode:
+              throw RemoteServerNotFoundException();
+              break;
+            case _resultError:
+              throw NoBackupFoundException(provider);
+          }
         }
-      }
-
-      throw error;
-    });
+        throw error;
+      },
+    );
   }
 
   _restoreAppData() async {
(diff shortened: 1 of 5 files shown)


lib/bloc/backup/backup_bloc.dart
```diff
@@ -650,6 +667,24 @@ class SignInFailedException implements Exception {
...
+class MethodNotFoundException implements Exception {
+  final BackupProvider provider;
+  MethodNotFoundException(this.provider);
+
+  String toString() {
+    return "Method not found";
+  }
+}
+
+class NoBacupFoundException implements Exception {
```
erdemyerebasmaz, line 679: typo in name

lib/bloc/backup/backup_bloc.dart
```diff
@@ -23,8 +23,11 @@ import '../async_action.dart';
 import 'backup_actions.dart';
 
 class BackupBloc {
-  static const String _signInFailedCode = "AuthError";
-  static const String _notFoundCode = "NotFoundError";
+  static const String _signInFailedCode = "401";
+  static const String _methodNotFound = "405";
+  static const String _notFoundCode = "404";
+  static const String _noAccess = "403";
+  static const String _resultError = "ResultError";
```
erdemyerebasmaz, line 30: Is ResultError code of the error?

lib/routes/security_pin/remote_server_auth.dart
```diff
@@ -370,6 +376,10 @@ class RemoteServerAuthPageState extends State<RemoteServerAuthPage> {
       return DiscoverResult.INVALID_AUTH;
     } on RemoteServerNotFoundException {
       return DiscoverResult.INVALID_URL;
+    } on MethodNotFoundException {
+      return DiscoverResult.METHOD_NOT_FOUND;
+    } on NoBacupFoundException {
+      return DiscoverResult.INVALID_URL;
```
erdemyerebasmaz, line 382: Shouldn't this return DiscoverResult.BACKUP_NOT_FOUND?

lib/bloc/backup/backup_bloc.dart
ubbabeck: Yes, this was wrongly added as the ResultError is the code for the 404 errormessage. Will remove

erdemyerebasmaz:
@ubbabeck I was able to restore with https://nextcloud05.webo.cloud/ set as the url on a Backup that has the older, deprecated file structure and config.
Will need to wait some more time on testing upload as synchronizing to the BTC network is taking a lot of time, stuck at 1% for 5 minutes now. @roeierez

erdemyerebasmaz:
@ubbabeck Backup works too, tested w/ and w/o removing the `/remote.php/webdav/` suffix from the url

ubbabeck:
thank you  for your help @erdemyerebasmaz

lib/routes/security_pin/remote_server_auth.dart
```diff
@@ -148,33 +157,39 @@ class RemoteServerAuthPageState extends State<RemoteServerAuthPage> {
...
+                                  connectionWarningResponse = false;
+                                  Navigator.of(context).push(FadeInRoute(
+                                    builder: (_) =>
+                                        withBreezTheme(context, NetworkPage()),
+                                  ));
+                                  // Navigator.of(context).popUntil((route) => route is RemoteServerAuthPage);
+                                  return false;
+                                });
+                          } else if (widget._torBloc.torConfig != null) {
+                            await _breezLib
```
roeierez, line 190: Why do we need this line? Don't we require restart anyway?

roeierez (changes requested):

lib/routes/security_pin/remote_server_auth.dart
ubbabeck: removed since we moved the logic to accountbloc as you recommend.

lib/routes/security_pin/remote_server_auth.dart
```diff
@@ -370,8 +409,25 @@ class RemoteServerAuthPageState extends State<RemoteServerAuthPage> {
       return DiscoverResult.INVALID_AUTH;
     } on RemoteServerNotFoundException {
       return DiscoverResult.INVALID_URL;
+    } on MethodNotFoundException {
+      return DiscoverResult.METHOD_NOT_FOUND;
     }
 
+    // Since we are restoring we cannot be sure whenever the account has any
+    // backups.
+    if (widget.restore) {
```
roeierez, line 418: This code doesn't belong here. We have a flow for restore and no available backups the restore won't work anyway without this change.
If we want to add an additional error message we can do that as part of the restore flow.

lib/bloc/backup/backup_bloc.dart
```diff
@@ -528,25 +533,32 @@ class BackupBloc {
...
-          case _notFoundCode:
-            throw RemoteServerNotFoundException();
-            break;
+        .catchError(
+      (error) {
+        log.info('backupBloc.testAuth caught error: $error');
+
+        if (error is PlatformException) {
+          var e = error;
+          switch (e.message) {
```
erdemyerebasmaz, line 542: Shouldn't this be e.code?
I'm a bit confused by the usage of message & code overall. _signInFailedCode is message but has code in it's name etc. Please rename the variables with Code or Message suffix, in doing so these issues should resolve.

roeierez (changes requested):

lib/routes/security_pin/remote_server_auth.dart
ubbabeck: I have removed this logic as agreed.

lib/bloc/backup/backup_bloc.dart
ubbabeck: Updated the names, it is still a bit confusing, i added a comment.

ubbabeck:
The latest update is rebased to [update_deprecated_theme_values](https://github.com/breez/breezmobile/tree/update_deprecated_theme_values). 


0df8651a720b8308221c9233f93c8fdd7e7ffd79 It removed the Errormessage where the where prompted with an errormessage if no backups are available. It still shows that no backups are available, but goes back to startpage. This is related to https://github.com/breez/breezmobile/issues/1066 

0df8651a720b8308221c9233f93c8fdd7e  takes in consideration erdems comment about the usage of `code` and `message` which still is confusing. I added a comment on it.

ubbabeck:
When testing I discovered that if a user restores from a backup where tor has been active on a device where tor currently is inactive it tries to dial nodes without a the tor service active.

If tor is active from before this will not create any  issues.

We either need to:
- prompt for restart in order to start the tor service and get the config 
- avoid "activating" tor even though the backup had tor active.
- other suggestions 

Should we create a new issue for this or handle it in this pr ?

What do you think @roeierez @erdemyerebasmaz ?

> [INF] BTCN: DNS discovery failed on seed x49.seed.bitcoin.jonasschnelli.ch: dial tcp 127.0.0.1:0: connect: connection refused
>  BTCN: unable to lookup IP for bb1.breez.technology: dial tcp 127.0.0.1:0: connect: connection refused
> dial tcp 127.0.0.1:1080: connect: connection refused

erdemyerebasmaz:
> When testing I discovered that if a user restores from a backup where tor has been active on a device where tor currently is inactive it tries to dial nodes without a the tor service active.
> 
> If tor is active from before this will not create any issues.
> 
> We either need to:
> 
> * prompt for restart in order to start the tor service and get the config
> * avoid "activating" tor even though the backup had tor active.
> * other suggestions
> 
> Should we create a new issue for this or handle it in this pr ?
> 
> What do you think @roeierez @erdemyerebasmaz ?
> 
> > [INF] BTCN: DNS discovery failed on seed x49.seed.bitcoin.jonasschnelli.ch: dial tcp 127.0.0.1:0: connect: connection refused
> > BTCN: unable to lookup IP for bb1.breez.technology: dial tcp 127.0.0.1:0: connect: connection refused
> > dial tcp 127.0.0.1:1080: connect: connection refused

@ubbabeck What's the status on this? Is this answered through Slack?
I am not a big fan of restarts, so I'd suggest going with 2nd option.

Let's confirm there's no regression with iCloud and merge this PR first and create a new issue/PR for this.

PR #974, opened 2022-09-02
Update deprecated theme values

erdemyerebasmaz:
- Updated and utilize AppBarTheme,
  - Updated textTheme to titleTextStyle & toolbarTextStyle,
  - Use elevation, iconTheme, backgroundColor from AppBarTheme,
- Replaced BreezDatePicker with DatePicker,
  - Filter only applies if filter has changed,
- Removed buttonColor references,
- Updated primary attribute of buttons to background & foreground color,
- Removed unused InvoiceBottomSheet and it's assets,
-  #1072 
-  #1069
- Added Lightsats to vendors
- #1070

PR #1071, opened 2023-02-20
Fix NPE when accessing tag on a null action

ademar111190:
How it looks:

|Before|After|
|---|---|
|before|after|


erdemyerebasmaz:
Different approach but fixed here https://github.com/breez/breezmobile/commit/5aea75f3c8812056725f79ff521d6b52ce281940 as part of https://github.com/breez/breezmobile/pull/974 which we soon like to merge after running some tests.

PR #1073, opened 2023-02-20
Add utc flag to csv exporter and fixes tests

ademar111190:
A minor fix to the CSV exporter; setting the time to UTC so the tests will pass no matter the timezone of the machine running the tests, Keeping the default value to false to maintain the current behavior.


ubbabeck (commented):
Code: looks good to me.

Test: How may I test this in the app ?

ademar111190:
> Code: looks good to me.
> 
> Test: How may I test this in the app ?

Great question @ubbabeck does the following:

- Make a sale within POS (you can skip this step if you already have at least one sale in your history)
- Export the CSV file
- Check if the date is correct (i.e., it is showing the date on your timezone)

A second test you can do is just run the `flutter test`; no matter the timezone your machine is in, the tests should pass.

P.S. If the tests are broken because the DB tests try again running at one thread only, `flutter test --concurrency=1`.

ubbabeck (approved):
Tested again and all the test passed ✅ 

Also exported the CSV and saw now issues ✅

Issue #1076, opened 2023-03-01
Backup to another Android user account fails


tlindi:
How to reproduce:

Restore Breez with Google account say X. Settings Backup and Security shows backup account X information at bottom of screen.
Setup new google account Y to phone. Remove account X from phone.
Go to breez, select backup. And see it does nothing. Exclamation sign is shown on top right and it stays there. 
Settings Backup and Security shows no backup information at bottom of screen.

Add old account X back to phone. Breez creates backups nicely. But it doesn't ask which of two account I'd like to backup.

I have video and lnd.log if needed. Please advice how to deliver them to you securely.

kingonly:
@ademar111190 please investigate and follow up with @tlindi

PR #1078, opened 2023-03-02
Investigation backup bug

ademar111190:
Some improvements I made by investigating https://github.com/breez/breezmobile/issues/1076

The bug itself is not fixed; I'll need to think of a mechanism that contemplates the account, as currently, we differentiate based on the backup method only.

P.S. I did pull some improvements on the unit test from c-breez.


roeierez (approved):

roeierez:
> The bug itself is not fixed; I'll need to think of a mechanism that contemplates the account, as currently, we differentiate based on the backup method only.

We already have a mechanism where we get auth failure from the backup provider we force login (which should prompt the user to choose account).
Maybe we need to understand why this mechanism is not working at this case. @ademar111190 I suggest to take a look at backup_bloc.dart (search for the `_backupServiceNeedLogin` field).

PR #1080, opened 2023-03-06
Fix the broken simnet configuration

MrPeteypoo:
I encountered a bunch of issues when trying to run my own debug application with my own simnet. The modifications to the config files fix the simnet connection and the error handling in the QR code warning message fixes a nullptr crash when running on a simnet.


roeierez (approved):
LGTM

PR #1085, opened 2023-03-13
Fix multiple account backup

ademar111190:
My approach here is to trigger the sign again, which I believe is reasonable as we can know the failure is a [UserRecoverableAuthException](https://developers.google.com/android/reference/com/google/android/gms/auth/UserRecoverableAuthException)

Fixes https://github.com/breez/breezmobile/issues/1076

How it looks like: (the black screen is not visible IRL just on the video record, probably some protection made by google sign in SDK)

https://user-images.githubusercontent.com/1225438/224729722-660f4960-b0ec-409c-89e5-146dadfe393d.mp4

### DIFF
--- a/android/app/src/main/java/com/breez/client/plugins/breez/breezlib/GoogleAuthenticator.java
+++ b/android/app/src/main/java/com/breez/client/plugins/breez/breezlib/GoogleAuthenticator.java
@@ -1,16 +1,15 @@
 package com.breez.client.plugins.breez.breezlib;
 
 import android.content.Intent;
-import androidx.annotation.NonNull;
 import android.util.Log;
 
+import com.google.android.gms.auth.UserRecoverableAuthException;
 import com.google.android.gms.auth.api.signin.GoogleSignIn;
 import com.google.android.gms.auth.api.signin.GoogleSignInAccount;
 import com.google.android.gms.auth.api.signin.GoogleSignInClient;
 import com.google.android.gms.auth.api.signin.GoogleSignInOptions;
 import com.google.android.gms.common.api.ApiException;
 import com.google.android.gms.common.api.Scope;
-import com.google.android.gms.tasks.OnCompleteListener;
 import com.google.android.gms.tasks.Task;
 import com.google.android.gms.tasks.TaskCompletionSource;
 import com.google.android.gms.tasks.Tasks;
@@ -22,13 +21,13 @@ import java.util.Collections;
 import io.flutter.embedding.engine.plugins.activity.ActivityPluginBinding;
 import io.flutter.plugin.common.PluginRegistry;
 
-public class GoogleAuthenticator implements PluginRegistry.ActivityResultListener{
+public class GoogleAuthenticator implements PluginRegistry.ActivityResultListener {
     private static final String TAG = "BreezGAuthenticator";
     private static final int AUTHORIZE_ACTIVITY_REQUEST_CODE = 84;
 
     TaskCompletionSource<GoogleSignInAccount> m_signInProgressTask;
     private GoogleSignInClient m_signInClient;
-    private ActivityPluginBinding activityBinding;
+    private final ActivityPluginBinding activityBinding;
 
     public GoogleAuthenticator(ActivityPluginBinding binding) {
         activityBinding = binding;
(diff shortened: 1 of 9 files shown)


erdemyerebasmaz:
I've updated [google_api_availability](https://pub.dev/packages/google_api_availability/changelog#400) plugin and there are some changes that could be useful in these flows: https://github.com/Baseflow/flutter-google-api-availability/pull/29

erdemyerebasmaz (approved):
LGTM

android/app/src/main/java/com/breez/client/plugins/breez/breezlib/GoogleAuthenticator.java
```diff
@@ -59,28 +64,66 @@ public GoogleSignInAccount ensureSignedIn(final boolean silent) throws Exception
...
+        } catch (Exception e) {
+            Log.w(TAG, "setSelectedAccount failed", e);
+            throw e;
+        }
+
+        try {
+            return credential.getToken();
+        } catch (Exception e) {
+            Log.w(TAG, "getAccessToken failed", e);
+            if (e instanceof UserRecoverableAuthException) {
```
roeierez, line 92: This is a nice finding indeed.
This will trigger UI during backup right? We have a mechanism to trigger the sign in UI if we throw AuthError like here: https://github.com/breez/breezmobile/pull/1085/files#diff-5826ba9b9fbe1aedc7cb899ee752912c591e4de18fb85c442a8f86dca296413eR60

Perhaps it is better only to throw here AuthError and let the app layer do the sign in job as it does today?

ademar111190: Let me test if it works.

ademar111190: @roeierez I just tried this https://gist.github.com/ademar111190/e0ed3c45350bd9a4a23589522dcb83c6 but it doesn't worked, am I missing something?

roeierez:
Looking again I see that I confused this with a different flow so sorry for that.
My only concerns with the current implementation is that it seems that the native page for selecting account will open also during a regular backup flow that fails with that error and not only from the settings. Am I correct?
If I am wrong then the current implementation looks good to me.
If am correct maybe we should consider prompting anyway in the settings page? This can be easily done by changing the `backupNowSink` to receive a flag of "force login" and then pass it to the `_backupNow` method inside the bloc.

roeierez:
@ademar111190 Looking at the original reported issue I see it is about the alert icon in the home screen.
Did you test the behavior I recommended for that scenario? The expected behavior is that when pressing on the warning icon and trying to login it will "force" UI login and give the user a chance to change the account.

ademar111190:
Hi @roeierez I just tested and recorded this scenario; It looks correct to me; Maybe you prefer to show the warning instead of automatically triggering the account chooser, let me know if you prefer to change this behavior.

# Selecting an account on prompt

https://user-images.githubusercontent.com/1225438/226336283-4a14641f-ee49-4448-a117-d62b615f1eb1.mp4

# Skipping the account selection on prompt

https://user-images.githubusercontent.com/1225438/226336275-462c81e6-0e1c-45e8-be43-556dd32e66ee.mp4

roeierez:
Thanks for the details @ademar111190 
I think we shouldn't prompt automatically to choose account as it looks intrusive (Backups triggers are running multiple times in the background during payment).
I think if clicking on the warning icon actually prompt when needed it is prefect.

ademar111190:
> Thanks for the details @ademar111190 I think we shouldn't prompt automatically to choose account as it looks intrusive (Backups triggers are running multiple times in the background during payment). I think if clicking on the warning icon actually prompt when needed it is prefect.

That is something I need to investigate how to do, IDK if we have context on the native side to trigger or not the native dialog.

ademar111190:
# Update after the recoverable flag


How it looks at home page 


https://user-images.githubusercontent.com/1225438/227006797-050db88e-37d9-4261-bb7f-0133aa2fa28f.mp4


How it looks at backup page

https://user-images.githubusercontent.com/1225438/227006821-00a94dd7-1bf8-4972-ac2e-63eeb2d08ae7.mp4

roeierez (approved):
LGTM

tlindi:
btw I DM'd @ademar111190 couple of notes at TG. Likely nothing, but just to let you know...

Issue #1087, opened 2023-03-15
Payment higher than the limit message correction


Nackoo2000:
1. Click Send  >  Send to BTC Address
2. Enter the amount of 5 000 000 sats 
3. Click Next

Please remove "4 294 967 sats" from the error message. Approved by @kingonly.

![photo_2023-03-15_11-06-11](https://user-images.githubusercontent.com/127208515/225261115-36ceaa64-5773-4dac-a6ac-cb859820add7.jpg)

ubbabeck:
Do we wish to remove this both when creating an invoice and receiving ?

kingonly:
@ubbabeck for creating an invoice for this specific message.

PR #1090, opened 2023-03-15
Limit message correction

ubbabeck:
# About PR

This pr references issue  https://github.com/breez/breezmobile/issues/1087.

## What is done

- Removed displaying the _maxPaymentAmount_
- Removed unused function _validateOutgoingOnChainPayment_

### DIFF
--- a/lib/bloc/account/account_model.dart
+++ b/lib/bloc/account/account_model.dart
@@ -269,22 +269,23 @@ class AccountModel {
     SyncUIState syncUIState,
   }) {
     return AccountModel(
-        accountResponse ?? _accountResponse,
-        currency ?? this.currency,
-        fiatShortName ?? _fiatShortName,
-        fiatCurrency ?? _fiatCurrency,
-        fiatConversionList ?? _fiatConversionList,
-        preferredCurrencies ?? this.preferredCurrencies,
-        posCurrencyShortName ?? _posCurrencyShortName,
-        addedFundsReply: addedFundsReply ?? this.addedFundsReply,
-        onChainFeeRate: onChainFeeRate ?? this.onChainFeeRate,
-        enableInProgress: enableInProgress ?? this.enableInProgress,
-        syncProgress: syncProgress ?? this.syncProgress,
-        syncedToChain: syncedToChain ?? this.syncedToChain,
-        serverReady: serverReady ?? this.serverReady,
-        syncUIState: syncUIState ?? this.syncUIState,
-        initial: initial ?? this.initial,
-        nodeUpgrading: nodeUpgrading ?? this.nodeUpgrading);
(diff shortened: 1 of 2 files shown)


roeierez (approved):
LGTM

lib/bloc/account/account_model.dart
```diff
@@ -403,15 +394,11 @@ class AccountModel {
   String validatePayment(Int64 amount, bool outgoing) {
     final texts = getSystemAppLocalizations();
     if (maxPaymentAmount != null && amount > maxPaymentAmount) {
-      return texts.valid_payment_error_exceeds_the_limit(
-        currency.format(maxPaymentAmount),
-      );
+      return texts.valid_payment_error_exceeds_the_limit("");
```
ademar111190, line 397: There are better approaches than passing an empty string here.
Removing the variable from the text entry point, you can call the texts without arguments by calling `return texts.valid_payment_error_exceeds_the_limit`

ademar111190 (changes requested):

lib/bloc/account/account_model.dart
ubbabeck: Regarding you feedback,I was not able to pass a null param so I ended up with generating a new variable without a param.

Is it possible to generate translations that are optional? 

If it is fact is not possible please see this pr https://github.com/breez/Breez-Translations/pull/7

ademar111190: Reviewed; thanks for addressing the issue.

lib/bloc/account/account_model.dart
```diff
@@ -269,23 +269,22 @@ class AccountModel {
...
+        posCurrencyShortName ?? _posCurrencyShortName,
+        addedFundsReply: addedFundsReply ?? this.addedFundsReply,
+        onChainFeeRate: onChainFeeRate ?? this.onChainFeeRate,
+        enableInProgress: enableInProgress ?? this.enableInProgress,
+        syncProgress: syncProgress ?? this.syncProgress,
+        syncedToChain: syncedToChain ?? this.syncedToChain,
+        serverReady: serverReady ?? this.serverReady,
+        syncUIState: syncUIState ?? this.syncUIState,
+        initial: initial ?? this.initial,
+        nodeUpgrading: nodeUpgrading ?? this.nodeUpgrading);
```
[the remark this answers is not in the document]
ubbabeck: Added the trailing. Think the dart format is doing its own thing 😕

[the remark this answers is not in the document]
ademar111190: Yeap, it is okay to format the files we are working on, but as we have two valid formatted options here, the one that changes the code less is preferable IMHO; thanks for changing it.

ademar111190 (approved):

PR #1106, opened 2023-03-24
Add a recover flow when user has removed the permission of its account

ademar111190:
How it looks like:

https://user-images.githubusercontent.com/1225438/227606638-e936cfd8-7739-4d54-9d28-fba46d8cdb11.mp4


android/app/src/main/java/com/breez/client/plugins/breez/breezlib/GoogleAuthenticator.java
```diff
@@ -152,6 +157,105 @@ public boolean validateAccessTokenAllowingPrompt() {
...
+            } else {
+                Log.w(TAG, "verifyCredentialHasWriteAccess failed", e);
+                return false;
+            }
+        }
+
+        return true;
+    }
+
+    private boolean recoverUnauthorizedAccess() {
```
roeierez, line 194: The logic seems to work but the code seems a bit complicated, probably because of this duplication.
Let's think how to make it more compact and better reused.

ademar111190: I did compact the tries/catches and other minor improvements, let me know if you see more things that we can improve to make it more readable.

roeierez (approved):

PR #1113, opened 2023-04-04
Add sk translation 🇸🇰🇪🇺🇷🇸

ademar111190:
Thank you @CS-HODL for adding the translation 🙏


roeierez (approved):

CS-HODL:
Thanks for guiding me, Just started the CZ translation. It'll take a while. 

Is there any chance to add all languages into the Preferences? I would love to switch between languages and test them instead of setting the language for the phone and installing Breez again.

PR #1114, opened 2023-04-05
Workaround on decimal round bug

ademar111190:


lib/bloc/account/account_model.dart
```diff
@@ -404,6 +408,7 @@ class AccountModel {
 
     if (outgoing && amount > maxAllowedToPay) {
       if (reserveAmount > 0) {
+        return null; // TODO: remove this workaround when we have a solution to decimals round at our library
```
roeierez, line 411: @ademar111190 I think we intended to disable the validation only in the case the user marked "Use all funds" in the UI. This validation now is disabled for all cases.

roeierez (changes requested):

lib/bloc/account/account_model.dart
ademar111190: 🤦 true, sorry I forget it

roeierez (approved):

PR #1117, opened 2023-04-13
support in-app NTAG 4xx on iOS

JssDWt:
We got a report from a user that scanning an NTAG 424 DNA card on iOS doesn't do anything on iOS. It looks like we're missing the select-identifier for these tags in the Info.plist file.


roeierez (approved):
LGTM

erdemyerebasmaz (approved):

ubbabeck (approved):

Issue #1116, opened 2023-04-12
Backup: `Don't prompt again` not working 


MarnixCroes:
selecting `Don't prompt again` doesn't do/change anything

remove it (& always display warning triangle when no backup) or make it work :)
![IMG_20230412_103744](https://user-images.githubusercontent.com/93143998/231402055-df980de6-1a98-4b8f-b385-7767926c2e47.jpg)

MarnixCroes:
long time bug, I couldn't find any related GH issues

kingonly:
@ademar111190 please follow up

ademar111190:
Hi @MarnixCroes thank you for the report.
I couldn't reproduce the bug, is it possible for you to record the dialog being prompt when the option is enabled? That could help me a lot 🙏

MarnixCroes:
> Hi @MarnixCroes thank you for the report. I couldn't reproduce the bug, is it possible for you to record the dialog being prompt when the option is enabled? That could help me a lot pray

https://user-images.githubusercontent.com/93143998/231759592-599b213f-040b-4695-abb5-fd7a8258a3ae.mp4


@ademar111190 
FYI, the recording is on GrapheneOS, but it also happens on (Samsung) One UI.
It's not related that it's a new empty wallet, I created this one just for the recording, it also happens on long time used wallets with funds.

kingonly:
If you mean the warning icon doesn't go away - this is by design. The option is to disable the dialog prompt after payments.

MarnixCroes:
I see, makes sense thx, 
What made question is that:
When the `Don't prompt again` isn't check marked, there is no difference in behavior? (after restarting Breez)
or when is the prompt supposed to show up?

kingonly:
If the prompt is unchecked, there should be a promotion after a payment

MarnixCroes:
> If the prompt is unchecked, there should be a promotion after a payment

hmm, that doesn't happen 😅

kingonly:
> > If the prompt is unchecked, there should be a promotion after a payment
> 
> hmm, that doesn't happen 😅

Can you share a video?

MarnixCroes:
> Can you share a video?

Nothing happens:
- prompt is unchecked
- generate invoice
- `Payment received!`
- click to confirm animation

that's it, there is no prompt after (as supposed to?)

ademar111190:
Closing as it was fixed by #1192

PR #1121, opened 2023-04-17
Improve backup prompt logic and fixes concurrency bugs

ademar111190:
Behavior changes:
- Makes `promptBackupController` a behavior subject fixing https://github.com/breez/breezmobile/issues/1116
- Add a `promptBackupDismissedStream` to avoid multiple prompts in the same session
- Add a few logs to help bug reports
- Makes `pushPrompt` take into account only `_enableBackupPrompt`
- Makes `AccountRequiredActionsIndicator` reads its own blocs
- Unify streams on `AccountRequiredActionsIndicator` using `combineLatestList`


roeierez:
@ademar111190  can you please put some description on what was improved from the user perspective? I need that in order to better review the PR.

ademar111190:
> * behavior
Sure, it is about the second item, which I'll add details for more clarity:
- Add a promptBackupDismissedStream to avoid multiple prompts in the same session

The previous code will pop the warning at any time the subscriber receives an event, even after the users dismiss the dialog; it is pretty easy to see this behavior by changing from home to another area, such as podcast; you get the warning every time.
By adding the promptBackupDismissedStream and taking its value into account before showing the dialog, we are not bugging the user at every page navigation, we still show the prompt when the dismissed stream is false, and we will show the warning icon, but if the user had dismissed the dialog we don't show again until we clear the promptBackupDismissedStream.

Nackoo2000:
@ademar111190 so when fixed this should behave like this right:
First case
1. Click Don't prompt again
2. Receive a payment
3. The message does not appear

Second case
1. Receive a payment
2. The message appears

lib/routes/account_required_actions.dart
```diff
@@ -44,60 +39,61 @@ class AccountRequiredActionsIndicator extends StatefulWidget {
...
-        }
-      });
-      _init = true;
-    }
+    final backupBloc = AppBlocsProvider.of<BackupBloc>(context);
+    _promptBackupSubscription?.cancel();
+    _promptBackupSubscription = Rx.combineLatestList([
+      backupBloc.backupSettingsStream,
+      backupBloc.promptBackupStream,
+      backupBloc.promptBackupDismissedStream,
```
roeierez, line 53: It seems that we can encapsulate the promptBackupDismissedStream and not handle it in the UI.
How about only push to the promptBackupStream if really prompt is needed?

lib/bloc/backup/backup_bloc.dart
```diff
@@ -44,10 +44,12 @@ class BackupBloc {
       BehaviorSubject<BackupState>();
   Stream<BackupState> get backupStateStream => _backupStateController.stream;
 
-  final StreamController<bool> _promptBackupController =
-      StreamController<bool>.broadcast();
+  final StreamController<bool> _promptBackupController = BehaviorSubject<bool>.seeded(false);
   Stream<bool> get promptBackupStream => _promptBackupController.stream;
 
+  final StreamController<bool> _promptBackupDismissedController = BehaviorSubject<bool>.seeded(false);
```
roeierez, line 50: Should the behavior of "Don't prompt again" be persistent?

roeierez (changes requested):

lib/bloc/backup/backup_bloc.dart
ademar111190: It is to avoid https://github.com/breez/breezmobile/issues/1116, the same reason I'm changing line 47 too

kingonly:
Status @ademar111190 @roeierez?

ademar111190:
> Status @ademar111190 @roeierez?

Still waiting for @roeierez's review @kingonly

lib/bloc/backup/backup_bloc.dart
```diff
@@ -87,6 +89,19 @@ class BackupBloc {
...
   Sink<AsyncAction> get backupActionsSink => _backupActionsController.sink;
 
+  Stream<bool> get promptBackupSubscription => Rx.combineLatestList([
+        backupSettingsStream,
+        promptBackupStream,
+        promptBackupDismissedStream,
+      ]).where((list) {
+        final settings = list[0] as BackupSettings;
+        final dismissed = list[2] as bool;
+        return settings.promptOnError && !dismissed;
```
roeierez, line 99: NIT: We can add `&& needSignIn` here and get rid of the map.

lib/routes/account_required_actions.dart
```diff
@@ -44,60 +39,57 @@ class AccountRequiredActionsIndicator extends StatefulWidget {
...
-  StreamSubscription<BackupSettings> _settingsSubscription;
-  BackupSettings _currentSettings;
+  StreamSubscription<dynamic> _promptBackupSubscription;
   bool showingBackupDialog = false;
-  bool _init = false;
 
   @override
   void didChangeDependencies() {
     super.didChangeDependencies();
-    if (!_init) {
```
roeierez, line 56: why was this removed? `if (!_init)`
As far as I understand it protects this logic from running more than once.

roeierez (changes requested):
Looks good! some minor feedback.

lib/routes/account_required_actions.dart
ademar111190: Using an init flag inside the change dependencies method is an error; It lead to lifecycle bugs (I had solved several bugs like that on previous PRs)

If there is something we want to do exactly one time, we shall use `initState` 

Honestly, I don't see the need for but I can move this code to init state if you prefer @roeierez

lib/bloc/backup/backup_bloc.dart
ademar111190: Not possible, doing that, if `needSignIn` is false we are not emitting the event.

We need to receive the event whenever `needSignIn` is true or false; the rules to filter are the promptOnError and not dismissed.

lib/routes/account_required_actions.dart
roeierez: I think it would be best to use initState as I think right now we also have a bug since the `didChangeDependencies` can be invoked several times.

ademar111190: Ok, I'll move to initState; There is no bug right now because I'm cleaning the subscription properly; the only side effect is we are wasting some cpu time here.

ademar111190:
Closing as it will be merged on top of the refactoring; new PR here https://github.com/breez/breezmobile/pull/1192

PR #1132, opened 2023-04-26
Ensure latest backup

ubbabeck:
# About PR

This PR refers to https://github.com/breez/breez/pull/198. Where we have added a mechanism to check if we are running an old backup of Breez. If this is the fact we shut down the lndaemon and display an error message as shown underneath.  
ensure_latest_backup

## Question

At the moment I have made it so that the user only can exit the app.

Alternatively we let the user continue to use the app, with little to no functionality since the lndaemon is shutted down. But the user might wish to eg. export sales, listen to podcast etc.

What is the wished behaviour in this case @kingonly ?


kingonly:
"Breez has detected that the app is using an old backup. Please reinstall and restore from the latest backup available in xxx. Note that you won't be able to restore Breez without your encryption key."

@ubbabeck xxx=provider. Add the additional sentence only if the encryption key is enabled.

ubbabeck:
Added the error message in the latest commit.

## encrypted 
ensurelatest_encrypted
## not encrypted
 
ensurelatest

kingonly:
Use the display name of the provider...

ubbabeck:
> Use the display name of the provider...

Updated to the correct value now.  🙏

roeierez (approved):
LGTM

Nackoo2000:
Tested creating new accounts and backups in each option - iCloud, Google Cloud, and Nextcloud
Restored from each could with and without encryption.
Tested switching a device between the 3 clouds.
Tested with multiple devices that it recognizes the older backup and shows an error that there is a newer backup in the same cloud.
Tested with multiple devices that it recognizes the older backup and shows an error that there is a newer backup in the different clouds.
All the relevant tests were done on Android AND iOS devices.

Issue #1111, opened 2023-04-04
Accept comma in POS mode


ubbabeck:
Some android models does not show dot `.` on the keyboard type `TextInputType.numberWithOptions`, but rather comma `,` so the user are not able to use decimal when adding a new item in the POS mode.
![comma](https://user-images.githubusercontent.com/36157890/229713981-1d2d753d-d212-40f6-8054-e45ca4b80ebd.jpeg)

erdemyerebasmaz:
Tested this now and can't pinpoint the reason why but I was able to isolate the case where this behavior happens.

If you click on the price field after switching a currency there's only a comma on the keyboard, however, if you click any other text field then price field, the keyboard is displayed properly with both comma and a dot option again.

tlindi:
FYI Comma is used as a decimal separator on some regions, so it would be nice to functionality be consistant.

PR #1137, opened 2023-05-09
allow comma in POS mode #1111

ubbabeck:
# About PR
This issue references #1111.  Which is a known issue on Samsung phones when using [`TextInputType.numberWithOptions`](https://github.com/flutter/flutter/issues/61175)

I have allowed for the use of either comma or dot in the `whitelistedPattern` function. 

We then proceed with replacing all commas with dots in order to adhere to the `double.parse` function. 
## Question 
Are there another place in the app we wish to or should accepts comma too?

https://github.com/breez/breezmobile/assets/36157890/e160cbc4-a4c1-4418-a0a0-c28a4d1391a6

### DIFF
--- a/lib/bloc/account/fiat_conversion.dart
+++ b/lib/bloc/account/fiat_conversion.dart
@@ -27,7 +27,7 @@ class FiatConversion {
 
   RegExp get whitelistedPattern => currencyData.fractionSize == 0
       ? RegExp(r'\d+')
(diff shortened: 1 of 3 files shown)


erdemyerebasmaz (approved):
LGTM

ubbabeck:
> There are no other places where this changes is needed as far as I'm aware.

Great, thanks 🚀

PR #1140, opened 2023-05-16
Upgrade AGP from version 7.1.3 to 7.4.2

erdemyerebasmaz:
Updated Android Gradle Plugin to resolve build issues from Android Studio up to latest Electric Eel version.
- Target Flutter 3.7.12 when building from CI
- Add Flutter.podspec to .gitignore
- Update dependencies to latest
  - forked md5_file_checksum and made it compatible with bundled Kotlin version
- Update minSdkVersion to 28
  - path_provider_android requirement

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -30,7 +30,7 @@ android {
 
     defaultConfig {
         applicationId "com.breez.client"
-        minSdkVersion 24
+        minSdkVersion 28
         targetSdkVersion 33
         multiDexEnabled true
         versionCode 1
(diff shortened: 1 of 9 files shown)


roeierez (approved):

erdemyerebasmaz:
@ubbabeck @roeierez It would be great if you could check if you can build this branch locally.

android/app/build.gradle
```diff
@@ -30,7 +30,7 @@ android {
 
     defaultConfig {
         applicationId "com.breez.client"
-        minSdkVersion 24
+        minSdkVersion 28
```
ademar111190, line 33: @erdemyerebasmaz @roeierez The minimum Android version required from Nougat to Pie. Have you had a chance to talk about that?

ademar111190: I did check the path provider, and it uses min SDK to 16 https://github.com/flutter/packages/blob/main/packages/path_provider/path_provider_android/android/build.gradle#L32 What makes sense; there is nothing special about its functionality to require API 28

erdemyerebasmaz: @ademar111190 Good points, most of the flutter plugins's minSDKVersion was set to flutter.minSDKVersion recently but even so, flutter.minSDKVersion is not 28 yet on Flutter 3.10. This was likely not a cache issue on my end from downgrading. (Still had the same error after repairing cache)

Let's discuss this and revert if possible on another issue.

PR #1141, opened 2023-05-17
Integrate dynamic fees

erdemyerebasmaz:
This PR introduces dynamic fee changes.

#### Changelist
- Deprecated LSPInfo.channelMinimumFeeMsat with respective OpeningFeeParams's minMsat.
- Deprecated LSPInfo.channelFeePermyriad with respective OpeningFeeParams's proportional.
  - Older values are multiplied by 100 to reflect permyriad(per ten thousand) to PPM(Parts per million) change
- CheapestOpeningFeeParams is used for invoice request while LongestValidOpeningFeeParams are used for swap requests to calculate their fees.

#### Other changes
- Added README on how to symlink messages.proto & updated instructions on how to generate it
- Removed rpc.proto artifacts
- Updated dependencies to latest
- Minor UI fix
  - Image type in asset path incompatible with widget(which caused a benign null exception on every startup)

### DIFF
--- a/lib/bloc/account/account_model.dart
+++ b/lib/bloc/account/account_model.dart
@@ -7,6 +7,7 @@ import 'package:breez/bloc/user_profile/currency.dart';
 import 'package:breez/logger.dart';
 import 'package:breez/services/breezlib/data/messages.pb.dart';
 import 'package:breez/utils/date.dart';
+import 'package:breez/utils/exceptions.dart';
 import 'package:breez_translations/breez_translations_locales.dart';
 import 'package:fixnum/fixnum.dart';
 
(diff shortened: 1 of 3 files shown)


JssDWt (commented):
The proto doesn't really show this (proto dropped support for 'optional' fields), but these fields can be null if the LSP doesn't support opening channels. Breez library doesn't assume these fields exist, so I'm afraid this might lead to crashes if we ever decide to remove all opening_fee_params from the opening_fee_param_menu.
Maybe making them nullable here would propagate the null checks throughout the rest of the code?

This comment was supposed to be a review on these lines:
https://github.com/breez/breezmobile/pull/1141/files#diff-239e63a30093bdad70519ef0930718a3eda97d7d861db3c89cb985e95ab1de65R57-R60

PR #1159, opened 2023-06-07
check that nodeID can be parsed to hex

ubbabeck:
# About PR

This pr check if we are able to parse the given nodeID to hex in order to avoid parsing lighting addresses which can be the same length as nodeID.

### DIFF
--- a/lib/utils/node_id.dart
+++ b/lib/utils/node_id.dart
@@ -1,3 +1,6 @@
+import 'package:convert/convert.dart';
+import 'package:hex/hex.dart';
+
 const NODE_URI_SEPARATOR = "@";
 const NODE_ID_LENGTH = 66;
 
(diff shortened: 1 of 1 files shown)


erdemyerebasmaz (approved):
LGTM

lib/utils/node_id.dart
```diff
@@ -5,15 +5,29 @@ String parseNodeId(String nodeID) {
...
+    nodeID = nodeID.substring(0, NODE_ID_LENGTH);
+    if (_isParsable(nodeID)) {
+      return nodeID;
+    }
   }
-
   return null;
 }
+
+bool _isParsable(String nodeID) {
```
roeierez, line 25: Perhaps better to use the` hex.Decode `?

roeierez (changes requested):

lib/utils/node_id.dart
ubbabeck: Good point, changed it to use hex.decode in the latest rebase.

PR #1166, opened 2023-06-20
Nostr nip07

sarthak13gupta:
Implementing NIP-07 and supporting the window.nostr object, Breez would be able to host nostr web client Snort.
Supporting these two methods:
async window.nostr.getPublicKey(): string // returns a public key as hex
async window.nostr.signEvent(event: Event): Event // takes an event object, adds id, pubkey and sig and returns it

added maven endpoint in android/build.gradle


lib/bloc/nostr/nostr_bloc.dart
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  // Methods to simulate the actual logic
+
+  Future<String> _fetchPublicKey() async {
+    // check if key pair already exists otherwise generate it
+
+    if (nostrPublicKey == null) {
+      final keyGenerator = KeyApi();
+      nostrPrivateKey = keyGenerator.generatePrivateKey();
```
roeierez, line 94: This key generator is not reproducible if the user restores her wallet.
If we want to make it reproducible we need to save it in the breez.db (on the golang side).

roeierez (changes requested):

roeierez (changes requested):
Looks good!

android/build.gradle
```diff
@@ -15,7 +15,9 @@ allprojects {
     repositories {
         google()
         mavenCentral()
-        maven { url "https://raw.githubusercontent.com/guardianproject/gpmaven/master" }
+        maven { url "https://raw.githubusercontent.com/guardianproject/gpmaven/master" 
+                url 'https://storage.googleapis.com/download.flutter.io'
```
ademar111190, line 19: Why is it necessary? And why is it not set at its own `maaven` body?

android/app/build.gradle
```diff
@@ -113,9 +113,11 @@ dependencies {
     implementation 'com.google.code.gson:gson:2.10.1'
     implementation 'com.theartofdev.edmodo:android-image-cropper:2.8.0'
 
-    implementation platform('com.google.firebase:firebase-bom:29.3.1')
+    implementation platform('com.google.firebase:firebase-bom:31.5.0')
     implementation 'com.google.firebase:firebase-messaging'
 
+    implementation 'com.google.firebase:firebase-analytics-ktx'
```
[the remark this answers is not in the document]
ademar111190: I think it is a block as it add some trackers we already deleted in the past

ios/Flutter/Flutter.podspec
```diff
@@ -1,18 +0,0 @@
-#
-# NOTE: This podspec is NOT to be published. It is only used as a local source!
-#       This is a generated file; do not edit or check into version control.
-#
-
-Pod::Spec.new do |s|
```
ademar111190, line 6: Why is it being deleted?

lib/bloc/marketplace/marketplace_bloc.dart
```diff
@@ -8,13 +8,25 @@ class MarketplaceBloc {
   final _vendorController = BehaviorSubject<List<VendorModel>>();
   Stream<List<VendorModel>> get vendorsStream => _vendorController.stream;
 
+  bool isVendorToggle = true;
```
ademar111190, line 11: Why are you adding this toggle for snort?

lib/bloc/nostr/nostr_bloc.dart
sarthak13gupta: Okay, so that means **key pairs** once generated should be saved in breez.db right?

lib/bloc/nostr/nostr_bloc.dart
```diff
@@ -0,0 +1,168 @@
...
+    } else {
+      event.id = eventObject['id'];
+    }
+
+    event.sig = eventApi.signEvent(event, nostrPrivateKey);
+
+    if (eventApi.verifySignature(event)) {
+      eventObject['sig'] = event.sig;
+    }
+    await Future.delayed(const Duration(seconds: 1));
```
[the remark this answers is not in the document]
sarthak13gupta: It's not necessary, used to give a delay between exiting the nostr_sign_event dialog box and the respective event to take effect on snort.

android/build.gradle
sarthak13gupta: Was facing this issue when I was trying to build the app and not sure but it seems that some countries block some Google APIs as I found the solution over [here](https://stackoverflow.com/questions/59663760/flutter-gradle-not-working-except-on-https)
Screenshot 2023-06-11 at 6 23 19 PM

ios/Flutter/Flutter.podspec
sarthak13gupta: I must have messed up with .gitignore. I have added it to the new commit.

lib/bloc/nostr/nostr_bloc.dart
```diff
@@ -0,0 +1,174 @@
...
+    return eventObject;
+  }
+
+  Future<List<String>> _fetchRelays() async {
+    return ['Relay1', 'Relay2', 'Relay3'];
+  }
+
+  Future<String> _encryptData(String data, String publicKey) async {
+    // Simulating an encryption operation
+    return 'EncryptedData';
```
roeierez, line 160: Is this mean we don't support encrypt/decrypt? In that case isn't it better to throw error?

roeierez (changes requested):

lib/bloc/nostr/nostr_bloc.dart
sarthak13gupta: I've made the required changes in the new commit.

lib/bloc/marketplace/marketplace_bloc.dart
erdemyerebasmaz: Can we move this option to `MarketplaceBloc`? We're using BLoC pattern on breezmobile and this and it's related logic belongs in `MarketplaceBloc` imo.

We could create a `NostrSettings`* class such as:
*: @roeierez or `MarketplaceSettings` ?


```
class NostrSettings {
  final bool showSnort;

  NostrSettings({this.showSnort = true});

  NostrSettings.start() : this(showSnort: true);
  
  NostrSettings copyWith({
    bool showSnort,
  }) {
    return NostrSettings(
      showSnort: showSnort ?? this.showSnort,
    );
  }

  NostrSettings.fromJson(Map<String, dynamic> json)
      : this(showSnort: json["showSnort"] ?? true);

  Map<String, dynamic> toJson() => {
        "showSnort": showSnort,
      };
}
```

_(in case we want to change default behavior or include user preference in backup and one other use case I'll explain on a later step)_
Then later we can expose it from `MarketplaceBloc` via `Stream`/`Sink`:

```
final _nostrSettingsController = BehaviorSubject<NostrSettings>();
Stream<NostrSettings> get nostrSettingsStream =>
    _nostrSettingsController.stream;
Sink<NostrSettings> get nostrSettingsSettingsSink =>
    _nostrSettingsController.sink;
```

to be able to toggle between show/hide.

```
void _toggleSnort(
    MarketplaceBloc bloc,
    NostrSettings nostrSettings,
    ) {
  bloc.nostrSettingsSettingsSink.add(
    nostrSettings.copyWith(
      showSnort: !nostrSettings.showSnort,
    ),
  );
}

```

Each time this setting changes we can update the `NostrSettings` and then reload vendors on the `MarketplaceBloc` by listening to `nostrSettingsStream` such as:

```
  MarketplaceBloc() {
    loadVendors();
    _listenNostrSettings();
  }
  
   _listenNostrSettings() async {
    nostrSettingsStream.listen((settings) async {
      _nostrSettingsController.add(settings);
      await loadVendors();
    });
  }
```

lib/routes/marketplace/nostrEvent_handlers.dart
```diff
@@ -0,0 +1,145 @@
...
+      if (!shouldReturnPubkey) {
+        return "";
+      }
+    }
+
+    _nostrBloc.actionsSink.add(GetPublicKey());
+
+    final publicKey = await _nostrBloc.publicKeyStream.first;
+
+    return publicKey;
```
erdemyerebasmaz, line 71:
```suggestion
    return await _nostrBloc.publicKeyStream.first;;
```

roeierez (approved):
LGTM

lib/widgets/nostr_sign_event_dialog.dart
```diff
@@ -0,0 +1,120 @@
...
+        title: Text(
+          widget.appName,
+          style: themeData.dialogTheme.titleTextStyle,
+        ),
+        contentPadding: const EdgeInsets.fromLTRB(8.0, 8.0, 8.0, 24.0),
+        content: SingleChildScrollView(
+          child: Column(
+            mainAxisSize: MainAxisSize.min,
+            crossAxisAlignment: CrossAxisAlignment.stretch,
+            children: _getContent(context),
```
erdemyerebasmaz, line 55: > We don't really need a helper method for this, the widget in `_getContent` can be moved here or extracted as a widget, `_NostrGetPublicKeyDialogContent`.
> 
> We're trying to avoid and replace existing helper methods in our codebase with Widgets where possible. Explanation [here](https://www.youtube.com/watch?v=IOyq-eTRhvo).


Same here.

erdemyerebasmaz (commented):
Thanks @sarthak13gupta for contributing, these changes look good!
I have some small nits and comment on something we'll need to discuss internally regarding using BLoC pattern to toggle Snort.

lib/bloc/marketplace/marketplace_bloc.dart
```diff
@@ -8,13 +8,25 @@ class MarketplaceBloc {
...
+
   MarketplaceBloc() {
     loadVendors();
   }
 
   Future loadVendors() async {
     String jsonVendors = await rootBundle.loadString('src/json/vendors.json');
-    _vendorController.add(vendorListFromJson(jsonVendors));
+    // _vendorController.add(vendorListFromJson(jsonVendors));
```
erdemyerebasmaz, line 19: Remove commented lines

lib/routes/dev/dev.dart
```diff
@@ -606,12 +616,12 @@ class DevViewState extends State<DevView> {
...
     choices.add(Choice(
-      title: marketplaceBloc.isVendorToggle == true
-          ? 'Display Snort'
-          : 'Hide Snort',
+      title: "${marketplaceBloc.isSnortToggle ? "Display" : "Hide"} Snort",
       icon: Icons.phone_android,
       function: () {
-        _toggleVendor(marketplaceBloc);
+        // NostrSettings nostrSettings = NostrSettings();
+        // the instance of nostrSettings is null correct it
```
erdemyerebasmaz, line 623: Isn't this addressed? It looks like you do set the default value as `NostrSettings.start()` when initializing `MarketplaceBloc`.

PR #1168, opened 2023-07-09
Do not remove trailing zeros for BTC values

erdemyerebasmaz:
Trailing zeros are no longer removed from BTC values.

Todo:
- Remove instances where format is called for Sat/BTC values with removeTrailingZeros: true.
- Test if removing trailing zeros break anything with payment flows.


MarnixCroes (approved):
utACK

ubbabeck (approved):
Look good to me 🙌 
Tested on android emulator and the latest TestFlight version too.

ubbabeck:
The POS looks good. 
tested a couple of sales, the transactions also looks good.

PR #1173, opened 2023-07-20
check for bip21 on validating address

ubbabeck:
This enables extracting bitcoin address from bip21 URI. 

Also introduces propperhandling for native segwit and legacy addresses if they are prefixed with **bitcoin:**.

### DIFF
--- a/lib/services/breezlib/breez_bridge.dart
+++ b/lib/services/breezlib/breez_bridge.dart
@@ -775,18 +775,9 @@ class BreezBridge {
   }
 
   Future<String> validateAddress(String address) {
-    String addr = address;
-    if (address == null) {
-      return Future.error("empty address");
-    }
-    addr = extractBitcoinAddressFromBip21(address) ?? addr;
-
-    if (addr.startsWith("bitcoin:") || addr.startsWith("BITCOIN:")) {
-      addr = addr.substring(8);
-      isLegacyOrNestedSegwit(addr) ? addr : addr.toLowerCase();
-    }
-    return _invokeMethodWhenReady("validateAddress", {"argument": addr})
-        .then((response) => addr);
+    address = extractBitcoinAddress(address);
+    return _invokeMethodWhenReady("validateAddress", {"argument": address})
+        .then((response) => address);
   }
 
   Future<Int64> getDefaultOnChainFeeRate() {
--- a/lib/services/breezlib/breez_bridge.dart
+++ b/lib/services/breezlib/breez_bridge.dart
@@ -6,6 +6,7 @@ import 'package:breez/bloc/lnurl/lnurl_model.dart';
 import 'package:breez/logger.dart' as logger;
 import 'package:breez/services/breezlib/data/messages.pb.dart';
 import 'package:breez/services/download_manager.dart';
+import 'package:breez/utils/bip21.dart';
 import 'package:dio/dio.dart';
 import 'package:fixnum/fixnum.dart';
 import 'package:flutter/services.dart';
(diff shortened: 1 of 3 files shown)


lib/services/breezlib/breez_bridge.dart
```diff
@@ -778,8 +779,11 @@ class BreezBridge {
     if (address == null) {
       return Future.error("empty address");
     }
-    if (addr.toLowerCase().startsWith("bitcoin:")) {
+    addr = extractBitcoinAddressFromBip21(address) ?? addr;
```
erdemyerebasmaz, line 782: Looks like we can reduce this whole logic into one function called `extractBitcoinFromAddress(String address)`. I can't really tell the difference between these two as `split` handles the case where `?` is non-existent. 

```
static const String URN_SCHEME = "bitcoin:";

String extractBitcoinAddress(String address) {
    if (address.toLowerCase().startsWith(URN_SCHEME)) {
      try {
        int split = address.indexOf("?");
        String addr =
            address.substring(URN_SCHEME.length, split == -1 ? null : split);
        if (addr.isNotEmpty) {
          isLegacyOrNestedSegwit(addr) ? addr : addr = addr.toLowerCase();
          return addr;
        }
      } on FormatException {
        // do nothing.
      }
    }
    return address;
  }
```
and use it in validateAddress as:

```
Future<String> validateAddress(String address) {
    if (address == null) {
      return Future.error("empty address");
    }
    return _invokeMethodWhenReady("validateAddress", {
      "argument": extractBitcoinAddress(address),
    }).then((address) => address as String);
  }
```

lib/services/breezlib/breez_bridge.dart
```diff
@@ -778,8 +779,11 @@ class BreezBridge {
...
+
+    if (addr.startsWith("bitcoin:") || addr.startsWith("BITCOIN:")) {
```
erdemyerebasmaz, line 784: Why is there a need to change `addr.toLowerCase().startsWith("bitcoin:")` to `(addr.startsWith("bitcoin:") || addr.startsWith("BITCOIN:")`?

erdemyerebasmaz (changes requested):

lib/services/breezlib/breez_bridge.dart
ubbabeck: If the user inserts a bitcoin address which is case sensitive Legacy and Nested Segwit we cannot change it toLowerCase. The address will then produce an invalid checksum since they are base58 encoded.

ubbabeck: Acording to the bip21 spec the scheme component is insensitive so me must be able to accept both upper and lower case. What I can add is better handling of this.

> The scheme component ("bitcoin:") is case-insensitive, and implementations must accept any combination of uppercase and lowercase letters. The rest of the URI is case-sensitive, including the query parameter keys.

https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki

erdemyerebasmaz: We're not making any changes to the variable value here, the address string is **immutable** this is only used for comparison.

```
String tmp = "ABcD";
tmp.toLowercase();
print(tmp); // Print's "ABcD";
tmp = tmp.toLowercase(); 
print(tmp); // Print's "abcd";
```

lib/services/breezlib/breez_bridge.dart
ubbabeck: Thanks good idea! Fixed

erdemyerebasmaz (approved):
Looks good!

PR #1177, opened 2023-07-29
Comments platform

sarthak13gupta:
**changes to support nip07 methods -> signEvent and getPubKey on anytime podcast player comments platform**

*added listener to **enable/disable nostr** which will toggle **snort** and **comments platform** on **anytime podcast player**.
*listen to calls for **signEvent** and **getPubKey** and return the result to **anytime podcast player**.


kingonly:
Please provide feedback @roeierez @erdemyerebasmaz

PR #1181, opened 2023-08-02
update Breez-Translations

### LINKED ISSUE #1179: Don't display the fee free amount if it's bigger than the max amount
Remove the "a setup fee..." sentence in this case.

ubbabeck:
fixes #1179

### DIFF
--- a/lib/routes/add_funds/deposit_to_btc_address_page.dart
+++ b/lib/routes/add_funds/deposit_to_btc_address_page.dart
@@ -181,8 +181,12 @@ class DepositToBTCAddressPageState extends State<DepositToBTCAddressPage> {
     final liquidity = accountModel.currency.format(
       connected ? accountModel.maxInboundLiquidity : Int64(0),
     );
+    final bool showFeeMessage =
+        maxAllowedDeposit > accountModel.maxInboundLiquidity;
 
-    if (connected && showMinFeeMessage) {
+    if (!showFeeMessage) {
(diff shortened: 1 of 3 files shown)


ademar111190:
When changing translation entries that add or remove variables, it is preferable to create a new entry than modifying existing ones, as this translation repo is being used in more than one project and sometimes in different branchs…

e.g.: An entry exists like this one `”sum_result”: “The sum of {a} + {b} = {c}”` and you want to remove entries, instead of modifying “sum_result” it is safer adding a new one such as `”sum_result_2”: “The sum result is {c}”`

lib/routes/add_funds/deposit_to_btc_address_page.dart
```diff
@@ -181,8 +181,11 @@ class DepositToBTCAddressPageState extends State<DepositToBTCAddressPage> {
     final liquidity = accountModel.currency.format(
       connected ? accountModel.maxInboundLiquidity : Int64(0),
     );
+    final bool showFee = maxAllowedDeposit <= accountModel.maxInboundLiquidity;
 
-    if (connected && showMinFeeMessage) {
+    if (showFee) {
+      return texts.invoice_btc_address_channel_not_needed(minSats, maxSats);
```
erdemyerebasmaz, line 187: Perhaps rename this variable showFeeMessage and use it as
```suggestion
    if (!showFeeMessage) {
      return texts.invoice_btc_address_channel_not_needed(minSats, maxSats);
```

erdemyerebasmaz (approved):

PR #1183, opened 2023-08-02
Refactor Initial Walkthrough restore flows

erdemyerebasmaz:
This PR fixes many issues surrounding restore backup.

#### Some background information:
App was getting stuck on `"Restoring data..."` loader after entering a valid but incorrect mnemonic phrase because the error handler wasn't working as expected. 

There was likely a change in Dart which deprecated this syntax:
`.catchError(_restoreFinishedController.addError));`

This has been replaced with:
```
.catchError((error) {
  _restoreFinishedController.addError(error);
}),
```
and it fails as expected and loader is popped.

---

The error message for incorrect mnemonic phrase which was illegible:

> FileSystemException: Failed to decode data using encoding 'utf-8', path = '/var/mobile/Containers/Data/Application/0866BE9F-3585-425F-B0D0-1013490F027C/Documents/app_data_backup/lightningFees.txt'

is now being replaced with:

> Failed to restore from backup. Please make sure backup phrase was correctly entered and try again.

---

**Solving** this wasn't enough because after a restore fails, we have many weird UI issues.
- Enter Backup Phrase route is multiplied
- Loader is still visible after going back to initial walkthrough page

These are some of them, solving them lead to more issues on most occasions. 

After fixing the loader issue, attempting to restore again throws either
```
flutter: Another exception was thrown: 'package:flutter/src/widgets/navigator.dart': Failed assertion: line 1995 pos 12: '!_debugLocked': is not true. 
```
 or
 ```
BackupProvider is not a subtype of type 'String?' of 'result'
```

- All loader logic on `InitialWalkthroughPage` and it's flows are replaced with contextless spinner(via `flutter_easyloading` plugin)
- All UI flows did live on `InitialWalkthroughPage` route and navigations were mostly handled by streams, which was not great as many things could, and were going wrong here. 
  - Widget-to-Widget communication is improved & redirections are now handled through `RestoreRequestHandler` & `BackupRestoreHandler` that make use of 
https://github.com/breez/c-breez/pull/586

to avoid UI locks. 10525b52d9d2438523be94227d19898d5504c50b

---

**BackupSettings** being persisted and the app data not being cleared after a failed restore attempt brought their own cascade of issues. These newer initialization logics seem to be implemented with only the successful case in mind.

App data is now being cleared before & on errors during backup restoration 3e231341db23010f62a904a2cba98d77f485d44d

BackupSettings is no longer used to remember last selected provider. Downside of this was it automatically picks the provider you've selected before, even after restart, you need a fresh install to change providers. This would mean if you've picked Google Drive by mistake, you'd be stuck with that option until a fresh install. Now you're able to switch providers on demand w/o requiring a fresh install.

---

**Switching** providers became problematic after this change(_or they already were_). BackupProvider settings were not being updated properly and the native implementation would try to sign in with the older settings.
A use case:
- Attempt restore with Remote Server backup
- Fill credentials and then change your mind to restore with Google Drive
- Selecting Google Drive would try to sign in with Remote Server settings
  - It would send a WebDav request instead of trying to sign in with Google

BackupProvider settings are now always being updated.

---

There was a regression where backup was not being triggered after
- encrypting cloud backup,
- disabling encryption.

This has been resolved w/ 157c57d700fc1068b22456ba518874ff22c5d233

---

#### Changelist:
- Removal of auto-restore when the snapshot list only has one item. 96dec7cf17257fb5b3e6f8503ddcf86e30e2f5e9
  - This prevented user to verify their backup visually by node id / time and isn't really helpful on support cases where we need users to download & share a snapshot.
- Added a "Sign into Google Drive failed." flushbar
- Update lastBackupTime of BackupState with modified time of restored snapshot
   - So we have latest backup information on Security & Backup settings page
#### Other Changelist:
- Launch links on external browser 9de62a6cc58193c9ad8ee5304fbe4f8f1c41352c

---

### TODO:
- Prevent problems caused by persisting BackupSettings. Which requires some changes on EnableBackupDialog(the dialog that opens up on clicking backup indicator on home page).
  - Let's say you picked Google Drive but decided not to restore and create a new account using Let's Breez.
The GDrive selection is persisted and clicking Backup Now through EnableBackupDialog forces you to login with GDrive. Unless you have an active backup, the Backup Now should behave similarly to selecting provider on Security & Backup settings page.

I'm also trying to
- understand the necessity of BackupSetting & BackupContext classes
- reproduce how we can end up with a Backup that has no node data

### DIFF
--- a/android/app/src/main/java/com/breez/client/plugins/breez/breezlib/GoogleAuthenticator.java
+++ b/android/app/src/main/java/com/breez/client/plugins/breez/breezlib/GoogleAuthenticator.java
@@ -31,9 +31,12 @@ import io.flutter.plugin.common.PluginRegistry;
 public class GoogleAuthenticator implements PluginRegistry.ActivityResultListener {
     private static final String TAG = "BreezGAuthenticator";
     private static final int AUTHORIZE_ACTIVITY_REQUEST_CODE = 84;
+    private static final int RC_REQUEST_PERMISSION_APP_DATA = 83;
     private static final int SIGN_IN_CANCELLED = 12501;
+    private static final Scope appDataScope = new Scope(DriveScopes.DRIVE_APPDATA);
     private final ActivityPluginBinding activityBinding;
     TaskCompletionSource<GoogleSignInAccount> m_signInProgressTask;
+    TaskCompletionSource<Boolean> m_scopeRequestTask;
     private GoogleSignInClient m_signInClient;
 
     public GoogleAuthenticator(ActivityPluginBinding binding) {
@@ -98,6 +101,11 @@ public class GoogleAuthenticator implements PluginRegistry.ActivityResultListene
             credential.setSelectedAccount(googleAccount.getAccount());
             return credential.getToken();
         } catch (Exception e) {
+            // TODO Handle NEED_REMOTE_CONSENT error
+           /* if (e.getMessage().contains("NEED_REMOTE_CONSENT")) {
+                Tasks.await(requestAppDataScope());
+                return getAccessToken();
+            }*/
             Log.w(TAG, "getAccessToken failed", e);
             throw e;
         }
@@ -108,7 +116,7 @@ public class GoogleAuthenticator implements PluginRegistry.ActivityResultListene
         final Activity context = activityBinding.getActivity();
         GoogleSignInOptions signInOptions =
                 new GoogleSignInOptions.Builder(GoogleSignInOptions.DEFAULT_SIGN_IN)
-                        .requestScopes(new Scope(DriveScopes.DRIVE_APPDATA))
+                        .requestScopes(appDataScope)
                         .requestEmail()
                         .requestIdToken(context.getString(R.string.default_web_client_id))
                         .build();
@@ -158,14 +166,19 @@ public class GoogleAuthenticator implements PluginRegistry.ActivityResultListene
 
     @Override
     public boolean onActivityResult(int requestCode, int resultCode, Intent data) {
-        if (requestCode == AUTHORIZE_ACTIVITY_REQUEST_CODE && m_signInProgressTask != null) {
-            Log.i(TAG, "onActivityResult requestCode = " + requestCode + " resultCode = " + resultCode + " intent came back with result");
-
-            // The Task returned from this call is always completed, no need to attach
-            // a listener.
-            Task<GoogleSignInAccount> task = GoogleSignIn.getSignedInAccountFromIntent(data);
-            handleSignInResult(task);
-            return true;
+        Log.i(TAG, "onActivityResult requestCode = " + requestCode + " resultCode = " + resultCode + " intent came back with result");
+        switch (requestCode) {
+            case RC_REQUEST_PERMISSION_APP_DATA:
+                handleScopeRequestResult(resultCode);
+                return true;
+            case AUTHORIZE_ACTIVITY_REQUEST_CODE:
+                if (m_signInProgressTask != null) {
+                    // The Task returned from this call is always completed, no need to attach
+                    // a listener.
+                    Task<GoogleSignInAccount> task = GoogleSignIn.getSignedInAccountFromIntent(data);
+                    handleSignInResult(task);
+                    return true;
+                }
         }
         return false;
     }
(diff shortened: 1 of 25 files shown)

### CODE REVIEW

**Review by ademar111190 — APPROVED** [review]

**Review by ubbabeck — APPROVED** [review]
Looks good to me. 

Tested with remote server.

File: lib/routes/initial_walkthrough/initial_walkthrough.dart [hunk 1]
```diff
@@ -0,0 +1,258 @@
...
+        setState(() {
+          // The state that has changed here is the animation object’s value
+        });
+      });
+    controller.forward();
+
+    SchedulerBinding.instance.addPostFrameCallback((_) {
+      /// Backup Request & Restore Handlers
+      handlers.addAll([
+        RestoreRequestHandler(),
```
roeierez, line 64: I don't understand why we need these handlers. Restore is a a process of loading a page with snapshots, picking one and calling the restore endpoint. Can't this be done using standard widgets and pages?
If the concern is about having multiple pages/dialogs (provider, snapshots) we can use a dedicated navigator so we can pop everything all together but currently (unless I am missing something) it looks too complicated to what we are trying to achieve.

roeierez (changes requested):

lib/routes/initial_walkthrough/initial_walkthrough.dart
erdemyerebasmaz: Main reason was to avoid UI locks and encapsulate the restore request & backup restore flows.

We can but wouldn't it require an overhaul of how these functions communicate both on BackupBloc & Breez library?

android/app/src/main/java/com/breez/client/plugins/breez/breezlib/GoogleAuthenticator.java
```diff
@@ -55,179 +60,186 @@ public void signOut() throws Exception {
...
         try {
             GoogleAccountCredential credential = credential();
             credential.setSelectedAccount(googleAccount.getAccount());
             return credential.getToken();
         } catch (Exception e) {
+            // TODO Handle NEED_REMOTE_CONSENT error
+           /* if (e.getMessage().contains("NEED_REMOTE_CONSENT")) {
+                Tasks.await(requestAppDataScope());
+                return getAccessToken();
+            }*/
```
erdemyerebasmaz, line 108: Remove

lib/bloc/backup/backup_bloc.dart
```diff
@@ -325,21 +369,227 @@ class BackupBloc {
...
+        log.warning("Failed to sign out.", error);
+        _promptBackupController.add(action.promptOnError);
+      },
+    );
+
+    backupServiceNeedLoginSink.add(true);
+    action.resolve(null);
+  }
+
+  Future _signIn() async {
```
erdemyerebasmaz, line 545: TODO: force signIn when backupServiceNeedLogin is true, however, this requires changes on iOS as force sign in isn't implemented there yet*. Afterwards, use this function on _listSnapshots. Remove all sign in error handling from _listSnapshots & _backupNow.

*: breezLib.signOut() can be called manually if sign in is needed

lib/routes/security_pin/security_and_backup/backup_tiles/backup_provider_tile.dart
```diff
@@ -0,0 +1,151 @@
...
+      ),
+    );
+  }
+
+  Future _logoutWarningDialog(BackupProvider previousProvider) async {
+    final backupBloc = AppBlocsProvider.of<BackupBloc>(context);
+    return await backupBloc.backupStateStream.first.then(
+      (backupState) async {
+        if ((backupState != BackupState.start() && previousProvider.isGDrive) ||
+            previousProvider.isGDrive) {
```
erdemyerebasmaz, line 100: The first condition should only be checked on Android platforms where default provider is set to Google Drive, otherwise the OR condition bypasses this check, prompting the user with this warning on a new account.

Also see EnableBackupDialog: https://github.com/breez/breezmobile/blob/7e1888598562cf6a6a3ac9954d923231123320a1/lib/widgets/enable_backup_dialog.dart#L263

cc: @ubbabeck

erdemyerebasmaz (commented):
I re-reviewed myself and added some comments for myself after our review session.

android/app/src/main/java/com/breez/client/plugins/breez/breezlib/GoogleAuthenticator.java
```diff
@@ -55,179 +60,186 @@ public void signOut() throws Exception {
...
+                throw new Exception("GoogleSignNotAvailable");
+            }
             Log.i(TAG, "silentSignIn continue to activity");
             return Tasks.await(signIn());
         }
     }
 
     public String getAccessToken() throws Exception {
         Log.d(TAG, "getAccessToken");
+        m_signInClient = createSignInClient();
```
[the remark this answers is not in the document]
erdemyerebasmaz: @roeierez  Doesn't seem necessary according to my tests. Removing this didn't had any side effects, there wasn't any performance bump either as far as I can tell.

This was likely a remnant from the time I was trying to fix the issues around "validateAccessTokenAllowingPrompt"

lib/routes/security_pin/security_and_backup/backup_tiles/backup_provider_tile.dart
erdemyerebasmaz: Resolved, removed the obsolete OR.
This condition was written in a way to avoid platform specific imports/conditions: `backupState != BackupState.start() && previousProvider.isGDrive`
as this only applies to Android, the initial backup state on iOS will never have previousProvider as GDrive.

PR #1186, opened 2023-08-11
Nostr Preferences

sarthak13gupta:
- Nostr Screen for Creating a New Account / Importing Private Key.
- Publishing User Defined Relays.
- Added Primal(Nostr client).
- Added support for Nip04 encrypt and decrypt operations.


erdemyerebasmaz (commented):
Have some NIT comments.
I'll get into reviewing bloc logic & UI at a later time.

sarthak13gupta:
Will be working on these changes, also I found that the nip04 operations need a bit of correction, will do it in the upcoming commits.

lib/bloc/marketplace/nostr_settings.dart
```diff
@@ -1,43 +1,79 @@
...
 
   static const String NOSTR_SETTINGS_PREFERENCES_KEY = "nostr_settings";
 
-  NostrSettings.start()
-      : this(
-          showSnort: true,
-          isRememberPubKey: false,
-          isRememberSignEvent: false,
+// start should be done by retrieving the values set in sharedPreferences
+  NostrSettings.start({
```
roeierez, line 30: NIT: rename `start` to `initial`

lib/home_page.dart
```diff
@@ -127,6 +140,41 @@ class HomeState extends State<Home> with WidgetsBindingObserver {
...
+
+      final publicKey = await widget.nostrBloc.publicKeyStream.first;
+      commentBloc.getPubKeyResult(publicKey);
+    });
+
+    // Listener for signing the comment
+    commentBloc.signEventStream.listen((event) async {
+      final nostrPrivateKey = widget.nostrBloc.nostrPrivateKey;
+
+      widget.nostrBloc.actionsSink.add(SignEvent(event, nostrPrivateKey));
```
roeierez, line 171: It seems that it is more correct to not pass the private key to the SignEvent since the private key is already saved in the bloc. 
All operations should not read (and pass) the private key which should be private field on the bloc and used there when needed.

lib/bloc/nostr/nostr_bloc.dart
```diff
@@ -39,6 +54,14 @@ class NostrBloc with AsyncActionsHandler {
     nostrPrivateKey = await _secureStorage.read(key: "nostrPrivateKey");
```
roeierez, line 54: It seems that we are storing the private key in the golang library but read it from the secure storage. Can you explain that?

lib/bloc/nostr/nostr_model.dart
```diff
@@ -32,3 +14,29 @@ Map<int, String> eventKind = {
   10002: 'RelayList',
   30078: 'Application Specific Data'
 };
+
+Event mapToEvent(Map<String, dynamic> eventObject) {
```
roeierez, line 18: NIT: in dart the best practice is to use `toJson` and `fromJson`

lib/home_page.dart
```diff
@@ -127,6 +140,41 @@ class HomeState extends State<Home> with WidgetsBindingObserver {
       });
     });
 
+    commentBloc = Provider.of<NostrCommentBloc>(context, listen: false);
```
roeierez, line 143: I don't see the NostrCommentBloc in the codebase. Did you forget to push?

lib/home_page.dart
```diff
@@ -127,6 +140,41 @@ class HomeState extends State<Home> with WidgetsBindingObserver {
...
+
+    widget.marketplaceBloc.nostrSettingsStream.listen((event) {
```
roeierez, line 145: All these wiring between blocs should not be in the UI layer. More than that it indicates there it too much coupling between these blocs and maybe consider unify them to one NostrBloc. I need to see the NostrCommentBloc before I suggest anything.
If this is not an option we can always pass one bloc stream to another and this stream dependency can happen inside the bloc.

roeierez (changes requested):
Thanks @sarthak13gupta !
Done initial pass. Dropped some comments and questions.
It seems to me that it will be easier to manage all nostr part in one bloc (NostrBloc) instead of splitting it into several blocs that depends on each other.
The marketplace bloc might depend on the NostrBloc (for the setting purposes).

lib/home_page.dart
sarthak13gupta: It is a part of any time, you will find it here https://github.com/breez/anytime_podcast_player/pull/48

lib/home_page.dart
sarthak13gupta: Actually, it is to connect the signevent method and getpubkey method of the nostrBloc to the nostrCommentBloc inside at any time. It might not be the best place to do this, but I wanted to be able to fetch both nostrBloc and nostrCommentBloc to be able to connect them.

lib/bloc/nostr/nostr_bloc.dart
sarthak13gupta: We are storing and fetching the private key in the golang library but only for the first time when a nostrBloc instance is made . Otherwise, if a new instance is made the private key is fetched from the secure storage(where it was stored the first time the key was created and fetched from the golang library).

lib/home_page.dart
sarthak13gupta: Yes, will make the changes.

lib/bloc/nostr/nostr_bloc.dart
roeierez: If it is stored in the golang library why use secure storage at all?

sarthak13gupta: okay got it, will remove the secure storage part.

Issue #1162, opened 2023-06-15
[UX] LNURL-withdraw, make min and max amount clickable


MarnixCroes:
### General Description

prior to https://github.com/breez/breezmobile/pull/1060 the user only had to scan/paste the LNURL-withdraw and then click `redeem`.
Now the user has to manually enter an amount.
Would be nice if the min and max amount are clickable, so when the user clicks it, the amount field gets filled in.

![image](https://github.com/breez/breezmobile/assets/93143998/466bf235-1c9d-4a87-8338-00cae3794003)

clicking 5 would enter 5, clicking 20 would enter 20.

Now it requires many clicks...

kingonly:
That's a good suggestion

RubenWaterman:
At Satsback.com, we recommend Breez and Phoenix as the go-to wallets for new users.

- We heavily use LNURL-withdraw for payouts.
- We "miss" the older flow where the max amount is pre-filled automatically.
- Currently, we don't expect users to input the right amount manually.
- As a result, when the withdraw is settled, our Satsback interface shows the full amount to be settled, even if a user has only claimed half in the Breez wallet.

Potential Solutions:

- I just discovered we can address this issue by setting the min and max amounts to be the same. We plan to implement this.
- It would also help if the min and max amounts were clickable, as suggested by @MarnixCroes.
- Lastly, would it be an idea if only a max. amount is given, this amount still gets prefilled. Perhaps this conflicts with https://github.com/breez/breezmobile/issues/1057 and https://github.com/breez/breezmobile/pull/1060 but I'm curious as to why anyone would *not* want to claim the full amount to be withdrawn? (if no min. value is given).

Would like to hear your thoughts @kingonly

kingonly:
In your case you should should set the min max to the same value, why offer a lower minimum if you're not interested in it? No min means semantically means there's no minimum.

PR #1193, opened 2023-08-22
make lnurl clickable

ubbabeck:
Fixes https://github.com/breez/breezmobile/issues/1162

Updated translations.

I did not find a way to make different parts of the string clickable so I used a textspan widget and split the text accordingly.

 
## Sample lnurlp
**Ps textstyle in video is wrong, but correct in commit**

https://github.com/breez/breezmobile/assets/36157890/031a123d-8879-4547-b8f5-f36bd0a94162

### DIFF
--- a/lib/routes/create_invoice/create_invoice_page.dart
+++ b/lib/routes/create_invoice/create_invoice_page.dart
@@ -12,6 +12,7 @@ import 'package:breez/bloc/lnurl/lnurl_bloc.dart';
 import 'package:breez/bloc/lnurl/lnurl_model.dart';
 import 'package:breez/bloc/lsp/lsp_bloc.dart';
 import 'package:breez/bloc/lsp/lsp_model.dart';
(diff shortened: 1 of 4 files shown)


MarnixCroes (commented):
utACK
nice!

erdemyerebasmaz (changes requested):
Looks good however these pasted texts will fail validation.

Please use these parameters when formatting the clicked amount:

includeDisplayName: false // Removes currency display name 'sats, BTC'
userInput: true // This only applies BTC values and needs to be tested

ubbabeck:
> Looks good however these pasted texts will fail validation.
> 
> Please use these parameters when formatting the clicked amount:
> 
> includeDisplayName: false // Removes currency display name 'sats, BTC' userInput: true // This removes whitespaces

Thank you, I addresses this in the latest rebase. It passed the validator now. 

Tested with https://lnurl.fiatjaf.com/

erdemyerebasmaz (approved):

PR #1194, opened 2023-08-22
Use local time for latest backup time

erdemyerebasmaz:
Latest backup time is stored in UTC and this can be confusing from the users perspective.

Latest backup time is converted to local time
- after restoring snapshot,
- when initializing `BackupState` from persistent data,
- When displaying snapshot list on `RestoreDialog`


ubbabeck (approved):
Looks good to me 🙏

PR #1196, opened 2023-08-22
Add tx id information of refund item on Get Refund page

erdemyerebasmaz:


roeierez (approved):
LGTM. We just need to see that it renders OK.

yaslama (commented):
LGTM

yaslama (approved):

erdemyerebasmaz:
### Here's how it looks with mock data:



This PR also includes fixes to app bar on light theme, SendOnChain page is opened with withBreezTheme. bae0ec0 ee097319768af7112bce7302b431641e67a2d0cd


There was one other thing:
The initial Tor Enabled setting & themeData that the app opened with was being used on Unexpected Error dialog. Now they are retrieved just before displaying the dialog. 24045ed

PR #1197, opened 2023-08-22
Refactor Security & Backup page

erdemyerebasmaz:
This PR contains no functional changes.
It includes refactoring changes to make it easier to work on Security & Backup page for devs.

#### Changelist:
- SecurityPage is renamed to Security & Backup page.
-  All helper methods to build widgets are converted to Widgets.
   - Tile widgets are grouped by their type, Security or Backup. 
- Their shared functions are on Security & Backup page to increase reusability.
  - Although, this may not be that great in terms of readability.

### DIFF
--- a/lib/bloc/backup/backup_bloc.dart
+++ b/lib/bloc/backup/backup_bloc.dart
@@ -460,7 +460,7 @@ class BackupBloc with AsyncActionsHandler {
         log.warning(error.toString());
       },
     );
-    _promptBackupController.add(true);
+    _promptBackupController.add(action.promptOnError);
     action.resolve(null);
   }
 
--- a/lib/bloc/backup/backup_bloc.dart
+++ b/lib/bloc/backup/backup_bloc.dart
@@ -24,6 +24,7 @@ import 'package:sqflite/sqflite.dart';
 class BackupBloc with AsyncActionsHandler {
   static const String _signInFailedCode = "401";
   static const String _signInFailedMessage = "AuthError";
+  static const String _googleSignNotAvailable = "GoogleSignNotAvailable";
   static const String _methodNotFound = "405";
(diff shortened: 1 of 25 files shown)


erdemyerebasmaz:
### These two functional changes are introduced after the initial changes:

- Do not display OK button for undismissable backup in progress dialog 985a695
  - Backup in progress dialog on Security & Backup page is undismissible if user clicked outside of the dialog, but it wasn't if the user clicked OK button. OK button is removed for the dialogs shown on Security & Backup page.
  
- Sign out of GDrive accounts if the user is switching to GDrive from another provider bb9ebf6
  - If our current provider is Google Drive, we won't ever sign out but, if we're changing from another provider to Google Drive, we sign out first so the user has the option to select another account.

erdemyerebasmaz:
@roeierez  iCloud sign-in errors were not being handled on iOS and it caused spinners hang on the UI.

Sign-in errors related to iCloud are now handled with this change 79bfbac8988cf8310144b720350d2e29c9ca9807

lib/bloc/backup/backup_bloc.dart
```diff
@@ -491,15 +539,48 @@ class BackupBloc with AsyncActionsHandler {
...
         await _breezLib.signOut();
+        log.info("Signed out of $backupProviderName");
       }
+      log.info("Signing into $backupProviderName");
       await _breezLib.signIn(_backupServiceNeedLogin, action.recoverEnabled);
+      log.info("Signed into $backupProviderName");
       await _saveAppData();
       await _breezLib.requestBackup();
       action.resolve(true);
+    } on PlatformException catch (e) {
```
ubbabeck, line 559: Can we not handle this only on sign in and throw an error which invokes the "EnableBackupDialog"?

PR #1199, opened 2023-08-25
Ubbabeck security page refactor

ubbabeck:
Tested with remote storage.

@erdemyerebasmaz Do you think this cause unwanted side effects with iCloud or GDrive ?

### DIFF
--- a/lib/routes/initial_walkthrough/initial_walkthrough.dart
+++ b/lib/routes/initial_walkthrough/initial_walkthrough.dart
@@ -254,7 +254,6 @@ class _InitialWalkthroughPageState extends State<InitialWalkthroughPage>
 
   void _restoreFromBackup() async {
     log.info("Restore from Backup");
-    await _signOut();
     _showSelectProviderDialog().then((snapshots) {
       _showRestoreDialog(snapshots).then((restoreRequest) {
         _restore(restoreRequest);
--- a/lib/widgets/enable_backup_dialog.dart
+++ b/lib/widgets/enable_backup_dialog.dart
@@ -6,7 +6,6 @@ import 'package:breez/bloc/backup/backup_bloc.dart';
 import 'package:breez/bloc/backup/backup_model.dart';
 import 'package:breez/bloc/blocs_provider.dart';
 import 'package:breez/routes/initial_walkthrough/dialogs/select_backup_provider_dialog.dart';
-import 'package:breez/routes/security_pin/remote_server_auth/remote_server_auth.dart';
 import 'package:breez/utils/min_font_size.dart';
 import 'package:breez/widgets/error_dialog.dart';
 import 'package:breez/widgets/flushbar.dart';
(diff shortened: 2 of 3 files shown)


lib/routes/initial_walkthrough/initial_walkthrough.dart
```diff
@@ -254,7 +254,6 @@ class _InitialWalkthroughPageState extends State<InitialWalkthroughPage>
 
   void _restoreFromBackup() async {
     log.info("Restore from Backup");
-    await _signOut();
```
erdemyerebasmaz, line 257: Why was this removed?

ubbabeck: When pressing restore from backup button nothing happens on fresh install. 

Why do we need it in the first place on a fresh install?

lib/widgets/enable_backup_dialog.dart
```diff
@@ -283,7 +279,9 @@ class _BackupNowButtonState extends State<_BackupNowButton> {
               backupProvider: BackupProvider.remoteServer(),
               remoteServerAuthData: auth,
             ),
-          ).then((_) => Navigator.pop(context));
```
erdemyerebasmaz, line 286: popUntil wasn't used because backup prompt dialog can be shown on top of any route and we want to pop the dialog itself, not redirect users to home page.

lib/widgets/enable_backup_dialog.dart
```diff
@@ -258,13 +258,9 @@ class _BackupNowButtonState extends State<_BackupNowButton> {
...
   ) async {
     try {
-      if (selectedProvider.isRemoteServer) {
-        await _enterRemoteServerCredentials();
-      } else {
-        await _backupNow(
-          widget.backupSettings.copyWith(backupProvider: selectedProvider),
-        ).then((_) => Navigator.pop(context));
-      }
+      await _backupNow(
```
erdemyerebasmaz, line 261: Don't we want users to be able to switch their remote server, re-enter their auth credentials here?

lib/routes/initial_walkthrough/initial_walkthrough.dart
erdemyerebasmaz: We want users to be able to switch their Google account without requiring a re-install.

Let's assume I have 2 google accounts: Account A, B
- I selected Google Drive & then Account A, (by mistake)
- Account A's snapshots are loaded or failed to load
- I cancel the flow, press restore again and select Google Drive
- Account A is now selected by default as it's the signed in account.
- User can't switch their Google account on the installed app anymore.

lib/widgets/enable_backup_dialog.dart
erdemyerebasmaz: I now see why this is required, we do prompt for auth data on Select Backup Provider dialog.

PR #1201, opened 2023-08-29
remove `_backupServiceNeedLogin` in backup_bloc

ubbabeck:
I think we can simplify the backup_block by removing the `_backupServiceNeedLogin` variable. This solved the issue where the user had to authenticate and accept terms twice on a fresh install. GDrive also seems to upload and sync properly when needed.

Removed the timeout from _prompt_backup.


erdemyerebasmaz:
Closing, decided to better manage this variable and actually use it during signIn processes.

PR #1202, opened 2023-08-29
Add google sign in not available error

ademar111190:
How it looks like:

<video src="https://github.com/breez/breezmobile/assets/1225438/1990a555-c403-4004-ac8f-7c79cb890354" />


Nackoo2000:
@ademar111190 when does this happen?

ademar111190:
> @ademar111190 when does this happen?

@Nackoo2000  It started happening on my cellphone for tests where I had installed LineageOS; it will likely happen to any device using non-official Google Play.

ademar111190:
More information about this issue: This also happens when a Google account is managed by an organization that blocks certain Google integrations. That also happens when the user is using Fire OS devices; (AFAIK, we don´t support this second case, but the first one we need to address for sure.)

PR #1203, opened 2023-08-30
Add debug build config

ademar111190:
Add a debug config section so we can share a debug keystore to test google/firebase integrations functionalities


ubbabeck (approved):
LGTM, tested with android emulator.

roeierez (approved):

PR #1204, opened 2023-08-30
Ubbabeck security page refactor

ubbabeck:
Fixes reported iCloud issue and issue where we need to sign in twice or looping when switching google drive provider

### DIFF
--- a/lib/routes/security_pin/security_and_backup/backup_tiles/backup_provider_tile.dart
+++ b/lib/routes/security_pin/security_and_backup/backup_tiles/backup_provider_tile.dart
@@ -60,6 +60,8 @@ class _BackupProviderTileState extends State<BackupProviderTile> {
                       await _signIn();
                       return true;
                     } catch (e) {
+                      log.info("backup_provider received error "
+                          "$e when attempting to sign out");
                       rethrow;
                     } finally {
                       EasyLoading.dismiss();
@@ -109,7 +111,7 @@ class _BackupProviderTileState extends State<BackupProviderTile> {
     final backupBloc = AppBlocsProvider.of<BackupBloc>(context);
     return await backupBloc.backupStateStream.first.then(
       (backupState) async {
-        if (backupState != BackupState.start() && previousProvider.isGDrive) {
+        if (backupState != BackupState.start() || previousProvider.isGDrive) {
           return await promptAreYouSure(
             context,
             "Logout Warning",
--- a/lib/routes/security_pin/security_and_backup/backup_tiles/backup_provider_tile.dart
+++ b/lib/routes/security_pin/security_and_backup/backup_tiles/backup_provider_tile.dart
@@ -3,6 +3,7 @@ import 'package:breez/bloc/backup/backup_actions.dart';
 import 'package:breez/bloc/backup/backup_bloc.dart';
 import 'package:breez/bloc/backup/backup_model.dart';
 import 'package:breez/bloc/blocs_provider.dart';
+import 'package:breez/logger.dart';
 import 'package:breez/theme_data.dart' as theme;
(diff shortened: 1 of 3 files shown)


lib/routes/security_pin/security_and_backup/security_and_backup_page.dart
```diff
@@ -269,6 +269,8 @@ class SecurityAndBackupPageState extends State<SecurityAndBackupPage>
             );
           } catch (e) {
             EasyLoading.dismiss();
+          } finally {
```
erdemyerebasmaz, line 272: This was removed because there's a delay between when backupNow action finishes and a BACKUP_REQUEST is in place.

lib/routes/security_pin/security_and_backup/backup_tiles/backup_provider_tile.dart
```diff
@@ -112,7 +111,7 @@ class _BackupProviderTileState extends State<BackupProviderTile> {
     final backupBloc = AppBlocsProvider.of<BackupBloc>(context);
     return await backupBloc.backupStateStream.first.then(
       (backupState) async {
-        if (backupState != BackupState.start() && previousProvider.isGDrive) {
+        if (backupState != BackupState.start() || previousProvider.isGDrive) {
```
erdemyerebasmaz, line 114: @ubbabeck I don't understand this change, what was this addressing? This condition was to prevent showing this dialog on new accounts that are setting Google Drive for the first time. If this is causing issues on iOS or any other condition, we might want to add that condition, 'c' to this clause, instead of changing this:

Let's consider this (a && b), instead of (a || b), we should have ((a && b) || c).

lib/routes/security_pin/security_and_backup/security_and_backup_page.dart
ubbabeck: Understand, will remove this

lib/routes/security_pin/security_and_backup/backup_tiles/backup_provider_tile.dart
ubbabeck: The reason for this change was because is so when the user switches from remote storage or iCloud back to GDrive they where prompted to sign out of their gDrive account, which is not necessary the wished behaviour.

do we wish it to be  ?
```suggestion
        if ((backupState != BackupState.start() && previousProvider.isGDrive) ||   previousProvider.isGDrive) {
```

lib/routes/security_pin/security_and_backup/security_and_backup_page.dart
erdemyerebasmaz: Have you encountered an issue where the spinner was being stuck?

ubbabeck: Yes, but I this can due to the backupserver loading as I used my own firebase setup which has proven to be fairly unstable. I will test it with our instance and report back.

lib/routes/security_pin/security_and_backup/backup_tiles/backup_provider_tile.dart
erdemyerebasmaz: Yes, we want to give the users the ability to switch their Google accounts associated with Breez. If they select no, they continue with their previously logged in account.

But, this can be confusing, what we should do imo is sign out at all times if user switches to a non Google Drive provider. What do you think?

lib/routes/security_pin/security_and_backup/security_and_backup_page.dart
erdemyerebasmaz: We should still understand why that happens and handle that case, either by error handling or adding timeouts to all EasyLoading.show() methods as a workaround.

PR #1211, opened 2023-09-19
Hotfix for backup prompt 

ubbabeck:
This pr add adds support for the user to manually back up if the backup fails due to an interruption (closed app or backup provider unavailable). If there is an auth error the user is prompted with the new select provider flow.

### DIFF
--- a/lib/widgets/enable_backup_dialog.dart
+++ b/lib/widgets/enable_backup_dialog.dart
@@ -210,9 +210,8 @@ class _BackupNowButtonState extends State<_BackupNowButton> {
             final activeBackupHasAuthError =
                 hasAuthError && !currentProvider.isNull;
 
-            /// Attempt backup before prompting the user to select a backup provider.
-            /// That is unless there was an authentication failure
-            /// TODO: and the user has not set up a backup provider on a new account on Android.
+            /// Trigger backup before prompting the user to select a backup provider
+            /// unless there was an authentication failure on previous backup attempt.
             if (!activeBackupHasAuthError) {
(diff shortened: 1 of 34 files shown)


roeierez (approved):

lib/widgets/enable_backup_dialog.dart
```diff
@@ -196,59 +188,82 @@ class _BackupNowButtonState extends State<_BackupNowButton> {
...
 
-    return TextButton(
-      onPressed: (() async {
-        final currentProvider = widget.backupSettings.backupProvider;
-
-        // First we wish to attempt to backup before making the user manually
-        // having to select a backup provider.
-        if (widget.signInNeeded == false && currentProvider != null) {
-          Navigator.pop(context);
-          await ServiceInjector().breezBridge.requestBackup();
```
ubbabeck: Why do we remove this ?

erdemyerebasmaz: The library methods shouldn't be called directly from the UI.
We're missing crucial info on all courts if you compare this to _backupNow on BackupBloc.

The condition is the same, it did rely on BackupState errors and specifically BackupFailedException's caused by authentication errors before. However, since it was not read through stream it was not being updated. Which is a problem if our backup trigger attempt fails with the same reason. The UI was not aware of it until EnableBackupDialog is opened again.

ubbabeck: i understand, thank you for the clarification

ademar111190 (approved):

Issue #1223, opened 2023-10-02
Receive to BTC not pushing the page


ademar111190:
We are not pushing the Receive to BTC address page when it is not connecting to the LSP or getting the fees so the button becomes unresponsive. We should push the page and instead show an error when/ if there is a connection error for a better UX.

https://github.com/breez/breezmobile/assets/1225438/d6919314-6ddb-4bdf-b0c7-09924394838c

PR #1224, opened 2023-10-02
Add error dialog when receive via BTC fails

ademar111190:
Fixes https://github.com/breez/breezmobile/issues/1223

Add an error dialog if pushing the receive via BTC page fails

https://github.com/breez/breezmobile/assets/1225438/81eb1969-b277-43e7-afd3-8d9fc361fdea


ubbabeck (approved):
Looks good to me.

Issue #1221, opened 2023-09-27
Backup prompt at `Receive via BTC address` dialog


MarnixCroes:
### General Description

![image](https://github.com/breez/breezmobile/assets/93143998/2561a0ed-ce6f-4f61-a488-3b5c0ce8e159)

### How To Reproduce?

I was able to repro with this, but not 100% of the time
I've been able to repro 4 times
to repro again, I have to restart Breez, iirc

1. At homescreen, click on Backup warning triangle
2. Click on `Don't prompt again`
3. Click on `Later`
4. Click on Back up triangle again
5. click on later
6. Go to backup triangle
7. disable `Don't prompt again`
8. click on later

10. Go to Receive via BTC address
11. see error

Now at Receive via BTC address, I don't get an address anymore and I can't repro since then.

### What type of machine are you observing the error on?

Pixel mobile

### Operating System

Graphene OS

### Breez Version

latest,
versionCode 1695315644

MarnixCroes:
I'm also able to repro without all the steps/touching Backup prompt 😅

kingonly:
@MarnixCroes this isn't bug. The backup prompt can be displayed in this context.

MarnixCroes:
ok. it feels weird though.
It's ok if it shows up at a different dialog like when some funds are received.
For me the issue was that it shows up sometimes randomly.

When user goes to `Receive via BTC address` dialog, and then is typing over the address or tries to scan it and all of a sudden the backup prompt shows up. It's bad UX and feels buggy imo.

(this never happened before, only since recent update?)

kingonly:
In most cases you should already have a backup though. Honestly, I think it's ok. Nothing changed in the UX.
We'll keep monitoring and see if others complain.

PR #1226, opened 2023-10-04
Fix backup dialog being shown outside the home screen

ademar111190:
Fixes https://github.com/breez/breezmobile/issues/1221

# How to test
- Let the don't prompt checkbox at the backup warning uncheck
- Execute some action that will trigger a backup (change theme is the easiest)
- Navigate to Receive via btc address
- Wait at least 4 seconds
- The backup dialog should not appear
- Once back to the home, you should see the backup dialog

### DIFF
--- a/lib/bloc/navigator/breez_navigator_observer.dart
+++ b/lib/bloc/navigator/breez_navigator_observer.dart
@@ -1,0 +1,50 @@
+import 'package:breez/logger.dart';
+import 'package:flutter/material.dart';
+import 'package:rxdart/rxdart.dart';
+
+class BreezNavigatorObserver extends NavigatorObserver {
+  static final BreezNavigatorObserver _singleton = BreezNavigatorObserver._internal();
+
+  final List<String> _routeStack = [];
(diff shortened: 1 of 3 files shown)


ubbabeck (approved):
Tested and reviewed for android. 

Please bump iOS version in info.plist before merging.

roeierez:
@ademar111190 I think we should separate the issues here:
1. Don't show the dialog if user marked "don't show again"
2. The fact that the dialog is shown not in the home screen.

The first one should be solved, the second one IMO not.
I don't see a problem that the backup dialog is shown (if needed to ) as early as we can.

ademar111190:
@roeierez if we want to show as early as possible, https://github.com/breez/breezmobile/issues/1221 is not a bug at all and we can close this PR.

Is item 1 happening to you? I didn't see this happening.
The bug reported happens If the user "disable Don't prompt again -> click on later -> Go to Receive via BTC address" i.e. prompt is enabled to the user

roeierez:
Of course. I was under the impression that the real bug is ignoring the checkmark.
The fact that the backup prompt on top of the address is not a bug. Let's close it.

PR #1229, opened 2023-10-11
build if snapshot has data

### LINKED ISSUE #1228: spinning loader on failed swap
When someone either sends a too low or too hight amount on-chain (swap) the waring action seems to be stuck until the timeout has passed.

ubbabeck:
fixes #1228 

test: awaiting failed swap transaction to be confirmed on chain...

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -34,7 +34,7 @@ android {
         targetSdkVersion 33
         multiDexEnabled true
         versionCode 1
-        versionName "0.16.4-beta"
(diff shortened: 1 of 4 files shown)


ubbabeck:
Removed using the futurebuilder since it was redundant and blocking the `streambuilder` to return. 
 


The `AutoSizedText` widget was causing issues with rendering so the widget wouldn't appear, changed it to a regular textwidget as we use in other dialogs. 

fetch_swapstatus

roeierez (approved):
LGTM

Issue #1232, opened 2023-10-14
Breez app died (frozen blue screen)


MarnixCroes:
Couldn't do anything anymore, had to restart Breez

### How To Reproduce?

1. Receive some corn via BTC address
2. Wait a while
3. After a while, at homescreen at the rop right a spinner appears (same icon as when syncing)
4. click on it
5. see `Breez is transferring funds` dialog
6. wait a bit
7. See ERROR
8. cannot do anything anymore, see screenshots

### Screenshots

![image](https://github.com/breez/breezmobile/assets/93143998/7d98de13-5188-4079-b168-50222120841f)

![image](https://github.com/breez/breezmobile/assets/93143998/0985bc68-77f2-4be4-a888-333addd614d5)

### Logs

I checked for the specific time, but logs are missing

### What type of machine are you observing the error on?

Pixel mobile

### Operating System

graphene OS

### Breez Version

latest,
1695315644

JssDWt:
Can confirm this happened to me too.
I think this is the relevant stacktrace for me.

```
2023-10-13 10:14:17.791 [INF] BIND: Bad state: No element
#0      ListMixin.lastWhere (dart:collection/list.dart:180)
#1      NavigatorState.pop (package:flutter/src/widgets/navigator.dart:4994)
#2      _TransferFundsInProgressDialogState._pop (package:breez/routes/transfer_funds_in_progress_dialog.dart:50)
#3      _TransferFundsInProgressDialogState.initState.<anonymous closure> (package:breez/routes/transfer_funds_in_progress_dialog.dart:38)
#4      _rootRunUnary (dart:async/zone.dart:1406)
#5      _CustomZone.runUnary (dart:async/zone.dart:1307)
#6      _CustomZone.runUnaryGuarded (dart:async/zone.dart:1216)
#7      _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:339)
#8      _BufferingStreamSubscription._add (dart:async/stream_impl.dart:271)
#9      _MultiStreamController.addSync (dart:async/stream_impl.dart:1044)
#10     _MultiControllerSink.add (package:rxdart/src/utils/forwarding_stream.dart:130)
#11     _StartWithStreamSink.onData (package:rxdart/src/transformers/start_with.dart:12)
#12     _rootRunUnary (dart:async/zone.dart:1406)
#13     _CustomZone.runUnary (dart:async/zone.dart:1307)
[...]

PR #1234, opened 2023-10-16
Transfer Funds in progress dialog state management

erdemyerebasmaz:
This PR addresses #1232

Instead of popping the last item on navigation stack, the route(dialog) itself is removed if it's active.

### DIFF
--- a/lib/routes/transfer_funds_in_progress_dialog.dart
+++ b/lib/routes/transfer_funds_in_progress_dialog.dart
@@ -29,15 +29,26 @@ class _TransferFundsInProgressDialog extends StatefulWidget {
 class _TransferFundsInProgressDialogState
     extends State<_TransferFundsInProgressDialog> {
   StreamSubscription<AccountModel> _stateSubscription;
+  ModalRoute _currentRoute;
 
   @override
   void initState() {
     super.initState();
     _stateSubscription = widget.accountStream.listen((state) {
-      if (state.transferringOnChainDeposit != true) {
-        _pop();
+      if (state.transferringOnChainDeposit != true && _currentRoute.isActive) {
+        Navigator.of(context).removeRoute(_currentRoute);
       }
-    }, onError: (err) => _pop());
+    }, onError: (err) {
+      if (_currentRoute.isActive) {
+        Navigator.of(context).removeRoute(_currentRoute);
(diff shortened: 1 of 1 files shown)


lib/routes/transfer_funds_in_progress_dialog.dart
```diff
@@ -29,15 +29,26 @@ class _TransferFundsInProgressDialog extends StatefulWidget {
...
+  ModalRoute _currentRoute;
 
   @override
   void initState() {
     super.initState();
     _stateSubscription = widget.accountStream.listen((state) {
-      if (state.transferringOnChainDeposit != true) {
-        _pop();
+      if (state.transferringOnChainDeposit != true && _currentRoute.isActive) {
+        Navigator.of(context).removeRoute(_currentRoute);
```
ubbabeck, line 39: I don't  think you can access build context in initState.

erdemyerebasmaz: These are not accessed in initState per se but after subscription to a stream has established, we're also checking if _currentRoute.isActive beforehand. It's not really needed on this example but what we can do to make things more robust is scheduling the stream subscription logic once the context is available as such:
```
    WidgetsBinding.instance.addPostFrameCallback((_) {
      /// Stream Subscription Logic
    });
```

ubbabeck: Why do we wish to do this in the `initstate` instead of calling this in `didChangeDependencies`

erdemyerebasmaz: We want to subscribe to the stream only once and `didChangeDependencies` may be called multiple times per widget lifecycle. We'd need some preventative measures in `didChangeDependencies`, we do have some examples using `_isInit` variable on breezmobile.

ubbabeck: Got it, thanks for the clarifying.

roeierez (approved):

JssDWt (approved):
I don't really understand. But I have no objections. LGTM!

roeierez:
Can we merge that?

ubbabeck (approved):
LGMT

Issue #1231, opened 2023-10-13
[BUG] unable to scan same invoice for a second time


MarnixCroes:
### How To Reproduce?

1. scan an invoice
2. See `You are requested to pay:` dialog
3. click CANCEL
4. try to scan the same invoice again
5. see ERROR, cannot scan the invoice again. Only after restarting Breez

### Operating System

Graphene OS

### Breez Version

latest,
1695315644

JssDWt:
A similar, but maybe unrelated flow is:
- Scan invoice
- pay
- payment fails
- Try to scan invoice again

In that case the QR scanner doesn't pick up on the invoice if it's the same.

PR #1236, opened 2023-10-16
Fix same invoice not read by qrcode scan twice in a row

ademar111190:
Fix https://github.com/breez/breezmobile/issues/1231
I took the opportunity to update the mobile scan lib (I had to update the kotlin too) and add logs with a more specific tag
The problem was `detectionSpeed: DetectionSpeed.noDuplicates,` being kept between sessions, so I just removed it as we are controlling on our side to consume just one code anyway.

# how to test
- Open any flow that use the camera qr code reader
- Read a qr code
- close the triggered flow and open the camera again
- Read the same qr code previously read
- The flow should be triggered again

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,4 +1,8 @@
 buildscript {
+    ext {
+        kotlin_version = "1.9.10"
+    }
+
     repositories {
         google()
         mavenCentral()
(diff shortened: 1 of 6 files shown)


roeierez:
Can we merge that?

erdemyerebasmaz:
> Can we merge that?

Rebased on top of master, will merge once CI passes.

Issue #413, opened 2020-12-21
F-droid


flashtheman:
Is there some way to install breez without having google play on Android? I didn't find the app on F-droid.

tifDev:
This would be nice

AbelLykens:
Would be good. Any other way of getting the APK would be useful too.

emanuelb:
Please open RFP (Request For Packaging) for this app in F-Droid repo: https://gitlab.com/fdroid/rfp#submission-queue which is the first step for f-droid inclusion.

n4HeVQSGqDeEu6:
This is the main reason I'm not using Breez at the moment. Many bitcoiners use an alternative OS on their phones (like grapheneos)...

kingonly:
We support it via our APK: https://doc.breez.technology/Can-I-download-Breez-APK.html

AbelLykens:
Do you expect users to manually check that page weekly? Or is there a built-in update check (there is no auto update, I know, but notifying about a new release at least would really help)?

kingonly:
There's an in-app 'update' CTA

daum3ns:
> There's an in-app 'update' CTA

yeah, but the link goes to the google playstore of the breeze app... this should go to the .apk download for ungoogled phones..

kingonly:
> > There's an in-app 'update' CTA
> 
> yeah, but the link goes to the google playstore of the breeze app... this should go to the .apk download for ungoogled phones..

True, @ademar111190 please follow up. We should have the logic to identify it (@yaslama can help).

PR #1238, opened 2023-10-20
Check instalation source on Andriod and open github page instead of google play store if the instalation source is not google play

ademar111190:
Related https://github.com/breez/breezmobile/issues/413

If the app is installed from Google Play, it opens Google Play as it is today, if the installation source is somewhere else, it opens the release page on GitHub

Also breaking the method into smaller methods and adding logs

How it looks like:

https://github.com/breez/breezmobile/assets/1225438/788e24cf-33e1-48a9-b535-07a028dfcd5b

### DIFF
--- a/lib/handlers/check_version_handler.dart
+++ b/lib/handlers/check_version_handler.dart
@@ -4,80 +4,141 @@ import 'package:breez/bloc/blocs_provider.dart';
 import 'package:breez/bloc/user_profile/user_actions.dart';
 import 'package:breez/bloc/user_profile/user_profile_bloc.dart';
 import 'package:breez/theme_data.dart' as theme;
+import 'package:breez/utils/exceptions.dart';
 import 'package:breez/widgets/flushbar.dart';
 import 'package:breez/widgets/no_connection_dialog.dart';
 import 'package:breez_translations/breez_translations_locales.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_svg/svg.dart';
+import 'package:logging/logging.dart';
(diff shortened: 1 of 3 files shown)


lib/handlers/check_version_handler.dart
```diff
@@ -4,80 +4,132 @@ import 'package:breez/bloc/blocs_provider.dart';
...
+  _log.info("_handleBadVersion");
+  final texts = context.texts();
+
+  showFlushbar(
+    context,
+    buttonText: texts.handler_check_version_action_update,
+    onDismiss: () {
+      _log.info("onDismiss $defaultTargetPlatform");
+      if (defaultTargetPlatform == TargetPlatform.iOS) {
+        launchUrlString('https://testflight.apple.com/join/wPju2Du7');
```
ubbabeck, line 102: Why are we not directing ios users to app store?

ubbabeck (approved):
lgmt

lib/handlers/check_version_handler.dart
ademar111190: Good question; do you have context on this @erdemyerebasmaz ?

erdemyerebasmaz: @ademar111190 No, but why not use `Source.IS_INSTALLED_FROM_APP_STORE` & `Source.IS_INSTALLED_FROM_TEST_FLIGHT` for redirects on iOS?

ademar111190: Good call out; I'll do it.

erdemyerebasmaz (approved):

PR #1239, opened 2023-10-20
Pull logs changes applied on c-breez to breezmobile

ademar111190:
Pull recent improvements on logging applied on c-breez to breeze mobile

# How to test
- Share logs
- It should share a zip with the land log and the app log

### DIFF
--- a/android/settings.gradle
+++ b/android/settings.gradle
@@ -1,3 +1,5 @@
+import org.apache.tools.ant.DirectoryScanner
+
 include ':app'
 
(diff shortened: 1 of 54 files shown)


erdemyerebasmaz:
I understand the git & device info changes but are we sure we're not losing any Breez library & native logs with these changes?

ademar111190:
@erdemyerebasmaz I see both logs when sharing the logs; The main difference other than the git and device info is the inclusion of the log tags.
Is there any specific test that you believe I should run to make sure we are not losing anything?

erdemyerebasmaz (approved):
LGTM

PR #1242, opened 2023-10-27
Allow two lines for shareable payment row

ademar111190:
Today the shareable payment row is blocking the text as a single line; Despite being an AutoSizeText that works for English; it will be broken in other languages that need bigger phrases.

In this PR I'm changing the row to allow 2 lines; this change plus the AutoSizeText, is reasonable to assume it will be good for all languages.

# How to test
- Open the payment dialog
- You should see all the text (even if needed two lines)

# Screenshots

|Real example with the Bulgarian language|Before|After|
|---|---|---|
|image|before|after|


ubbabeck (approved):
Looks good to me

PR #1263, opened 2023-12-28
Restrict path version

PeteClubSeven:
When adding iOS support to my Satscard implementation I noticed that v1.9.0 of the path package is built with Dart 3 which is currently unsupported.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -63,7 +63,7 @@ dependencies:
   nfc_manager: ^3.3.0
   nostr_tools: <1.0.8 # Requires Flutter 3.10(depends on http ^1.1.0)
   package_info_plus: <5.0.1 # Requires Flutter 3.10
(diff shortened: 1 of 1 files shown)


ubbabeck (approved):

PeteClubSeven:
Hey it seems like the CI wasn't configured correctly so the build failed

erdemyerebasmaz:
> Hey it seems like the CI wasn't configured correctly so the build failed

That's OK, make sure you fetch latest changes and run

`flutter pub upgrade && flutter pub get && cd ios && pod update && pod install --repo-update && cd ..`

before merging. 

Thanks @PeteClubSeven

PeteClubSeven:
Closing this as the fix will be in my updated Satscard integration PR (with iOS support) which I expect to have done by the end of the week.

PR #1272, opened 2024-02-23
add min/max payment size fields to opening fee params

JssDWt:
Incorporates https://github.com/breez/breez/pull/234


JssDWt:
Closing, because these fields will only be included in LSPS2.

PR #1280, opened 2024-05-03
schedule chain sync as unique work

JssDWt:
An attempt to deduplicate sync jobs at the root level, namely where they're scheduled.

Tested with a firebase notification. The job is still triggered at least. I'm not sure whether it will actually be deduplicated.

### DIFF
--- a/android/app/src/main/java/com/breez/client/job/JobManager.java
+++ b/android/app/src/main/java/com/breez/client/job/JobManager.java
@@ -2,6 +2,7 @@ package com.breez.client.job;
 
 import com.breez.client.plugins.breez.breezlib.ChainSync;
 import androidx.work.Constraints;
+import androidx.work.ExistingWorkPolicy;
 import androidx.work.OneTimeWorkRequest;
 import androidx.work.WorkManager;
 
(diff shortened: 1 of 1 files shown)


android/app/src/main/java/com/breez/client/job/JobManager.java
```diff
@@ -31,6 +32,6 @@ private void startSyncJob(){
                                 .build())
                 .build();
 
-        WorkManager.getInstance().enqueue(oneTime);        
+        WorkManager.getInstance().enqueueUniqueWork(UNIQUE_WORK_NAME, ExistingWorkPolicy.REPLACE, oneTime);
```
roeierez, line 35: @JssDWt did you consider using https://developer.android.com/reference/androidx/work/ExistingWorkPolicy#KEEP ? It looks like the difference is that it won't cancel the pending job and just let it finish.

JssDWt: I did not consider that, that looks better indeed

JssDWt: Testing this out it seems the job is not executed when I use `ExistingWorkPolicy.KEEP`

PR #1282, opened 2024-05-06
Co-operative Close Channels

erdemyerebasmaz:
This PR implements the UI components required for changes in:
- https://github.com/breez/breez/pull/243

that introduces co-operative close channel operations.

### Changelist:
- Generate proto messages a70a127
  - Add comments that explain why "protoc_plugin v20.0.1" needs to be used.
- Add 'closeChannels' to Breez Library's iOS Bindings 46481ed80d0b182f2bd76b2a554a2ec4331b7316
- Create & add co-operative close channels page under Preferences 4d59d33aa125a694d92cf32f3a8ed5e8a5d1f6ac
  - Declare CloseChannels as account action
  - Open mail or share support mail when clicked
  - Prevent dismissing dialog when closing channels is in progress

### Other changelist:
- Update dependencies to latest
- Increase compileSdkVersion to 34(Android camera & lifecycle requirement)
- Use "given version or earlier" syntax for dependencies that require Flutter 3.10
- Add .flutter-version to .gitignore 38d90b45447abee101ed0ec893d201b0703f571c


lib/routes/close_channels/coop_closed_channels.dart
```diff
@@ -0,0 +1,241 @@
...
+                TextSpan(
+                  text: "?",
+                  style: themeData.dialogTheme.contentTextStyle,
+                ),
+              ],
+            ),
+          ),
+        ).then((ok) async {
+          if (ok) {
+            await closeChannels();
```
roeierez, line 194: It looks like we are ignoring the result here. It seems that the API returns information on each channel that was closed (or failed).
I am only worried if for some reason some of the channels weren't able to be closed so perhaps we should reflect that in the UI.
I suggest to involve @kingonly regarding the user experience here.

roeierez (changes requested):
Looks good.

lib/routes/close_channels/coop_closed_channels.dart
```diff
@@ -0,0 +1,272 @@
...
+        // TODO: Handle the case where all channels are not closed successfully
+        for (var channel in closedChannels.channels) {
+          if (channel.isSkipped) {
+            // Do sth
+          }
+          if (channel.failErr.isNotEmpty) {
+            // Do sth
+          }
+        }
+      }
```
erdemyerebasmaz, line 258: > It looks like we are ignoring the result here. It seems that the API returns information on each channel that was closed (or failed). I am only worried if for some reason some of the channels weren't able to be closed so perhaps we should reflect that in the UI. I suggest to involve @kingonly regarding the user experience here.

@roeierez Added placeholder handlers & waiting for feedback on how to move forward.

I also wonder what is the 100% success case here, is it where no channels are skipped and closed with no fail errors? Can there be a good reason to skip channels?

lib/routes/close_channels/coop_close_channels_dialog.dart
```diff
@@ -0,0 +1,235 @@
...
+  Future<void> _closeChannels() async {
+    final texts = context.texts();
+    final themeData = Theme.of(context);
+    try {
+      final accountBloc = AppBlocsProvider.of<AccountBloc>(context);
+      final closeChannelAction = CloseChannelsAction(widget.closingAddress);
+      accountBloc.userActionsSink.add(closeChannelAction);
+      await closeChannelAction.future.then((result) {
+        final closedChannels = result as List<CloseChannelResult>;
+        if (closedChannels.any((c) => !c.isSkipped || !c.failErr.isNotEmpty)) {
```
JssDWt, line 175:
```suggestion
        if (closedChannels.all((c) => !c.isSkipped || !c.failErr.isNotEmpty)) {
```

JssDWt (commented):
Just one comment

lib/routes/close_channels/coop_close_channels_dialog.dart
erdemyerebasmaz: Whoops, had `closedChannels.any((c) => c.isSkipped || c.failErr.isNotEmpty)` in mind when writing it :)
Great catch!

erdemyerebasmaz: Done dc9075fc53f5e4af3dc2b751ff250ee51a9de89f

JssDWt: Looks good

erdemyerebasmaz: @JssDWt Actually, I think it was working as expected and in a more efficient way.

This is the case where there aren't any skipped channels or channels with failure errors. Since we're interested in a full success case here, there's no need to check every channel, it's enough to find one bad apple.

erdemyerebasmaz: Reverted to syntax in 

> Whoops, had !closedChannels.any((c) => c.isSkipped || c.failErr.isNotEmpty) in mind when writing it :)
> Great catch!

and added descriptive comment 7d4a6b8f20e1f275e36a77e25cae0bfe40151ef1

JssDWt: Hmm, 
`if (closedChannels.any((c) => !c.isSkipped || !c.failErr.isNotEmpty))`
This falls into the if statement if one or more channels were successful. I think you want to fall into this if `all` channels were successful?
Maybe turn it around by doing what you suggested?
`if (!closedChannels.any((c) => c.isSkipped || c.failErr.isNotEmpty))`

erdemyerebasmaz: Done 76244e3ee92d70ce04622088391c2635990bf66a

Issue #1286, opened 2024-05-12
Auto paste doesn't work


Nackoo2000:
When I copy an invoice in the clipboard Breez does not paste it automatically after opening.
I have to paste it manually in Send > Send via ....
The same is for a BTC Adress or LNURL

erdemyerebasmaz:
Is this on iOS or Android?
Could you add logs?

Nackoo2000:
For iOS and Android the behavior is the same and consistent you can easily reproduce it.

1. Copy invoice from WoS
2. Open BreezM
3. Nothing happens

PR #1288, opened 2024-05-13
Clipboard issues

erdemyerebasmaz:
This PR addresses 
- #1286

#### Changelist:
- Remove redundant `isLightningAddress` check for clipboard item on `listenLNUrl` 0fa706dc539de8e1e13997d714348c8a90a650bc
- Don't check if account is connected or not before displaying payment request dialog edab5e4f98d7d4ce2535ef98fcafecaebbf1ed3d
- Wrap `_listenPaymentRequests` handler with `try/catch` block 62de82df2f99f8a01badeaf43b0cf144a9eb611c
- Reset `_handlingRequest` on error 97f4ccb9c8e8db6d4a9df45781f34c869a2f7a4d

### DIFF
--- a/lib/handlers/received_invoice_notification.dart
+++ b/lib/handlers/received_invoice_notification.dart
@@ -37,7 +37,7 @@ class InvoiceNotificationsHandler {
     _receivedInvoicesStream.where((payreq) => payreq != null && !_handlingRequest).listen((payreq) async {
       final navigator = Navigator.of(_context);
       var account = await _accountBloc.accountStream.firstWhere((a) => !a.initial, orElse: () => null);
-      if (account == null || !account.connected) {
+      if (account == null) {
         return;
       }
       if (!payreq.loaded) {
--- a/lib/handlers/received_invoice_notification.dart
+++ b/lib/handlers/received_invoice_notification.dart
@@ -35,54 +35,62 @@ class InvoiceNotificationsHandler {
   _listenPaymentRequests() {
     // show payment request dialog for decoded requests
     _receivedInvoicesStream.where((payreq) => payreq != null && !_handlingRequest).listen((payreq) async {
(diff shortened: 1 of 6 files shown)


lib/handlers/received_invoice_notification.dart
```diff
@@ -35,56 +35,65 @@ class InvoiceNotificationsHandler {
   _listenPaymentRequests() {
     // show payment request dialog for decoded requests
     _receivedInvoicesStream.where((payreq) => payreq != null && !_handlingRequest).listen((payreq) async {
-      final navigator = Navigator.of(_context);
-      var account = await _accountBloc.accountStream.firstWhere((a) => !a.initial, orElse: () => null);
-      if (account == null || !account.connected) {
```
roeierez, line 40: What is the reason for skip the connected condition?

erdemyerebasmaz: @roeierez I'll have to dig into this to make sure but I believe there's a race condition, or sth quite normal and `AccountModel` is assembled in parts. The first AccountModel that was not `initial`, wasn't necessarily `connected` and the invoice got skipped. The user had no feedback as to why what happened to the invoice on their clipboard.

I don't know which would be the right approach in terms of UX:

1. **Wait for account to be connected that's coming from `accountStream`**
   - `_accountBloc.accountStream.firstWhere((a) => !a.initial && a.connected,...`
   - This has the downside of payment request dialog showing much later, the user may forget what the payment was for, it could interrupt their current operations etc.

2. **Display the payment request right away** 
  - hope account is connected by the time it's paid or let it fail
  - either way, the user does not wonder what happened to the invoice on their clipboard and can retry it via "Paste Invoice or ID"
  
but I've opted for 2 as the feedback to the user is much more clear.

PR #1292, opened 2024-05-16
Only update POS currency list if there's a change

erdemyerebasmaz:
This PR improves upon previous approach which was updating currency list and adding logs on each rebuild.

POS currency list is now populated 
- once on `initState` when the widget is built
- every time there's a change* on currency list
  - *: removal/addition/reordering of preferred fiat currencies

### DIFF
--- a/lib/routes/charge/cart/cart_bar_drop_down_button.dart
+++ b/lib/routes/charge/cart/cart_bar_drop_down_button.dart
@@ -3,12 +3,13 @@ import 'package:breez/bloc/user_profile/currency.dart';
 import 'package:breez/routes/charge/cart/cart_bar_drop_down_item.dart';
 import 'package:breez/routes/charge/currency_wrapper.dart';
 import 'package:breez/theme_data.dart' as theme;
+import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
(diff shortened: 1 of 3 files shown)


lib/routes/charge/cart/cart_bar_drop_down_button.dart
```diff
@@ -20,50 +21,77 @@ class CartBarDropDownButton extends StatelessWidget {
...
+  State<CartBarDropDownButton> createState() => _CartBarDropDownButtonState();
+}
+
+class _CartBarDropDownButtonState extends State<CartBarDropDownButton> {
+  List<CurrencyWrapper> currencyList = [];
+
+  @override
+  void initState() {
+    super.initState();
+    WidgetsBinding.instance.addPostFrameCallback((_) {
```
roeierez, line 34: Do we need the postFrameCallback?

roeierez (changes requested):

lib/routes/charge/cart/cart_bar_drop_down_button.dart
erdemyerebasmaz: No, it's residue code from another implementation. Removing.

PR #1319, opened 2024-11-28
"Update URL for Sms4Sats"

Gudnessuche:
Changes made: Corrected the URL for the 'Sms4Sats' service from 'http://sms4sats.com/' to 'https://sms4sats.com/'. This ensures that the application uses a secure (HTTPS) connection when accessing the Sms4Sats API, which helps protect user data and maintain trust in the service.

PR #1320, opened 2024-11-28
Update Key Name for The Bitcoin Company

Gudnessuche:
Changes made: Updated the key name from 'tbc' to 'TBC' in the configuration for The Bitcoin Company.


kingonly:
Why is th

Gudnessuche:
It just makes more sense that if 3 words that are all proper nouns, it'd be represented that way.
minor change tbh, but things like tftc for example is usually capitalized as TFTC.

Gudnessuche:
Just going through the json object I also found little corrections like shown in this [PR](https://github.com/breez/breezmobile/pull/1319) as well

roeierez (approved):
LGTM

PR #1340, opened 2026-01-12
fix: increase boost message character limit from 500 to 1024

### LINKED ISSUE #1318: Increase character limit for boost message in podcasting 2.0
### General Description

When boosting in the podcast feature the app allows only for 128 characters long message.  It would be helpfull and user friendly if it would support longer messages.
If there needs to be a limit. 1024 seems to be around 130 words, which seems at least for me to be a bit more functional.

### How To Reproduce?

1. Go to 'podcasts'
2. Click on a show. if nothing is there, add one show and after adding click on that.
3. In the episode list click play on any episode.
4. At bottom left, tap on 'BOOST!' 
5. in the field of showwing used characters see that only 128 are available.

### Breez Version
i do not find where in the UI to show for the version. It is a version that is current today, which is 2024-11-26.

dakshhhhh16:
fixes #1318 
Length limit increased

### DIFF
--- a/lib/routes/podcast/boost_message_dialog.dart
+++ b/lib/routes/podcast/boost_message_dialog.dart
@@ -78,7 +78,7 @@ class BoostMessageDialogState extends State<BoostMessageDialog> {
             keyboardType: TextInputType.multiline,
             textInputAction: TextInputAction.next,
             maxLines: null,
-            maxLength: 128,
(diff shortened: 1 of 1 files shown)


kingonly:
Thanks for the PR, but this project is on maintenance mode. We're not accepting external PRs unless they address critical issues.

